PR #3, opened 2021-07-15
Add CTC training

csukuangfj:
There are various code formatting and style issues in snowfall since 
it is written by different people with different preferred styles.

This pull request tries to ensure that code styles in icefall are as consistent as possible; all happens automagically with the help of the following tools:
- https://github.com/psf/black
- https://github.com/PyCQA/isort
- ~https://github.com/python/mypy~
- https://github.com/PyCQA/flake8


pzelasko:
+1 for black, not sure about mypy — it will be very strict about typing and we might end up spending a lot of extra effort to adhere to its strict checks.

csukuangfj:
>  it will be very strict about typing and we might end up spending a lot of extra effort to adhere to its strict checks.

ok, in that case, I can remove mypy.

csukuangfj:
I am trying to add the data preparation part for the LibriSpeech recipe and am trying to put everything in Python.

pzelasko:
Are you considering porting `prepare_lang.sh` to Python? I have wanted to do it for some time now... it will be very useful for adding new recipes.

danpovey:
If we do port it to Python I don't think we need an entire Kaldi-compatible lang directory, we can just keep the things we need for k2.  We might have several different possible formats, e.g. a lexicon-based version and a BPE-based version with different information.  But I think writing things to files in a directory is a good idea as it makes them easy to inspect.

egs/librispeech/ASR/prepare.sh
```diff
@@ -19,8 +18,53 @@ fi
...
+  # If you have pre-downloaded it to /path/to/LibriSpeech,
+  # you can create a symlink to avoid downloading it again:
+  #
+  #   ln -sfv /path/to/LibriSpeech data/
+  #
+
   mkdir -p data/LibriSpeech
-  # TODO
+
+  if [ ! -f data/LibriSpeech/train-other-500/.completed ]; then
```
pzelasko, line 29: I don't think that this check is technically needed; i.e. the download script would figure the completion status by itself

OTOH I am concerned about fixing the location of the dataset in `data/LibriSpeech` -- people tend to have corpora downloaded in some standard locations on their own setups, I think this should remain customizable

egs/librispeech/ASR/prepare.sh
```diff
@@ -19,8 +18,53 @@ fi
...
+    ./local/download_data.py
+  fi
+
+  # If you have pre-downloaded it to /path/to/musan,
+  # you can create a symlink to avoid downloading it again:
+  #
+  #   ln -s /path/to/musan data/
+  #
+  if [ ! -e data/musan ]; then
+    wget https://www.openslr.org/resources/17/musan.tar.gz
```
pzelasko, line 40: There is `download_musan()` here: https://github.com/lhotse-speech/lhotse/blob/master/lhotse/recipes/musan.py#L27

and `$ lhotse download musan` here: https://github.com/lhotse-speech/lhotse/blob/master/lhotse/bin/modes/recipes/musan.py#L25

csukuangfj:
> Are you considering porting prepare_lang.sh to Python?

Yes, but I am going to implement only a subset of prepare_lang.sh that is needed by snowfall, i.e., it contains
no lexiconp.txt, no unk_fst, no position-dependent phones,  no extra questions.txt, no silprob,  no word boundary, no grammar_opts.   Only those that are currently used in snowfall will be ported to Python.

If that's ok, then I will go ahead. Otherwise, I will use the current prepare_lang.sh

egs/librispeech/ASR/local/prepare_librispeech_manifest.py
```diff
@@ -0,0 +1,29 @@
...
+
+from lhotse.recipes import prepare_librispeech
+
+
+def prepare_librispeech_mainfest():
+    corpus_dir = Path("data/LibriSpeech")
+    output_dir = Path("data/manifests")
+    num_jobs = min(15, os.cpu_count())
+
+    librispeech_manifests = prepare_librispeech(
```
pzelasko, line 20: I think this and the download script can be completely replaced with:

```bash
$ lhotse download librispeech --full $CORPUS_DIR
$ lhotse prepare librispeech -j $NUM_JOBS $CORPUS_DIR $MANIFEST_DIR
```

egs/librispeech/ASR/local/compute_fbank_musan.py
```diff
@@ -0,0 +1,97 @@
...
+import subprocess
+from contextlib import contextmanager
+from pathlib import Path
+
+from lhotse import CutSet, Fbank, FbankConfig, LilcomHdf5Writer, combine
+from lhotse.recipes.utils import read_manifests_if_cached
+
+
+@contextmanager
+def get_executor():
```
pzelasko, line 19: This executor bit seems like a good candidate to move to the library-level?

pzelasko:
> > Are you considering porting prepare_lang.sh to Python?
> 
> Yes, but I am going to implement only a subset of prepare_lang.sh that is needed by snowfall, i.e., it contains
> no lexiconp.txt, no unk_fst, no position-dependent phones, no extra questions.txt, no silprob, no word boundary, no grammar_opts. Only those that are currently used in snowfall will be ported to Python.
> 
> If that's ok, then I will go ahead. Otherwise, I will use the current prepare_lang.sh

That sounds good to me!

egs/librispeech/ASR/prepare.sh
csukuangfj: > the download script would figure the completion status by itself

If the zipped files are removed after extraction, the script will download it again. That's why I add this check to avoid invoking the download script.


>  people tend to have corpora downloaded in some standard locations on their own setups, I think this should remain customizable

Fix the location of the dataset makes the code simpler. Can we let the users create a symlink to its original dataset path, i.e., 
```
ln -s /path/to/LibriSpeech data/
```
as mentioned in the script
https://github.com/k2-fsa/icefall/blob/0b19aa09c1d1921affae37688115ebfdb090b188/egs/librispeech/ASR/prepare.sh#L21-L25

pzelasko: ah, I see, I missed the bit about the symbolic link before. I am OK with that. 

> If the zipped files are removed after extraction, the script will download it again. That's why I add this check to avoid invoking the download script.

I see... maybe this is the right way then. I'm not sure if there is a straightforward way to address this issue inside of `lhotse download`.

pzelasko: ... I take that back -- I think we can change Lhotse so that the "completed detector" is executed before downloading the files ([move this line](https://github.com/lhotse-speech/lhotse/blob/master/lhotse/recipes/librispeech.py#L58) a few lines up)

I can make this change later for all the recipes. WDYT @csukuangfj ?

csukuangfj: > I can make this change later for all the recipes.

That's would be great. In that case, I won't need to add that check here.

egs/librispeech/ASR/local/compute_fbank_musan.py
csukuangfj: Will move it to a new file local/utils.py

csukuangfj:
I've ported a subset of `prepare_lang.sh` to `local/prepare_lang.py` 

Here are some test results.

### Input lexicon.txt
```
!SIL SIL
<SPOKEN_NOISE> SPN
<UNK> SPN
f f
a a
foo f o o
bar b a r
bark b a r k
food f o o d
food2 f o o d
fo f o
```

The following are outputs:

### lexicon_disambig.txt
```
!SIL SIL
<SPOKEN_NOISE> SPN #1
<UNK> SPN #2
f f #1
a a
foo f o o #1
bar b a r #1
bark b a r k
food f o o d #1
food2 f o o d #2
fo f o #1
```

### phones.txt
```
<eps> 0
SIL 1
SPN 2
a 3
b 4
d 5
f 6
k 7
o 8
r 9
#0 10
#1 11
#2 12
```

### words.txt
```
<eps> 0
!SIL 1
<SPOKEN_NOISE> 2
<UNK> 3
a 4
bar 5
bark 6
f 7
fo 8
foo 9
food 10
food2 11
#0 12
<s> 13
</s> 14
```

### L.fst


### L_disambig.fst

danpovey:
Great!!
BTW, I anticipate possibly having different versions of the lexicon for BPE and phones, either separately (separate systems) or at the same time
(jointly trained systems).  The BPE version might use a more compact representation, e.g as a ragged array, at the python level, although that's separate from the on-disk representation where we may choose something maximally human readable.
I'm not trying to dictate anything here about the formats-- I would probably overthink things-- I'm just saying this stuff so you can keep it in mind and make your own decisions.
It might be that the Lexicon object, if there is such a thing, would have phone and BPE versions with different characteristics and perhaps some methods in common but not all.  Again, I don't want to dictate any of this, just giving ideas.

csukuangfj:
I just found that `prepare_lang.sh` adds base phones to the lexicon. 
That is, there are AA0, AA1, AA2 in the lexicon; that script adds a new phone `AA` to the `phones.txt`.

I cannot think of any benefits of adding `AA`. It never appears in the training data since it does not
exist in the lexicon. Moreover, it increases the number of neural output units and never gets trained.

@danpovey Should the python version `prepare_lang.py` follow `prepare_lang.sh` to add `AA` or just ignore it?

danpovey:
You can ignore it.  The only time it's really needed is for the optional silence.
Note: technically the silence can appear inside words-- at least, we have no check against this-- which is why SIL_B snd so on still exist.

egs/librispeech/ASR/local/prepare_lang.py
```diff
@@ -213,12 +215,46 @@ def generate_id_map(symbols: List[str]) -> Dict[str, int]:
...
+        A list-of-list. The sublist contains
+        `[src_state, dest_state, label, aux_label, score]`
+
+    Return:
+      Return new `arcs` that contain self-loops.
+    """
+    states_needs_self_loops = set()
+    for arc in arcs:
+        src, dst, ilable, olable, score = arc
+        if olable != 0:
```
danpovey, line 241: lable -> label

csukuangfj: Thanks. Fixed.

egs/librispeech/ASR/prepare.sh
```diff
@@ -0,0 +1,120 @@
...
+
+if [ $stage -le 4 ] && [ $stop_stage -ge 4 ]; then
+  echo "Stage 4: Compute fbank for musan"
+  mkdir -p data/fbank
+  ./local/compute_fbank_musan.py
+fi
+
+if [ $stage -le 5 ] && [ $stop_stage -ge 5 ]; then
+  echo "Stage 5: Prepare phone based lang"
+  # TODO: add BPE based lang
```
danpovey, line 81: Incidentally, something I want to try (I was working on this in Snowfall), is to generate a BPE lexicon parallel to the phone-based lexicon, and generate a dual lexicon that contains both phones (with disambig symbols) and BPE symbols.  So we could call this a "dual lang directory".  It would have words.txt, phones.txt, bpe.txt.

danpovey: .. then, I was thinking, we could train on both phone and BPE symbols-- perhaps alternating them on different minibatches, if time is a concern.  We can even decode like this, by manipulating/rescoring lattices.

danpovey:
... IDK whether this is a good idea, but I'm thinking that our model at the top level could simply be a collection of named things, possibly just a nn.Module with sub-modules but no forward() function.  So I was thinking we could have an 'encoder', and then, supposing we have a dual system with BPE and phones, have submodules called phones_ctc and bpe_ctc which would produce the output to be decoded with CTC or LF-MMI.  They might even have internal layers.  (I'm not sure what names are appropriate here-- we should probably share the same name for CTC and LF-MMI since they are decoded the same).
I think this dual-objective approach may turn out to provide the same generalization benefit that training alongside a transformer objective provides.

.github/workflows/test.yml
```diff
@@ -0,0 +1,78 @@
...
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install Python dependencies
+        run: |
+          python3 -m pip install --upgrade pip pytest kaldialign
+          pip install k2==${{ matrix.k2-version }}+cpu.torch${{ matrix.torch }} -f https://k2-fsa.org/nightly/
+
+          # Don't use: pip install lhotse
+          # since it installs a version of PyTorch that is not predictable
```
pzelasko, line 54: Any suggestions how to fix it in on Lhotse side? I ran into this issue too..

icefall/dataset/asr_datamodule.py
```diff
@@ -0,0 +1,248 @@
+import argparse
```
pzelasko, line 1: Regarding this file, since I added it to snowfall and used it I realized it’s a bit limiting for experimentation with the data pipeline. If you look at my GigaSpeech recipe, you’ll notice I made a local copy of this file in the recipe directory, modified it accordingly, and the train/decode scripts import it locally like this `from asr_datamodule import GigaSpeechAsrDataModule`. I think this design is much more flexible because if you need drastic changes you can copy this around or even make multiple versions.

icefall/dataset/librispeech.py
```diff
@@ -0,0 +1,68 @@
+import argparse
```
pzelasko, line 1: Regarding my previous comment: this file naturally gets merged together with `asr_datamodule` when it’s a local copy

icefall/lexicon.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class Lexicon(object):
+    """Phone based lexicon.
+
+    TODO: Add BpeLexicon for BPE models.
+    """
+
+    def __init__(
+        self, lang_dir: Path, disambig_pattern: str = re.compile(r"^#\d+$")
```
pzelasko, line 17: I think that `__init__` should be changed to be able to construct a Lexicon from existing symbol tables in program memory, which also makes it easier to test this class, e.g.:

```python
# Construct arbitrary object
lexicon = Lexicon(phones=phone_symbol_table, words=word_symbol_table, L=L_fst)

# Construct from a lang dir
lexicon = Lexicon.from_dir(‘data/lang’)

# Store in a lang dir
lexicon.to_dir(‘data/lang’)
```

egs/librispeech/ASR/prepare.sh
pzelasko: Cool idea.

egs/librispeech/ASR/local/prepare_lang.py
```diff
@@ -0,0 +1,417 @@
...
+from pathlib import Path
+from typing import Any, Dict, List, Tuple
+
+import k2
+import torch
+
+Lexicon = List[Tuple[str, List[str]]]
+
+
+def read_lexicon(filename: str) -> Lexicon:
```
pzelasko, line 35: It’s confusing because we already have a `Lexicon` class elsewhere; do you think this function could be a constructor-method on that class? Technically you can derive all of L.fst, phones.txt, and words.txt from lexicon[p].txt.

egs/librispeech/ASR/local/prepare_lang.py
```diff
@@ -0,0 +1,417 @@
...
+    arcs = [[str(i) for i in arc] for arc in arcs]
+    arcs = [" ".join(arc) for arc in arcs]
+    arcs = "\n".join(arcs)
+
+    fsa = k2.Fsa.from_str(arcs, acceptor=False)
+    return fsa
+
+
+def main():
+    out_dir = Path("data/lang")
```
pzelasko, line 354: If these constants became default arguments with `argparse` or `click`, then this is a pretty cool example of a Python “library-module+script” (you can both import from it and run it as an executable). Do you think that it needs to stay in local (i.e. do you expect other recipes will deviate from this workflow?)

egs/librispeech/ASR/local/prepare_lang.py
pzelasko: After reading the whole file, I think we should either stick to the `Lexicon` class and move most of these functions there, or remove the `Lexicon` class altogether, these are clearly very overlapping.

egs/librispeech/ASR/tdnn_lstm_ctc/train.py
```diff
@@ -0,0 +1,493 @@
...
+            "beam_size": 10,
+            "reduction": "sum",
+            "use_double_scores": True,
+        }
+    )
+
+    return params
+
+
+def load_checkpoint_if_available(
```
pzelasko, line 135: Why not move this and `save_checkpoint` to the library part?

egs/librispeech/ASR/tdnn_lstm_ctc/train.py
```diff
@@ -0,0 +1,493 @@
...
+        use_double_scores=params.use_double_scores,
+    )
+
+    assert loss.requires_grad == is_training
+
+    # train_frames and valid_frames are used for printing.
+    if is_training:
+        params.train_frames = supervision_segments[:, 2].sum().item()
+    else:
+        params.valid_frames = supervision_segments[:, 2].sum().item()
```
pzelasko, line 283: I think that using global configuration for passing “surprise outputs” around is confusing and not a good idea.

egs/librispeech/ASR/local/prepare_lang.py
csukuangfj: I agree.

.github/workflows/test.yml
csukuangfj: The following are the installation logs from GitHub actions:
```
Requirement already satisfied: torch>=1.7.1 in /opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages (from lhotse===0.8.0.dev-e95d134-clean) (1.8.1)
Collecting torchaudio>=0.7.1
  Downloading torchaudio-0.9.0-cp37-cp37m-manylinux1_x86_64.whl (1.9 MB)

...

Collecting torch>=1.7.1
  Downloading torch-1.9.0-cp37-cp37m-manylinux1_x86_64.whl (831.4 MB)
```

---

I think the line `torchaudio>=0.7.1` in lhotse's requirements.txt is problematic.

Imagine the following case:
(1) A user has not installed torchaudio before
(2) The user has installed torch==1.8.1
(3) If he tries to install lhotse, pip sees that torch 1.8.1 is newer than 1.7.1, so torch>=1.7.1 is skipped
(4) pip sees `torchaudio>=0.7.1`, it tries to install the latest torchaudio, i.e., 0.9.0
(5) torchaudio 0.9.0 depends on torch 1.9.0
(6) pip replaces the already installed torch 1.8.1 with 1.9.0. OOPS

----

One solution I would recommend is to modify lhotse's setup.py:
(1) Detect if PyTorch is installed. If it is not installed or its version is < 1.7.1, then no changes to lhotse are needed
(2) If it is installed and its version is >= 1.7.1, set the version of torchaudio to match the version of PyTorch.
     For example, if torch==1.8.1 is detected, fix the version of torchaudio to 0.8.0.

pzelasko: Thanks, good idea. Will do.

icefall/dataset/librispeech.py
danpovey: I wonder if it might be possible to have Python scripts like this (librispeech.py) localized to the individual recipes somehow, so users feel more free to copy and modify them?

egs/librispeech/ASR/local/prepare_lang.py
danpovey: Guys, let's be cautious about putting things in the shared global space until we feel the recipes have stabilized.
I don't want to build up too much shared infrastructure while there are questions about whether something is a stable pattern yet.  It's always easy to put something in the icefall library, but taking things out can be hard.

danpovey: .. also, @pzelasko, I actually agree with the way @csukuangfj has done it here.

There may be some overlap, yes, but I think it's nice to have a separate script for turning the lexicon.txt into the lang dir.  This makes it easy to modify this script separately from the reading code (and we can know, for example, that the code responsible for reading the lexicon won't break when we change this).  I always prefer having a self-contained script, versus one that depends on a large library, because the self-contained script can be reasoned about in isolation, without worrying about which library version it depends on and what the interface of that library is.  So it is more self-documenting.

Also, at some point we should formalize the requirements on the input lexicon.txt file; and we'll eventually need a validation process for the lexicon.txt.  I'm thinking that we should have the concept of a "dict dir" like in Kaldi, and later on we may choose to have the phones.txt be an optional extra input in that dict dir in case we need to sync phone-sets across lexicons that have some phones missing.   By keeping the "dict dir" and "lang dir" separate, we can clarify inputs vs. output, and make the task of validation more straightforward to understand.

icefall/dataset/librispeech.py
pzelasko: Yes, see asr_datamodule.py in gigaspeech recipe after my latest changes, it’s exactly what I did there.

egs/librispeech/ASR/local/prepare_lang.py
pzelasko: Alright, I see your point.

egs/librispeech/ASR/tdnn_lstm_ctc/train.py
csukuangfj: Thanks. Will fix the remaining comments after getting the BPE decoding part working (probably tomorrow)

csukuangfj:
Here are the decoding results with icefall (using only the encoder part) from the pre-trained model (downloaded from https://huggingface.co/GuoLiyong/snowfall_bpe_model/tree/main/exp-duration-200-feat_batchnorm-bpe-lrfactor5.0-conformer-512-8-noam, as mentioned in https://github.com/k2-fsa/snowfall/pull/227)

### HLG - no LM rescoring

(output beam size is 8)

#### 1-best decoding

```
[test-clean-no_rescore] %WER 3.15% [1656 / 52576, 127 ins, 377 del, 1152 sub ]
[test-other-no_rescore] %WER 7.03% [3682 / 52343, 220 ins, 1024 del, 2438 sub ]
```

#### n-best decoding

For n=100,

```
[test-clean-no_rescore-100] %WER 3.15% [1656 / 52576, 127 ins, 377 del, 1152 sub ]
[test-other-no_rescore-100] %WER 7.14% [3737 / 52343, 275 ins, 1020 del, 2442 sub ]
```

For n=200,

```
[test-clean-no_rescore-200] %WER 3.16% [1660 / 52576, 125 ins, 378 del, 1157 sub ]
[test-other-no_rescore-200] %WER 7.04% [3684 / 52343, 228 ins, 1012 del, 2444 sub ]
```

### HLG - with LM rescoring

#### Whole lattice rescoring

```
[test-clean-lm_scale_0.8] %WER 2.77% [1456 / 52576, 150 ins, 210 del, 1096 sub ]
[test-other-lm_scale_0.8] %WER 6.23% [3262 / 52343, 246 ins, 635 del, 2381 sub ]
```

WERs of different LM scales are:

```
For test-clean, WER of different settings are:
lm_scale_0.8    2.77    best for test-clean
lm_scale_0.9    2.87
lm_scale_1.0    3.06
lm_scale_1.1    3.34
lm_scale_1.2    3.71
lm_scale_1.3    4.18
lm_scale_1.4    4.8
lm_scale_1.5    5.48
lm_scale_1.6    6.08
lm_scale_1.7    6.79
lm_scale_1.8    7.49
lm_scale_1.9    8.14
lm_scale_2.0    8.82

For test-other, WER of different settings are:
lm_scale_0.8    6.23    best for test-other
lm_scale_0.9    6.37
lm_scale_1.0    6.62
lm_scale_1.1    6.99
lm_scale_1.2    7.46
lm_scale_1.3    8.13
lm_scale_1.4    8.84
lm_scale_1.5    9.61
lm_scale_1.6    10.32
lm_scale_1.7    11.17
lm_scale_1.8    12.12
lm_scale_1.9    12.93
lm_scale_2.0    13.77
```

#### n-best LM rescoring

n = 100

```
[test-clean-lm_scale_0.8] %WER 2.79% [1469 / 52576, 149 ins, 212 del, 1108 sub ]
[test-other-lm_scale_0.8] %WER 6.36% [3329 / 52343, 259 ins, 666 del, 2404 sub ]
```

WERs of different LM scales are:

```
For test-clean, WER of different settings are:
lm_scale_0.8    2.79    best for test-clean
lm_scale_0.9    2.89
lm_scale_1.0    3.03
lm_scale_1.1    3.28
lm_scale_1.2    3.52
lm_scale_1.3    3.78
lm_scale_1.4    4.04
lm_scale_1.5    4.24
lm_scale_1.6    4.45
lm_scale_1.7    4.58
lm_scale_1.8    4.7
lm_scale_1.9    4.8
lm_scale_2.0    4.92
For test-other, WER of different settings are:
lm_scale_0.8    6.36    best for test-other
lm_scale_0.9    6.45
lm_scale_1.0    6.64
lm_scale_1.1    6.92
lm_scale_1.2    7.25
lm_scale_1.3    7.59
lm_scale_1.4    7.88
lm_scale_1.5    8.13
lm_scale_1.6    8.36
lm_scale_1.7    8.54
lm_scale_1.8    8.71
lm_scale_1.9    8.88
lm_scale_2.0    9.02
```

n = 150

```
[test-clean-lm_scale_0.8] %WER 2.80% [1472 / 52576, 149 ins, 218 del, 1105 sub ]
[test-other-lm_scale_0.8] %WER 6.35% [3325 / 52343, 262 ins, 660 del, 2403 sub ]
```

```
For test-clean, WER of different settings are:
lm_scale_0.8    2.8     best for test-clean
lm_scale_0.9    2.89
lm_scale_1.0    3.05
lm_scale_1.1    3.3
lm_scale_1.2    3.56
lm_scale_1.3    3.83
lm_scale_1.4    4.1
lm_scale_1.5    4.32
lm_scale_1.6    4.56
lm_scale_1.7    4.73
lm_scale_1.8    4.88
lm_scale_1.9    5.01
lm_scale_2.0    5.14
For test-other, WER of different settings are:
lm_scale_0.8    6.35    best for test-other
lm_scale_0.9    6.44
lm_scale_1.0    6.63
lm_scale_1.1    6.91
lm_scale_1.2    7.25
lm_scale_1.3    7.62
lm_scale_1.4    7.92
lm_scale_1.5    8.23
lm_scale_1.6    8.43
lm_scale_1.7    8.67
lm_scale_1.8    8.89
lm_scale_1.9    9.08
lm_scale_2.0    9.23
```

danpovey:
Great!  By the n-best LM rescoring, are you talking about a neural or n-gram LM?

pzelasko:
Nice! Do we have the scripts for training the BPE models from scratch somewhere?

csukuangfj:
> Great! By the n-best LM rescoring, are you talking about a neural or n-gram LM?

I am only using the transformer encoder + HLG decoding + 4-gram rescoring.

Will integrate the attention decoder for rescoring.

csukuangfj:
> Nice! Do we have the scripts for training the BPE models from scratch somewhere?

Yes, the training code is in the pull-request: https://github.com/k2-fsa/snowfall/pull/219

We're porting it to icefall and polishing the training code.

.github/workflows/test.yml
pzelasko: ... I think a better solution might be to remove torch and torchaudio from requirements.txt. 

I started implementing what you suggested but I hate the idea of keeping track of which version of pytorch corresponds to which version of torchaudio. It would mean every time they release even a minor update, I'd have to update Lhotse's setup.py. 

If this causes issues I'd rather leave it up to the user to install the pytorch and torchaudio version they want, and if they forget to do it, run into an import error in Lhotse (at least it's fairly obvious how to fix that one).

BTW at some point I might try to decouple Lhotse from PyTorch, i.e. make it an optional dependency. The core part of the library (except Dataset API and torchaudio augmentations) should work just fine.

danpovey: .. but in that case how will we enforce the requirement of torchaudio?  would it be part of the install instructions?

pzelasko: We wouldn't enforce it, which I agree is not ideal...

I have another idea. If torch is installed, we'll detect it, and change the line in requirements to `torch==<installed-version>`, and leave torchaudio without a specified version. Then pip will have to match torchaudio to torch's version. Would that work?

csukuangfj: > I have another idea. If torch is installed, we'll detect it, and change the line in requirements to torch==<installed-version>, and leave torchaudio without a specified version. Then pip will have to match torchaudio to torch's version. Would that work?

I believe not. I just ran a simple test on my computer and verify that it does not work:
```
(py3) fangjuns-MacBook-Air:~ fangjun$ python3 -c "import torch; print(torch.__version__)"
1.7.0
(py3) fangjuns-MacBook-Air:~ fangjun$ pip install torchaudio
Looking in indexes: https://mirrors.aliyun.com/pypi/simple/
Collecting torchaudio
  Using cached https://mirrors.aliyun.com/pypi/packages/82/72/4ba3300f2d74c56e1c2a6ea3882dd17ef47c90a46fc65271ea4a93dd48cb/torchaudio-0.9.0-cp37-cp37m-macosx_10_9_x86_64.whl (1.5 MB)
Collecting torch==1.9.0
  Downloading https://mirrors.aliyun.com/pypi/packages/51/cd/3fa53975000c37b0a0e4f5db7dd2df45749bfbc5f04c5b99812df6794511/torch-1.9.0-cp37-none-macosx_10_9_x86_64.whl (127.9 MB)
     |██                              | 7.9 MB 1.9 MB/s eta 0:01:02^C
ERROR: Operation cancelled by user
WARNING: You are using pip version 20.2.4; however, version 21.2.1 is available.
You should consider upgrading via the '/Users/fangjun/software/py3/bin/python3.7 -m pip install --upgrade pip' command.
```

csukuangfj: I think torchaudio and PyTorch share the same minor version. That is, if you detect that the install pytorch is 1.7.1, you can just run
```
pip install torchaudio==0.7
```
If pytorch is 1.9.0, you can run `pip install torchaudio==0.9`.
If pytorch is 1.x.y, you can run `pip install torchaudio==0.x`

csukuangfj: If a user is using k2, he/she must have installed PyTorch, so I agree to remove `torch` from Lhotse's requirements.txt

pzelasko: I think the right command would have been `pip install torch==1.7.1 torchaudio`. Then it seems to work:

```
~ pip install torch==1.7.1
Collecting torch==1.7.1
  Using cached torch-1.7.1-cp38-none-macosx_10_9_x86_64.whl (108.9 MB)
Collecting typing-extensions
  Using cached typing_extensions-3.10.0.0-py3-none-any.whl (26 kB)
Collecting numpy
  Downloading numpy-1.21.1-cp38-cp38-macosx_10_9_x86_64.whl (16.9 MB)
     |████████████████████████████████| 16.9 MB 702 kB/s
Installing collected packages: typing-extensions, numpy, torch
Successfully installed numpy-1.21.1 torch-1.7.1 typing-extensions-3.10.0.0
(testpt) ➜  ~ pip install torch==1.7.1 torchaudio
Requirement already satisfied: torch==1.7.1 in ./miniconda3/envs/testpt/lib/python3.8/site-packages (1.7.1)
Collecting torchaudio
  Downloading torchaudio-0.9.0-cp38-cp38-macosx_10_9_x86_64.whl (1.5 MB)
     |████████████████████████████████| 1.5 MB 3.0 MB/s
Requirement already satisfied: typing-extensions in ./miniconda3/envs/testpt/lib/python3.8/site-packages (from torch==1.7.1) (3.10.0.0)
Requirement already satisfied: numpy in ./miniconda3/envs/testpt/lib/python3.8/site-packages (from torch==1.7.1) (1.21.1)
  Downloading torchaudio-0.8.1-cp38-cp38-macosx_10_9_x86_64.whl (1.5 MB)
     |████████████████████████████████| 1.5 MB 23.3 MB/s
  Downloading torchaudio-0.8.0-cp38-cp38-macosx_10_9_x86_64.whl (1.5 MB)
     |████████████████████████████████| 1.5 MB 20.9 MB/s
  Using cached torchaudio-0.7.2-cp38-cp38-macosx_10_9_x86_64.whl (1.4 MB)
Installing collected packages: torchaudio
Successfully installed torchaudio-0.7.2
```

>  I think torchaudio and PyTorch share the same minor version. That is, if you detect that the install pytorch is 1.7.1, you can just run

What if PyToch/Torchaudio move to versions 2.x and 1.x? Or patch versions go out of sync? It feels a bit too fragile to me..

> If a user is using k2, he/she must have installed PyTorch, so I agree to remove torch from Lhotse's requirements.txt

Also Lhotse has users that don't rely on k2 so ideally we'd not make it harder for them.

pzelasko: (it's not ideal because it downloads all the torchaudio versions from the newest going down to the right one, but at least it seems to work)

danpovey:
I'd rather worry about the present than the future.  This issue could
easily become a limiting factor in whether people use Lhotse.

.github/workflows/test.yml
csukuangfj: A user normally installs Lhotse by
```
python setup.py install
```
I guess users of Lhotse at present are mainly from k2 (snowfall), who have already installed a specific version of PyTorch. It is not convenient to use `pip install torch==1.7.1 torchaudio` to install PyTorch again.

It would be great if Lhotse solves all the dependencies automatically without user intervention and
without re-installing the already installed PyTorch.



You may get some ideas from what ESPnet is doing to solve the installation problems of torch and torchaudio
by looking at its `setup.py`. (Note: there is no `requirements.txt` in it)

csukuangfj: > What if PyToch/Torchaudio move to versions 2.x and 1.x?  

It requires only minor effort to update setup.py it this happens.

> Or patch versions go out of sync? 

You can ask people from torchaudio whether it is
possible that `patch versions go out of sync` (I don't think it will happen).

csukuangfj: > .. but in that case how will we enforce the requirement of torchaudio? would it be part of the install instructions?

I think we can enforce it inside `setup.py`. The version of torchaudio should be determined by the version
of PyTorch. The version of PyTorch is chosen by the user (by detecting that PyTorch is already installed).

csukuangfj: >  I hate the idea of keeping track of which version of pytorch corresponds to which version of torchaudio. It would mean every time they release even a minor update, I'd have to update Lhotse's setup.py.

You don't have to if there is a minor update, I think, at least for the current version of torchaudio and PyTorch.

pzelasko: > I guess users of Lhotse at present are mainly from k2 (snowfall), who have already installed a specific version of PyTorch. It is not convenient to use `pip install torch==1.7.1 torchaudio` to install PyTorch again.
> 
> It would be great if Lhotse solves all the dependencies automatically without user intervention and
> without re-installing the already installed PyTorch.

Precisely that’s the idea — lhotse would modify the requirements to ``torch==1.7.1 torchaudio`` internally within setup.py, so it all happens automatically. I will also look at what ESPnet does.

pzelasko: https://github.com/lhotse-speech/lhotse/pull/348

csukuangfj:
The following are the partial results (1best decoding without LM rescoring and without attention decoder)
 of the following training command:
```
./conformer_ctc/train.py \
  --bucketing-sampler 1 \
  --num-buckets 1000 \
  --concatenate-cuts 0 \
  --max-duration 200 \
  --full-libri 1 \
  --word-size 3
```

It seems it is able to reproduce what Liyong has been doing.  I think it is safe to merge now. I will fix all the comments in
other PRs.

The tensorboard log is available at
https://tensorboard.dev/experiment/78GsxqGaSvm28grxalEZpA/#scalars&_smoothingWeight=0


## epoch-0

```

[test-clean-no_rescore] %WER 34.43% [18104 / 52576, 56 ins, 13896 del, 4152 sub ]
[test-other-no_rescore] %WER 47.82% [25032 / 52343, 49 ins, 19273 del, 5710 sub ]

```

## epoch-1

```

[test-clean-no_rescore] %WER 15.64% [8225 / 52576, 121 ins, 4689 del, 3415 sub ]
[test-other-no_rescore] %WER 29.53% [15455 / 52343, 130 ins, 9454 del, 5871 sub ]

```

## average of epoch 0 and 1

```

[test-clean-no_rescore] %WER 16.69% [8775 / 52576, 222 ins, 4173 del, 4380 sub ]
[test-other-no_rescore] %WER 30.20% [15810 / 52343, 282 ins, 8426 del, 7102 sub ]

```

## epoch-2

```

[test-clean-no_rescore] %WER 11.83% [6220 / 52576, 140 ins, 3309 del, 2771 sub ]
[test-other-no_rescore] %WER 23.83% [12471 / 52343, 162 ins, 7268 del, 5041 sub ]

```

## average of epoch 1 and 2

```

[test-clean-no_rescore] %WER 9.40% [4940 / 52576, 172 ins, 2036 del, 2732 sub ]
[test-other-no_rescore] %WER 19.25% [10075 / 52343, 247 ins, 4521 del, 5307 sub ]

```

## average of epoch 0, 1 and 2

```
[test-clean-no_rescore] %WER 15.06% [7918 / 52576, 244 ins, 3516 del, 4158 sub ]
[test-other-no_rescore] %WER 29.58% [15482 / 52343, 288 ins, 8137 del, 7057 sub ]

```

## epoch-3

```

[test-clean-no_rescore] %WER 10.65% [5600 / 52576, 102 ins, 3013 del, 2485 sub ]
[test-other-no_rescore] %WER 23.10% [12091 / 52343, 127 ins, 7583 del, 4381 sub ]

```

## average of epoch 2 and 3

```
[test-clean-no_rescore] %WER 7.38% [3880 / 52576, 147 ins, 1483 del, 2250 sub ]
[test-other-no_rescore] %WER 16.26% [8512 / 52343, 207 ins, 3808 del, 4497 sub ]

```

## average of epoch 1, 2 and 3

```
[test-clean-no_rescore] %WER 8.49% [4464 / 52576, 164 ins, 1746 del, 2554 sub ]
[test-other-no_rescore] %WER 18.20% [9528 / 52343, 248 ins, 4221 del, 5059 sub ]

```

## average of epoch 0, 1, 2 and 3

```
[test-clean-no_rescore] %WER 12.93% [6800 / 52576, 224 ins, 3028 del, 3548 sub ]
[test-other-no_rescore] %WER 26.00% [13609 / 52343, 312 ins, 6635 del, 6662 sub ]

```

## epoch 4

```
[test-clean-no_rescore] %WER 8.40% [4414 / 52576, 131 ins, 2048 del, 2235 sub ]
[test-other-no_rescore] %WER 18.93% [9907 / 52343, 149 ins, 5410 del, 4348 sub ]

```

## average of epoch 3 and 4

```
[test-clean-no_rescore] %WER 6.24% [3280 / 52576, 150 ins, 1100 del, 2030 sub
[test-other-no_rescore] %WER 14.12% [7390 / 52343, 197 ins, 3058 del, 4135 sub ]

```

## average of epoch 2, 3 and 4

```

[test-clean-no_rescore] %WER 6.27% [3295 / 52576, 156 ins, 1051 del, 2088 sub ]
[test-other-no_rescore] %WER 14.61% [7645 / 52343, 232 ins, 2977 del, 4436 sub ]

```
## average of epoch 1, 2, 3 and 4

```

[test-clean-no_rescore] %WER 8.83% [4643 / 52576, 170 ins, 1952 del, 2521 sub ]
[test-other-no_rescore] %WER 20.18% [10564 / 52343, 247 ins, 4855 del, 5462 sub ]

```

## average of epoch 0, 1, 2, 3 and 4

```
[test-clean-no_rescore] %WER 14.56% [7657 / 52576, 202 ins, 3777 del, 3678 sub ]
[test-other-no_rescore] %WER 30.02% [15714 / 52343, 277 ins, 8482 del, 6955 sub ]
```


## epoch 5

```
[test-clean-no_rescore] %WER 6.97% [3665 / 52576, 126 ins, 1484 del, 2055 sub ]
[test-other-no_rescore] %WER 16.21% [8483 / 52343, 157 ins, 4283 del, 4043 sub ]

```

## average of epoch 4 and 5

```
[test-clean-no_rescore] %WER 5.47% [2877 / 52576, 144 ins, 911 del, 1822 sub ]
[test-other-no_rescore] %WER 12.90% [6752 / 52343, 218 ins, 2636 del, 3898 sub ]

```

## average of epoch 3, 4 and 5

```
[test-clean-no_rescore] %WER 5.34% [2808 / 52576, 154 ins, 781 del, 1873 sub ]
[test-other-no_rescore] %WER 12.39% [6487 / 52343, 217 ins, 2265 del, 4005 sub ]

```

## average of epoch 2, 3, 4 and 5

```
[test-clean-no_rescore] %WER 6.00% [3153 / 52576, 154 ins, 941 del, 2058 sub ]
[test-other-no_rescore] %WER 14.27% [7471 / 52343, 247 ins, 2648 del, 4576 sub ]
```

## average of epoch 1, 2, 3, 4 and 5

```
[test-clean-no_rescore] %WER 10.10% [5311 / 52576, 160 ins, 2288 del, 2863 sub ]
[test-other-no_rescore] %WER 23.36% [12229 / 52343, 221 ins, 5877 del, 6131 sub ]

```

## average of epoch 0, 1, 2, 3, 4 and 5

```
test-clean-no_rescore] %WER 22.07% [11606 / 52576, 149 ins, 6744 del, 4713 sub ]
[test-other-no_rescore] %WER 41.52% [21733 / 52343, 159 ins, 13730 del, 7844 sub ]

```

## epoch 6

```
[test-clean-no_rescore] %WER 6.67% [3507 / 52576, 101 ins, 1603 del, 1803 sub ]
[test-other-no_rescore] %WER 15.59% [8160 / 52343, 141 ins, 4465 del, 3554 sub ]

```

## average of epoch 5 and 6

```
[test-clean-no_rescore] %WER 5.23% [2750 / 52576, 127 ins, 944 del, 1679 sub ]
[test-other-no_rescore] %WER 12.23% [6403 / 52343, 187 ins, 2649 del, 3567 sub ]

```

## average of epoch 4, 5 and 6

```

[test-clean-no_rescore] %WER 5.07% [2666 / 52576, 134 ins, 821 del, 1711 sub ]
[test-other-no_rescore] %WER 11.69% [6118 / 52343, 205 ins, 2322 del, 3591 sub ]

```

## average of epoch 3, 4, 5 and 6

```
[test-clean-no_rescore] %WER 5.09% [2674 / 52576, 151 ins, 732 del, 1791 sub ]
[test-other-no_rescore] %WER 11.86% [6209 / 52343, 207 ins, 2125 del, 3877 sub ]

```
## average of epoch 2, 3, 4, 5 and 6

```
[test-clean-no_rescore] %WER 5.96% [3136 / 52576, 152 ins, 951 del, 2033 sub ]
[test-other-no_rescore] %WER 14.77% [7730 / 52343, 260 ins, 2776 del, 4694 sub ]

```

## average of epoch 1, 2, 3, 4, 5 and 6

```
[test-clean-no_rescore] %WER 12.04% [6328 / 52576, 151 ins, 2859 del, 3318 sub ]
[test-other-no_rescore] %WER 27.15% [14213 / 52343, 230 ins, 7208 del, 6775 sub ]

```

## average of epoch 0, 1, 2, 3, 4, 5 and 6

```
[test-clean-no_rescore] %WER 34.10% [17928 / 52576, 145 ins, 11849 del, 5934 sub ]
[test-other-no_rescore] %WER 55.40% [29000 / 52343, 117 ins, 20099 del, 8784 sub ]

```

danpovey:
We'll address the remaining comments from @pzelasko later on.

danpovey:
Great!
BTW, perhaps we could put shared executable scripts like parse_options.sh in shared/, which can be a soft link, rather
than local/.

csukuangfj:
Where should the `shared` directory be, `icefall/shared`?

danpovey:
Sure, that sounds OK.
.. also I think it might be a good idea, in our data-directory hierarchy, to make a very clear distinction between data that might be written to by local scripts, and data that is simply downloaded from elsewhere.  I'd like to make it possibe to set the data-download dir to some common (possibly-writable) directory so that if multiple people share it, it will just do the appropriate caching.

csukuangfj:
> .. also I think it might be a good idea, in our data-directory hierarchy, to make a very clear distinction between data that might be written to by local scripts, and data that is simply downloaded from elsewhere. I

I agree. Will do it.

pzelasko:
Cool! Nice work. I wondered about the choice of num buckets 1000, what was your motivation?

csukuangfj:
> Cool! Nice work. I wondered about the choice of num buckets 1000, what was your motivation?

The training options are copied from Liyong's work in 
https://github.com/k2-fsa/snowfall/pull/219/files#diff-fab538403388f199d70e4194f70d212991f5f72cad14f8357c4c40d4a2699ad4

@glynpu Maybe Liyong has something to say about this.

PR #6, opened 2021-08-07
Add MMI training with word pieces as modelling unit.

csukuangfj:
Will post the results once they are available.


pzelasko:
IIRC the alignments from alimdl did not help before when we checked them in snowfall; do you expect different results with the current setup?

danpovey:
We are thinking they might help training get started, where that's a problem, e.g. for MMI with BPE.

csukuangfj:
I just started the  MMI training with pre-computed alignments.

The tensorboard logs are:

- without attention decoder: https://tensorboard.dev/experiment/RsOln3DOSROOTQw9qDvcGA/#scalars&_smoothingWeight=0

- with attention decoder: https://tensorboard.dev/experiment/nL75ImgVTpOo4n62TX4v4A/#scalars&_smoothingWeight=0

---

## Without attention decoder
It throws the following warnings at some point (after several hundred batches):
Screen Shot 2021-09-28 at 7 42 00 PM


At some other point, it stops printing the above warnings and the MMI loss starts to decrease:
Screen Shot 2021-09-28 at 7 37 26 PM

Screen Shot 2021-09-28 at 7 44 35 PM

---

You can see that pre-computed alignment is helpful to make the training converge.
(We will see whether it will diverge later)

icefall/lexicon.py
```diff
@@ -142,69 +205,66 @@ def tokens(self) -> List[int]:
         return ans
 
 
-class BpeLexicon(Lexicon):
+class UniqLexicon(Lexicon):
```
pzelasko, line 208: Why is it named `UniqLexicon`? Not sure how to interpret it.

PR #7, opened 2021-08-10
Add doc about installation and usage

csukuangfj:
Will add WERs later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,13 @@
 
+# Table of Contents
+
+- [Installation](#installation)
+  * [Install k2](#install-k2)
+  * [Install lhotse](#install-lhotse)
+  * [Install icefall](#install-icefall)
+- [Run recipes](#run-recipes)
+
+
 ## Installation
 
 `icefall` depends on [k2][k2] for FSA operations and [lhotse][lhotse] for
--- a/README.md
+++ b/README.md
@@ -1,1 +1,77 @@
-Working in progress.
+
+## Installation
+
+`icefall` depends on [k2][k2] for FSA operations and [lhotse][lhotse] for
+data preparations. To use `icefall`, you have to install its dependencies first.
+The following subsections describe how to setup the environment.
+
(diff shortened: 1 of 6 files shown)


egs/librispeech/ASR/conformer_ctc/train.py
```diff
@@ -124,19 +127,20 @@ def get_params() -> AttributeDict:
     """
     params = AttributeDict(
         {
-            "exp_dir": Path("conformer_ctc/exp"),
+            "exp_dir": Path("conformer_ctc/exp_new"),
```
pkufool, line 130: I would suggest to put the exp_dir to the parent folder, e.g. `"exp_dir": Path("exp/conformer_ctc"), Path("exp/tdnn_lstm")`. So we will not copy the exp data when we need a new version model, say, conformer_ctc_v2, confomer_ctc_v3. It is better to keep only source code in the model directory.

README.md
```diff
@@ -1 +1,84 @@
...
+
+CAUTION: There are various ways to setup the environment. What we describe
+here is just one alternative.
+
+### Install k2
+
+Please refer to [k2's installation documentation][k2-install] to install k2.
+If you have any issues about installing k2, please open an issue at
+<https://github.com/k2-fsa/k2/issues>.
+
```
pkufool, line 24: It will be more friendly to put conda or pip installation guide here rather than build from source code.

egs/librispeech/ASR/conformer_ctc/train.py
danpovey: mm, it was my idea to co-locate the models and code like this.  The idea is, it's easier to keep them in sync, and given a model, it would generally be fairly clear which code generated it.  And we'd add the code to git (at least that was my idea.. although of course that would mean the code would disappear once you switched branches, which might not be ideal.)

README.md
csukuangfj: Thanks. I just remove this kind of information and refer the reader to the documentation of k2 for installation instructions.

PR #8, opened 2021-08-14
[Not for merge] Madam oom

csukuangfj:
Put it here for discussion. Not ready for merge.

It contains
- madam optimizer from Dan
- Handle OOM when using a larger `max-duration` (e.g., from 200 to 350)

### DIFF
--- a/egs/librispeech/ASR/conformer_ctc_madam_no_warmup/conformer.py
+++ b/egs/librispeech/ASR/conformer_ctc_madam_no_warmup/conformer.py
@@ -1,0 +1,918 @@
+#!/usr/bin/env python3
+
+# Copyright (c)  2021  University of Chinese Academy of Sciences (author: Han Zhu)
+# Apache 2.0
+
+import math
+import warnings
+from typing import Optional, Tuple
+
+import torch
+from torch import Tensor, nn
+from transformer import Supervisions, Transformer, encoder_padding_mask
+
+
+class Conformer(Transformer):
+    """
+    Args:
+        num_features (int): Number of input features
+        num_classes (int): Number of output classes
+        subsampling_factor (int): subsampling factor of encoder (the convolution layers before transformers)
+        d_model (int): attention dimension
+        nhead (int): number of head
+        dim_feedforward (int): feedforward dimention
+        num_encoder_layers (int): number of encoder layers
+        num_decoder_layers (int): number of decoder layers
+        dropout (float): dropout rate
+        cnn_module_kernel (int): Kernel size of convolution module
+        normalize_before (bool): whether to use layer_norm before the first block.
+        vgg_frontend (bool): whether to use vgg frontend.
+    """
+
+    def __init__(
+        self,
+        num_features: int,
+        num_classes: int,
+        subsampling_factor: int = 4,
+        d_model: int = 256,
+        nhead: int = 4,
+        dim_feedforward: int = 2048,
+        num_encoder_layers: int = 12,
+        num_decoder_layers: int = 6,
+        dropout: float = 0.1,
+        cnn_module_kernel: int = 31,
+        normalize_before: bool = True,
+        vgg_frontend: bool = False,
+        is_espnet_structure: bool = False,
+        mmi_loss: bool = True,
+        use_feat_batchnorm: bool = False,
+    ) -> None:
+        super(Conformer, self).__init__(
+            num_features=num_features,
+            num_classes=num_classes,
+            subsampling_factor=subsampling_factor,
+            d_model=d_model,
+            nhead=nhead,
+            dim_feedforward=dim_feedforward,
+            num_encoder_layers=num_encoder_layers,
+            num_decoder_layers=num_decoder_layers,
+            dropout=dropout,
+            normalize_before=normalize_before,
+            vgg_frontend=vgg_frontend,
+            mmi_loss=mmi_loss,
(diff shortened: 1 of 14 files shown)


egs/librispeech/ASR/conformer_ctc_madam_no_warmup/train.py
```diff
@@ -0,0 +1,767 @@
...
+):
+    try:
+        return compute_loss_impl(
+            params=params,
+            model=model,
+            batch=batch,
+            graph_compiler=graph_compiler,
+            is_training=is_training,
+        )
+    except RuntimeError as ex:
```
csukuangfj, line 363: This is where OOM is handled.  I am testing it. Not sure whether it works or not.

pzelasko: Not sure if it's helpful -- there is a repo that has some utilties for finding the optimal batch size in PyTorch and it has some code to deal with OOM. Maybe there is something useful that can be borrowed https://github.com/BlackHC/toma

egs/librispeech/ASR/conformer_ctc_madam_no_warmup/train.py
```diff
@@ -0,0 +1,767 @@
...
+#  from transformer import Noam
+from madam import Moam
+from torch.nn.parallel import DistributedDataParallel as DDP
+from torch.nn.utils import clip_grad_norm_
+from torch.utils.tensorboard import SummaryWriter
+
+from icefall.bpe_graph_compiler import BpeCtcTrainingGraphCompiler
+from icefall.checkpoint import load_checkpoint
+from icefall.checkpoint import save_checkpoint as save_checkpoint_impl
+from icefall.dataset.librispeech import LibriSpeechAsrDataModule
```
pzelasko, line 28: I strongly suggest making this class local too (e.g. in a local `data.py` file) -- in the current repo layout, it will make it much easier to experiment with different types of data setups and augmentations.

egs/librispeech/ASR/conformer_ctc_madam_no_warmup/train.py
danpovey: Hmm, after looking into the DDP implementation, it looks to me like the backward gradient reduction is not done in one pass after the backward is completed on individual machines, but is done during the .backward() of the model.  So I think catching errors that happen during .backward() is not going to be possible, and likewise for errors that happen during the top-level model forward() function, because ddp seems to have sync points there.  Errors in the forward for CTC or transformer decoder may be catchable though.

danpovey: ... but we can't do a new top-level model forward on new data.  Might be possible to re-do a CTC or transformer forward with a subset of the minibatch, since its backward pass would be structurally similar to the one on the entire minibatch.

egs/librispeech/ASR/conformer_ctc_madam_no_warmup/transformer.py
```diff
@@ -0,0 +1,993 @@
...
+
+        Returns:
+          A scalar tensor containing the loss without normalization.
+        """
+        assert x.size(2) == self.size
+        #  batch_size = x.size(0)
+        x = x.view(-1, self.size)
+        target = target.view(-1)
+        with torch.no_grad():
+            true_dist = x.clone()
```
danpovey, line 838: instead of x.clone(), torch.empty_like(x) would be more approprate..

egs/librispeech/ASR/conformer_ctc_madam_no_warmup/train.py
csukuangfj: The current code is able to catch CUDA OOM exceptions during `model.forward()` when trained with a single GPU,
but it hangs when trained with two GPUs using DDP. I am looking into where it hangs.

https://github.com/pytorch/pytorch/issues/18853#issuecomment-583779161
says it is possible to catch CUDA OOM exceptions also for DDP training with a single GPU.

---

We are currently catching exceptions only for the forward pass, i..e, `model.forward`, `get_tot_scores`, `attetion decoder`.

---

> So I think catching errors that happen during .backward() is not going to be possible,

Agreed.

egs/librispeech/ASR/conformer_ctc_madam_no_warmup/conformer.py
```diff
@@ -869,7 +869,10 @@ def __init__(
             groups=channels,
             bias=bias,
         )
-        self.norm = nn.BatchNorm1d(channels)
+        # NOTE(fangjun): The process hangs when using DDP
+        # if we try to recover from CUDA OOM, so we disable
+        # batchnorm layer here.
+        #  self.norm = nn.BatchNorm1d(channels)
```
csukuangfj, line 875: After disabling batch norm, training with DDP can now recover from OOM in `model.forward()` as expected.

csukuangfj:
I will port OOM handling to LF-MMI training as well.

egs/librispeech/ASR/conformer_ctc_madam_no_warmup/conformer.py
danpovey: Mm.  Hopefully with the madam optimizer, the training will still be stable without the batchnorm.  We'll have to see.  Obviously would have to compare the performance after this change.

PR #9, opened 2021-08-16
Minor fixes.

csukuangfj:
Refactor `prepare.sh` so that it supports different vocab sizes for BPE models.
It's useful to try different vocab sizes with BPE models.


danpovey:
OK; let's merge this, as I don't want to get out of date with my own work.

PR #38, opened 2021-09-07
Use new APIs with k2.RaggedTensor

csukuangfj:
Requires https://github.com/k2-fsa/k2/pull/812

`k2.RaggedTensor` is introduced to replace `k2.RaggedInt` and `k2.RaggedFloat`. The code is not backward compatible.
Also, you have to **regenerate** `HLG.pt`.

The advantages of usage `k2.RaggedTensor` are that it is more Pythonic and functions about ragged tensors are easier to discover. (See https://k2-fsa.github.io/k2/python_api/index.html)


csukuangfj:
I will merge this PR after the following GitHub actions
https://github.com/k2-fsa/k2/runs/3541188255?check_suite_focus=true
finishes, which is building the v1.7 wheel for k2.

GitHub actions in icefall needs the latest k2 (v1.7) in order to pass all the checks (tests).

icefall/utils.py
```diff
@@ -199,26 +198,25 @@ def get_texts(best_paths: k2.Fsa) -> List[List[int]]:
       Returns a list of lists of int, containing the label sequences we
       decoded.
     """
-    if isinstance(best_paths.aux_labels, k2.RaggedInt):
+    if isinstance(best_paths.aux_labels, k2.RaggedTensor):
```
csukuangfj, line 201: @glynpu

Could you help to update k2 in ESPnet?

https://github.com/espnet/espnet/blob/26902eed11eb0ce143b99d2dbce922f3d1064e35/espnet2/bin/k2_asr_inference.py#L70

needs to be changed since there is no  `k2.RaggedInt` and `k2.RaggedFloat`
in k2 v1.7.

csukuangfj: Also, we have to install k2 v1.7 in ESPnet GitHub actions.

csukuangfj:
Tests are passed, so merging it.

PR #5, opened 2021-08-04
[WIP] Multi round nbest rescoring

pkufool:
There are several decoding methods in icefall now,  their pipelines are demonstrated  as the picture below. What I want to do in this pull request locates in the **red rectangle**, it was proposed here(https://github.com/k2-fsa/snowfall/issues/232) several weeks ago.   

image

This is just at the very beginning, just copy the related previous work from k2 & snowfall.

PR #44, opened 2021-09-13
Refactor decode.py to make it more readable and more modular.

csukuangfj:


test/test_decode.py
```diff
@@ -0,0 +1,59 @@
...
+    # 20->40
+    #
+    # So there should be only 4 paths for each lattice in the Nbest object
+    assert nbest.fsa.shape[0] == 4 * 2
+    assert nbest.shape.row_splits(1).tolist() == [0, 4, 8]
+
+    nbest2 = nbest.intersect(lattice)
+    tot_scores = nbest2.tot_scores()
+    argmax = tot_scores.argmax()
+    best_path = k2.index_fsa(nbest2.fsa, argmax)
```
csukuangfj, line 58: This shows how to do nbest decoding. 

It is more readable than before, I think.

danpovey:
.. of course if it's more convenient to work on the class in Icefall, to avoid compatibility issues, that's fine with me.

test/test_decode.py
pkufool: There is a implementation of rescore nbest with attention decoder in  #5,  the return value is also a `Nbest`. 

https://github.com/k2-fsa/icefall/blob/355e3244c82de7ec9b3e24fe24d8efa6e295e86e/icefall/decode.py#L892-L900

danpovey: That's cool.
I'll let Fangjun decide what to do about this.

Yes  I think the code in this PR is clearer than before

pkufool: Ok, Fangjun could refactor decoding as his plan first, I will merge with his code later. It will cost more time to make that PR work, I am afraid :(

danpovey:
cool

icefall/decode2.py
```diff
@@ -0,0 +1,419 @@
...
+        for word in text.split():
+            if word in word_table:
+                word_ids.append(word_table[word])
+            else:
+                word_ids.append(oov_id)
+        word_ids_list.append(word_ids)
+    levenshtein_graphs = [levenshtein_graph(ids) for ids in word_ids_list]
+    refs = k2.Fsa.from_fsas(levenshtein_graphs).to(device)
+
+    # Now compute the edit distance between hyps and refs
```
csukuangfj, line 397: This shows how to use k2 to compute the **levenshtein distance** among multiple sequences at the same time.
(Ideas are from @danpovey)

If anyone is interested in it, I wrote a colab notebook to demonstrate it step by step
https://colab.research.google.com/drive/1hAo5RUb8cMY4qhdn8HASJu9foXrRdo6m?usp=sharing

pzelasko: That’s pretty cool, does it mean we could use it directly to optimize WER as a loss function?

danpovey: There would have to be some kind of expectation involved for that to work.  There are other uses for this though.

csukuangfj:
This pull-request is ready for review.
(Will use `k2.levenshtein_alignment` and `k2.levenshtein_graph` once  https://github.com/k2-fsa/k2/pull/828 gets merged)

There are fewer code duplicates. Also, it is more readable, I believe.

It produces the same WER as the previous code. Would be great if someone can test it using an existing model.

---

I can move `class Nbest` to k2 if @danpovey you think that looks nicer.
(The code for Nbest here has several new functions (e.g., compute_am_scores, compute_lm_scores) )

danpovey:
Cool!
Regarding moving that code to k2, we should definitely do that at some point, but I'm OK to keep it temporarily here if that would reduce compatibility issues in the short term.  We can also duplicate it with here and k2, and later remove it from k2.

csukuangfj:
Ok, I am merging it.

icefall now requires k2 >= `v1.9`.

PR #51, opened 2021-09-20
add tqdm for getting a progress bar of the training

songmeixu:
Through tqdm, and the bar only shows on console (not the log file).

e.g. a screenshot:
image

### DIFF
--- a/egs/librispeech/ASR/conformer_ctc/train.py
+++ b/egs/librispeech/ASR/conformer_ctc/train.py
@@ -28,6 +28,7 @@ import torch
 import torch.distributed as dist
 import torch.multiprocessing as mp
 import torch.nn as nn
+from tqdm import tqdm
 from asr_datamodule import LibriSpeechAsrDataModule
 from conformer import Conformer
 from lhotse.utils import fix_random_seed
@@ -46,6 +47,7 @@ from icefall.utils import (
     encode_supervisions,
     setup_logger,
     str2bool,
+    is_main_process,
 )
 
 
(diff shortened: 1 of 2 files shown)


icefall/utils.py
```diff
@@ -411,3 +411,20 @@ def write_error_stats(
...
         print(f"{word}   {corr} {tot_errs} {ref_count} {hyp_count}", file=f)
     return float(tot_err_rate)
+
+
+def is_main_process():
+    """Check if the current process is main.
+
+    - For DistributedDataParallel (DDP) mode:
+        The current is main process if its rank is 0, and its rank get from
+        os.environ["RANK"] (which already be set by DDP mode).
```
csukuangfj, line 421: > os.environ["RANK"] (which already be set by DDP mode).

We are not using `torch.distributed.launch` to do DDP training and I am afraid
`RANK` is not an environment variable in our training script.

Could you use `torch.distributed.get_rank()` to get the rank of a node?

Example usages can be found in lhotse, e.g.,

https://github.com/lhotse-speech/lhotse/blob/master/lhotse/dataset/sampling/base.py#L77

```python
    def _maybe_init_distributed(self, world_size: Optional[int], rank: Optional[int]):
        if world_size is not None:
            assert world_size >= 1
        if rank is not None:
            assert rank >= 0
        if not dist.is_available() or not dist.is_initialized():
            self.world_size = 1 if world_size is None else world_size
            self.rank = 0 if rank is None else rank
            return
        self.world_size = dist.get_world_size() if world_size is None else world_size
        self.rank = dist.get_rank() if rank is None else rank
        assert self.rank < self.world_size
```

---

`torch.distributed.launch` is useful for multi-node multi-GPU training. Currently, icefall
supports only single-node multi-GPU training.

csukuangfj:
Looks great and it shows the estimated remaining time of an epoch.

Just left a minor comment.

Could you fix the code style problems reported by CI)
(https://icefall.readthedocs.io/en/latest/contributing/code-style.html provides some information about how to fix code style issues)

egs/librispeech/ASR/conformer_ctc/train.py
```diff
@@ -478,126 +480,142 @@ def train_one_epoch(
...
-            compute_validation_loss(
+    with tqdm(
+        train_dl,
+        initial=0,
+        total=train_dl.sampler.num_cuts,
+        leave=False,
+        unit="cut",  # a data fragment concept comes from "lhotse"
+        dynamic_ncols=True,
+        disable=not is_main_process(),
+    ) as t:
```
danpovey, line 492: I'm sorry, I don't think I want to add this dependency.
I just want to keep the script simple and I don't think this is really necessary enough to make the complexity worth it.
Is the num_cuts field available for all samplers?  If so I'd be OK to modify the printed message to say, for instance, batch 200/16434.

csukuangfj: `lhotse` depends on `tqdm` and icefall depends on `lhotse`, so we can use `tqdm` for free.

csukuangfj: > Is the num_cuts field available for all samplers?

I don't think so.

From https://github.com/lhotse-speech/lhotse/blob/master/lhotse/dataset/sampling/bucketing.py#L157

```python3
    @property
    def num_cuts(self) -> Optional[int]:
        """
        Total number of cuts in the sampler.
        Not available when the CutSet is read in lazy mode (returns None).
        .. note: For BucketingSampler, it's the sum of num cuts in all buckets.
        """
        try:
            return sum(s.num_cuts for s in self.bucket_samplers)
        except TypeError:
            return None
```

danpovey: Mm OK, but it's also about adding complexity to the training script.  I was really trying to keep it to the minimum needed to work, and I don't think this adds enough value for the complexity.  I think adding /{...}, where the {...} is the total num_cuts, or None if not specified, would be enough.

csukuangfj: ok, then let us use as few dependencies as possible and keep the training scripts simple.

PR #54, opened 2021-09-24
Conformer lm

danpovey:


csukuangfj:
We will add decoding script to it.

danpovey:
See also https://tensorboard.dev/experiment/unF4gSyjRjua2DSKgb3BMg/
and /ceph-dan/icefall/egs/librispeech/ASR/conformer_m/exp_6

egs/librispeech/ASR/conformer_lm/dataset.py
```diff
@@ -0,0 +1,823 @@
...
+        # Total size on axis 0, == num sentences
+        return self.sentences.tot_size(0)
+
+    def __getitem__(self, i: int):
+        """
+        Return the i'th sentence, as a list of ints (representing BPE pieces, without
+        bos or eos symbols).
+        """
+        # in future will just do:
+        #return self.words[self.sentences[i]].tolist()
```
csukuangfj, line 34: Supported by https://github.com/k2-fsa/k2/pull/833

danpovey:
Cool!

egs/librispeech/ASR/conformer_lm/conformer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+        encoder_layer = MaskedLmConformerEncoderLayer(
+            d_model,
+            nhead,
+            dim_feedforward,
+            dropout,
+            cnn_module_kernel,
+        )
+        self.encoder = MaskedLmConformerEncoder(encoder_layer, num_encoder_layers,
+                                                norm=nn.LayerNorm(d_model))
```
csukuangfj, line 69: We are using pre-normalization here.
You have placed a layer norm at the end of the encoder layer, see
https://github.com/k2-fsa/icefall/blob/0cfa8c80e0fad48e14d38784baa4802e0714f842/egs/librispeech/ASR/conformer_lm/conformer.py#L907

You are using an extra layer norm here, which means you are doing
```
x = layernorm(layernorm(x))
```

See
https://github.com/k2-fsa/icefall/blob/0cfa8c80e0fad48e14d38784baa4802e0714f842/egs/librispeech/ASR/conformer_lm/conformer.py#L967-L968

----

I just realized that it is even worse.

You are using a layer norm at both ends of an encoder layer but
encoder layers are connected end-by-end, which means the output of the layer norm
from the previous encoder layer is used as the input of the layer norm of the next encoder.

egs/librispeech/ASR/conformer_lm/conformer.py
```diff
@@ -0,0 +1,1484 @@
...
+          key_padding_mask:
+             Either None, or a Tensor of shape (batch_size, seq_len) i.e. (N, T),
+             and dtype=torch.bool which has True in positions to be masked in attention
+             layers and convolutions because they represent padding at the ends of
+             sequences.
+
+
+        Returns:
+          Returns (memory, pos_emb), where:
+            `memory` is a Tensor containing the encoded data; it is of shape (N, T, C)
```
csukuangfj, line 115:
```suggestion
            `memory` is a Tensor containing the encoded data; it is of shape (T, N, C)
```

egs/librispeech/ASR/conformer_lm/conformer.py
danpovey: This "even worse" part is not quite right, because there are bypass connections, so there are paths involving residuals where the input is used without layer norm.

danpovey: .. but yes, I agree that the LayerNorm at the end of the conformer encoder is redundant.  I have since stopped using that.  But in this particular case, it would take a long time to retrain the model if we were to fix it, so I'd say leave it as-is for now.

egs/librispeech/ASR/conformer_lm/dataset.py
```diff
@@ -698,8 +712,20 @@ class does not retain a reference to the LmDataset.
...
-        self.set_epoch(0)  # this is responsible for setting self.sorted_data_indexes
-
+        if not delay_init:
+            self.set_epoch(0)  # this is responsible for setting self.sorted_data_indexes
+
+    def _sync_sizes(self, device: torch.device = torch.device('cuda')):
+        # Calling this on all copies of a DDP setup will sync the sizes so that
+        # all copies have the exact same number of batches.  I think
+        # this needs to be called with the GPU device, not sure if it would
+        # work otherwise.
```
csukuangfj, line 726: It will not work for CPU devices as DDP requires GPU devices.

egs/librispeech/ASR/conformer_lm/dataset.py
```diff
@@ -698,8 +712,20 @@ class does not retain a reference to the LmDataset.
...
         self.sentence_lengths = sentence_lengths
 
-        self.set_epoch(0)  # this is responsible for setting self.sorted_data_indexes
-
+        if not delay_init:
+            self.set_epoch(0)  # this is responsible for setting self.sorted_data_indexes
+
+    def _sync_sizes(self, device: torch.device = torch.device('cuda')):
+        # Calling this on all copies of a DDP setup will sync the sizes so that
+        # all copies have the exact same number of batches.  I think
```
csukuangfj, line 724: Shall we mention that without doing this, the training process 
will hang indefinitely at the end?

danpovey: Mm, sure...

PR #55, opened 2021-09-24
Conformer ctc bn

danpovey:
This is not at all for merge, but there is some stuff in here relating to LossRecord that turns out to be quite convenient and useful for storing and displaying loss values.  I was hoping someone could extract this part (you can change names etc. if you want, and move the code around), and add it to the repo so we can simplify the calling code.

We previously discussed doing something like this and I was against it because I felt it would make the code less self-explanatory, but after seeing the effects I think it is a net win.

### DIFF
--- a/egs/librispeech/ASR/conformer_ctc_bn_2d/train.py
+++ b/egs/librispeech/ASR/conformer_ctc_bn_2d/train.py
@@ -171,8 +171,9 @@ def get_params() -> AttributeDict:
             "reduction": "sum",
             "use_double_scores": True,
             "accum_grad": 1,
-            "att_scale": 0.6,
-            "reverse_att_scale": 0.1,  # ctc_scale == 1.0 - att_scale - reverse_att_scale
+            "att_scale": 0.3,
+            "reverse_att_scale": 0.2,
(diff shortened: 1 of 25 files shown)


luomingshuang:
Maybe I will have a try.

egs/librispeech/ASR/conformer_ctc_bn_2d/train.py
```diff
@@ -0,0 +1,858 @@
...
+            ans[k] = v * alpha
+        return ans
+
+
+    def __str__(self) -> str:
+        ans = ''
+        for k, v in self.norm_items():
+            norm_value = '%.4g' % v
+            ans += (str(k) + '=' + str(norm_value) + ', ')
+        frames = str(self['frames'])
```
csukuangfj, line 315: `__str__` assumes there is a key `frames` in `self`. However,
`__str__` invokes `norm_items`, which does not assume there is always a key `frames`.

---

Also, 
```python3
num_frames = self.get('frames', 1)
```
is clearer  and shorter than

```python3
num_frames = self['frames'] if 'frames' in self else 1
```

PR #60, opened 2021-09-29
Use LossRecord to record and print the loss for the training process

luomingshuang:
This PR is a response of the PR #55 from @danpovey. 

Now, I have extracted the lossrecord part and added to the newest version icefall. After adding the lossrecord into the newest version icefall, I also run the three training files and they run well. I have checked the tensorboard files. There is no problems for us to use LossRecord to record and print loss.  

I think this PR is ready for merging.  I also fix some spelling errors in prepare.sh, such as change `state` to `stage`. I have build a PR #57.


egs/librispeech/ASR/tdnn_lstm_ctc/train.py
```diff
@@ -260,14 +264,79 @@ def save_checkpoint(
         best_valid_filename = params.exp_dir / "best-valid-loss.pt"
         copyfile(src=filename, dst=best_valid_filename)
 
+class LossRecord(collections.defaultdict):
```
csukuangfj, line 267: Could you move it to `icefall/utils.py` so that it can be reused by other `xxx/train.py`?

csukuangfj: Also, could you ensure that GitHub actions get passed, i.e., fix style issues?

luomingshuang: > Could you move it to `icefall/utils.py` so that it can be reused by other `xxx/train.py`?

OK, I will do it. In fact, I also consider this method.

luomingshuang: > Also, could you ensure that GitHub actions get passed, i.e., fix style issues?

Let me see.

egs/librispeech/ASR/conformer_ctc/train.py
```diff
@@ -281,13 +287,80 @@ def save_checkpoint(
...
+        """
+        keys = sorted(self.keys())
+        s = torch.tensor([ float(self[k]) for k in keys ],
+                         device=device)
+        dist.all_reduce(s, op=dist.ReduceOp.SUM)
+        for k, v in zip(keys, s.cpu().tolist()):
+            self[k] = v
+
+    def write_summary(self, tb_writer: SummaryWriter, prefix: str, batch_idx: int) -> None:
+        """
```
csukuangfj, line 346: Could you follow the google style guide for docstring?
See https://google.github.io/styleguide/pyguide.html

We are using the following variant:

```python3
def fetch_smalltable_rows(table_handle: smalltable.Table,
                          keys: Sequence[Union[bytes, str]],
                          require_all_keys: bool = False,
) -> Mapping[bytes, Tuple[str]]:
    """Fetches rows from a Smalltable.

    Retrieves rows pertaining to the given keys from the Table instance
    represented by table_handle.  String keys will be UTF-8 encoded.

    Args:
      table_handle:
        An open smalltable.Table instance.
      keys:
        A sequence of strings representing the key of each table row to
        fetch.  String keys will be UTF-8 encoded.
      require_all_keys:
        If True only rows with values set for all keys will be returned.

    Returns:
      A dict mapping keys to the corresponding table row data
      fetched. Each row is represented as a tuple of strings. For
      example:

      {b'Serak': ('Rigel VII', 'Preparer'),
       b'Zim': ('Irk', 'Invader'),
       b'Lrrr': ('Omicron Persei 8', 'Emperor')}

      Returned keys are always bytes.  If a key from the keys argument is
      missing from the dictionary, then that row was not found in the
      table (and require_all_keys must have been False).

    Raises:
      IOError: An error occurred accessing the smalltable.
    """
```

luomingshuang: Will have a look.

PR #62, opened 2021-09-29
Use LossRecord to record and print the loss for the training process

luomingshuang:
This PR is a response of the PR #55 from danpovey. 

Now, I have extracted the lossrecord part and added to the newest version icefall. After adding the lossrecord into the newest version icefall, I also run the three training files and they run well. I have checked the tensorboard files. There is no problems for us to use LossRecord to record and print loss.  

I think this PR is ready for merging.  I also fix some spelling errors in prepare.sh, such as change `state` to `stage`. I have build a PR #57.   

The code style has been checked.


csukuangfj:
Thanks! Will merge it tomorrow.

pzelasko:
A bit of a nit pick but I think MetricsTracker (or some other name) would be more appropriate, as I think you are keeping track of more metrics than just the loss.

luomingshuang:
It seems that "MetricsTracker" is more elegant and richer than "LossRecord". Yes, here, we take it out as a single class, maybe it can play a more important role, not limited to "loss". 
> A bit of a nit pick but I think MetricsTracker (or some other name) would be more appropriate, as I think you are keeping track of more metrics than just the loss.

PR #64, opened 2021-09-29
setup.py

pzelasko:
I saw that the documentation mentions not needing setup.py and recommends setting PYTHONPATH instead, but I found it inconvenient and I think for some users it's simpler to run the dev installation once and not have to set the env vars in new terminals/sessions. We definitely don't lose anything by providing a setup.py.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+icefall.egg-info/
 data
 __pycache__
 path.sh
--- a/setup.py
+++ b/setup.py
@@ -4,15 +4,15 @@ from setuptools import find_packages, setup
 from pathlib import Path
 
 icefall_dir = Path(__file__).parent
-install_requires = (icefall_dir / 'requirements.txt').read_text().splitlines()
+install_requires = (icefall_dir / "requirements.txt").read_text().splitlines()
 
 setup(
-    name='icefall',
-    version='1.0',
-    python_requires='>=3.6.0',
-    description='Speech processing recipes using k2 and Lhotse.',
-    author='The k2 and Lhotse Development Team',
-    license='Apache-2.0 License',
+    name="icefall",
+    version="1.0",
+    python_requires=">=3.6.0",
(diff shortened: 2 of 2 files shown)


csukuangfj:
The convention is that `setup.py` installs everything for you.
```
python3 setup.py install xxx
```
and then you are ready to use the package `xxx`.

---

But for icefall, we have to install k2 separately since it is compiled against different PyTorch and CUDA versions.

By the way, not everybody knows the development mode. By setting the environment variable, we can have several versions of icefall in a single environment.

pzelasko:
I see your point about k2. 

Maybe we should have a `path.sh` file in the main icefall directory that we can source? I think many people don't know about PYTHONPATH either.

```bash
#!/usr/bin/env bash
ICEFALL_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
export PYTHONPATH="$ICEFALL_ROOT:$PYTHONPATH"
```

pzelasko:
.. or just merge this with a caveat that it doesn't give a "complete" installation. It's convenient when trying to import methods from icefall in other projects (such as decoding) because it makes icefall visible the "standard" Python way, like any other package. If it's more convenient to set env vars for other applications, then this is still possible when providing setup.py.

csukuangfj:
> I see your point about k2.
> 
> Maybe we should have a `path.sh` file in the main icefall directory that we can source? I think many people don't know about PYTHONPATH either.
> 
> ```shell
> #!/usr/bin/env bash
> ICEFALL_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
> export PYTHONPATH="$ICEFALL_ROOT:$PYTHONPATH"
> ```

I have been doing this locally, though I am using the absolute path. Your approach is more general.

---

> .. or just merge this with a caveat that it doesn't give a "complete" installation. 

@danpovey What do you think?

danpovey:
i think merging the.change is fine.  ppl who know about pythonpath can use
that.

PR #65, opened 2021-10-03
WIP: Support torchscript.

csukuangfj:
See https://github.com/k2-fsa/k2/pull/839

### DIFF
--- a/egs/librispeech/ASR/conformer_ctc/export.py
+++ b/egs/librispeech/ASR/conformer_ctc/export.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python3
+#
+# Copyright 2021 Xiaomi Corporation (Author: Fangjun Kuang)
+#
+# See ../../../../LICENSE for clarification regarding multiple authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


egs/librispeech/ASR/conformer_ctc/export.py
```diff
@@ -0,0 +1,164 @@
...
+            if start >= 0:
+                filenames.append(f"{params.exp_dir}/epoch-{i}.pt")
+        logging.info(f"averaging {filenames}")
+        model.load_state_dict(average_checkpoints(filenames))
+
+    model.to("cpu")
+
+    if params.jit:
+        logging.info("Using torch.jit.script")
+        model = torch.jit.script(model)
```
danpovey, line 145: Mm, I imagine this somehow inspects the forward() function of the model?  Would we still be able to use functions like decoder_nll()

egs/librispeech/ASR/conformer_ctc/transformer.py
```diff
@@ -301,6 +302,7 @@ def decoder_forward(
 
         return decoder_loss
 
+    @torch.jit.export
     def decoder_nll(
```
csukuangfj, line 306: > Mm, I imagine this somehow inspects the forward() function of the model? Would we still be able to use functions like decoder_nll()

@danpovey 
As long as you use the decorator `@torch.jit.export`, you can access it.

egs/librispeech/ASR/conformer_ctc/export.py
csukuangfj: > Mm, I imagine this somehow inspects the forward() function of the model?

Yes, by default it inspects only the `forward()` method and any methods called in `forward()`.

>  Would we still be able to use functions like decoder_nll()

For methods that are not called either directly or indirectly in `forward()`, you can use `@torch.jit.export`
to decorate it so that it is treated similarly as `forward()`.

PR #72, opened 2021-10-08
[WIP]decoding and computing ppl with nnlm

glynpu:
This is a baseline for #54.

Performance on librispeech **test_clean**:
|wer%|ppl|
|--|--|
|2.38(baseline without nnlm 2.60)|89.71|

Above tested model is the released model by [espnet](https://zenodo.org/record/4604066#.YWA8S5pBxPY), whose keys in state_dict is modified to compatible with icefall model definition (i.e. the model definition is identical to espnet released model).

TODO list:
1. This pr **only** aims to implement the **decoding** related code. 
   Could you give me some reviews to make it qualified to merge.

2. training part will be added in another pr which may based on #54, since data preparation for training is re-usable.


egs/librispeech/ASR/conformer_ctc/utils/text_dataset.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env python3
+
```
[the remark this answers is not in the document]
csukuangfj: Yes, that sounds good.

egs/librispeech/ASR/conformer_ctc/compute_ppl.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python3
```
csukuangfj, line 1: Can we have a new directory, e.g., `conformer_ctc_lm`, and leave `conformer_ctc` as it is.

danpovey: Would it use the model trained in the conformer_ctc dir for ASR decoding?  And how about the modeling code-- would that just be copied?  Soft linked?

csukuangfj: How about copying the modeling code?
If the modeling code is copied, I think the two can share model checkpoints if needed.
Or do you have other suggestions?

egs/librispeech/ASR/conformer_ctc/utils/text_dataset.py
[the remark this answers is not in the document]
glynpu: > If you borrowed parts of this from ESPNet we should acknowledge that.




Currently, only the parameters are borrowed from ESPNet, (to make sure the model structure is identical to ESPNet and pretend we have a good trained model).

I think the code is not from ESPNet, because the LM is used in a different way with ESPNet.
In this pr, LM is used as a rescorer of a whole candidate sentence. 
While in ESPNet it is used in beam-search process during which P(Xt|Xo...Xt-1) is computed step by step. 

Anyway I will confirm this and add references in the code if needed.

egs/librispeech/ASR/conformer_ctc/compute_ppl.py
danpovey: I guess it's OK to copy.. seems a bit non-optimal to duplicate code, but I suppose it would keep the conformer_ctc code clean while we are still playing with this LM rescoring stuff, in case more changes are coming(?)

danpovey: RE where to put the scripts in the utils dir: we could possibly put them in a new directory ^/icefall/lm, but to do that we might want to look at them a bit more carefully to make sure they are well structured and well documented.  Would definitely need more documentation.  We can see what Fangjun thinks too.

icefall/decode.py
```diff
@@ -854,11 +855,21 @@ def rescore_with_attention_decoder(
...
+
     assert nll.ndim == 2
     assert nll.shape[0] == len(token_ids)
 
     attention_scores = -nll.sum(dim=1)
 
+    if nnlm_evaluator is not None:
+        aux_labels = k2.RaggedTensor(tokens_shape, nbest.fsa.aux_labels)
+        aux_labels = aux_labels.remove_values_leq(0)
+        aux_labels = aux_labels.tolist()
```
danpovey, line 867: Eventually it would be nice if we can make more use of ragged tensors and less of lists (more efficient).  But I'd say this is very low priority.

PR #95, opened 2021-10-27
WIP: Decoding scripts using conformer LM.

csukuangfj:
Decoding code for https://github.com/k2-fsa/icefall/pull/54

Still working in progress. Hope to get some results this week.

### DIFF
--- a/egs/librispeech/ASR/conformer_lm/dataset.py
+++ b/egs/librispeech/ASR/conformer_lm/dataset.py
@@ -33,7 +33,13 @@ class LmDataset(torch.utils.data.Dataset):
         # self.sentences[i] returns a 1-D tensor containing word indexes
         # self.words[self.sentences[i]] returns a ragged tensor with axes
         # [word][token].
-        return self.words[self.sentences[i]].values.tolist()
+        word_tokens =  self.words[self.sentences[i]]
+        # TODO(fangjun): Need to figure out why `word_tokens`
+        # can be a torch.Tensor
+        if isinstance(word_tokens, torch.Tensor):
+            return word_tokens
+        else:
+            return word_tokens.values.tolist()
 
 
 def load_train_test_lm_dataset(archive_fn: Union[str,Path],
--- a/egs/librispeech/ASR/conformer_lm/train.py
+++ b/egs/librispeech/ASR/conformer_lm/train.py
@@ -74,6 +74,44 @@ def get_parser():
         help="Should various information be logged in tensorboard.",
     )
 
+    parser.add_argument(
+        "--exp-dir",
+        type=str,
+        default="conformer_lm/exp",
+        help="Path to save files generated during training",
+    )
+
+    parser.add_argument(
+        "--lm-dataset",
+        type=str,
+        default="data/lm_training_500/lm_data.pt",
+        help="LM training data. See local/prepare_lm_training_data.py",
+    )
+
+    parser.add_argument(
+        "--num-tokens",
+        type=int,
+        default="500",
+        help="BPE model vocab size",
+    )
+
+    parser.add_argument(
+        "--num-epochs",
+        type=int,
+        default=30,
+        help="Number of epochs to train.",
+    )
+
+    parser.add_argument(
+        "--start-epoch",
+        type=int,
+        default=0,
+        help="""Resume training from from this epoch.
+        If it is positive, it will load checkpoint from
+        exp_dir/epoch-{start_epoch-1}.pt
+        """,
+    )
+
     return parser
 
 
@@ -88,18 +126,10 @@ def get_params() -> AttributeDict:
 
     Explanation of options saved in `params`:
 
-        - exp_dir: It specifies the directory where all training related
-                   files, e.g., checkpoints, log, etc, are saved
-
         - lr: It specifies the initial learning rate
 
         - feature_dim: The model input dim. It has to match the one used
                        in computing features.
-
-        - start_epoch:  If it is not zero, load checkpoint `start_epoch-1`
-                        and continue training from that checkpoint.
-
-        - num_epochs:  Number of epochs to train.
 
         - num_valid_batches:  Number of batches of validation data to use each
                         time we compute validation loss
@@ -132,19 +162,12 @@ def get_params() -> AttributeDict:
     """
     params = AttributeDict(
         {
-            # exp_3, vs. exp_2, is using 5e-04 not 2d-04 as max learning rate.
-            # exp_4, vs. exp_3, is using the Gloam optimizer with
-            # in exp_5, vs. exp_4, we change Gloam to have a 1/sqrt(t) factor
-            #      as well as the exponential part.
-            # exp_6, we change the decay from 0.85 to 0.9.
-            "exp_dir": Path("conformer_lm/exp_6"),
-            "lm_dataset": Path("data/lm_training_5000/lm_data.pt"),
-            "num_tokens": 5000,
+            #  "exp_dir": Path("conformer_lm/exp_6"),
+            #  "lm_dataset": Path("data/lm_training_5000/lm_data.pt"),
+            #  "num_tokens": 5000,
             "blank_sym": 0,
             "bos_sym": 1,
             "eos_sym": 1,
-            "start_epoch": 2,
-            "num_epochs": 20,
             "num_valid_batches": 200,
             "symbols_per_batch": 5000,
             "best_train_loss": float("inf"),
@@ -152,7 +175,7 @@ def get_params() -> AttributeDict:
             "best_train_epoch": -1,
             "best_valid_epoch": -1,
             "batch_idx_train": 0,
-            "log_interval": 10,
+            "log_interval": 50,
             "reset_interval": 200,
             "valid_interval": 3000,
             "beam_size": 10,
@@ -607,6 +630,8 @@ def run(rank, world_size, args):
 def main():
     parser = get_parser()
     args = parser.parse_args()
+    args.exp_dir = Path(args.exp_dir)
+    args.lm_dataset = Path(args.lm_dataset)
 
     world_size = args.world_size
     assert world_size >= 1


csukuangfj:
Here is the progress so far:

(1) Generate `masked_src`, `src`, `tgt`, `src_key_padding_mask`, and `weight` for the conformer lm model from
an nbest object

(2) Able to load the pre-trained conformer lm model into decode.py

Will add another option to decode.py to use conformer lm for rescoring.

----

Here is an example of generating the inputs for conformer lm from an nbest object.

Assume we have two utterances: utterance 0 has two paths and utterance 1 has 3 paths:
```python3
    # utterance 0
    path00 = k2.linear_fsa([1, 2, 0, 3, 0, 5])
    path01 = k2.linear_fsa([1, 0, 5, 0])

    # utterance 1
    path10 = k2.linear_fsa([9, 8, 0, 3, 0, 2])
    path11 = k2.linear_fsa([9, 8, 0, 0, 3, 2])
    path12 = k2.linear_fsa([9, 0, 8, 4, 0, 2, 3])

    shape = k2.RaggedShape("[[x x] [x x x]]")
```

The resulting inputs for conformer lm are:
(Assume bos_id=10, eos_id=20, blank_id=0, unmasked_weight=0.25.
For each pair of paths in an utterance, when one is used as a src, the other is used as a tgt.
)

### masked src
```
 tensor([[10,  1,  2,  3,  5, 20,  0],
        [10,  1,  0,  0,  5, 20,  0],
        [10,  1,  0,  0,  5, 20,  0],
        [10,  1,  5, 20,  0,  0,  0],
        [10,  9,  8,  3,  2, 20,  0],
        [10,  9,  8,  3,  2, 20,  0],
        [10,  9,  8,  0,  2,  0, 20],
        [10,  9,  8,  3,  2, 20,  0],
        [10,  9,  8,  3,  2, 20,  0],
        [10,  9,  8,  0,  2,  0, 20],
        [10,  9,  8,  0,  2,  0, 20],
        [10,  9,  8,  0,  2,  0, 20],
        [10,  9,  8,  4,  2,  3, 20]], dtype=torch.int32)
```

### src
```
 tensor([[10,  1,  2,  3,  5, 20,  0],
        [10,  1,  0,  0,  5, 20,  0],
        [10,  1,  2,  3,  5, 20,  0],
        [10,  1,  5, 20,  0,  0,  0],
        [10,  9,  8,  3,  2, 20,  0],
        [10,  9,  8,  3,  2, 20,  0],
        [10,  9,  8,  4,  2,  3, 20],
        [10,  9,  8,  3,  2, 20,  0],
        [10,  9,  8,  3,  2, 20,  0],
        [10,  9,  8,  4,  2,  3, 20],
        [10,  9,  8,  3,  2,  0, 20],
        [10,  9,  8,  3,  2,  0, 20],
        [10,  9,  8,  4,  2,  3, 20]], dtype=torch.int32)
```

### tgt
```
 tensor([[ 1,  2,  3,  5, 20,  0,  0],
        [ 1,  2,  3,  5, 20,  0,  0],
        [ 1,  0,  0,  5, 20,  0,  0],
        [ 1,  5, 20,  0,  0,  0,  0],
        [ 9,  8,  3,  2, 20,  0,  0],
        [ 9,  8,  3,  2, 20,  0,  0],
        [ 9,  8,  3,  2,  0, 20,  0],
        [ 9,  8,  3,  2, 20,  0,  0],
        [ 9,  8,  3,  2, 20,  0,  0],
        [ 9,  8,  3,  2,  0, 20,  0],
        [ 9,  8,  4,  2,  3, 20,  0],
        [ 9,  8,  4,  2,  3, 20,  0],
        [ 9,  8,  4,  2,  3, 20,  0]], dtype=torch.int32)
```

### src_key_padding_mask
```
 tensor([[False, False, False, False, False, False,  True],
        [False, False, False, False, False, False,  True],
        [False, False, False, False, False, False,  True],
        [False, False, False, False,  True,  True,  True],
        [False, False, False, False, False, False,  True],
        [False, False, False, False, False, False,  True],
        [False, False, False, False, False, False, False],
        [False, False, False, False, False, False,  True],
        [False, False, False, False, False, False,  True],
        [False, False, False, False, False, False, False],
        [False, False, False, False, False, False, False],
        [False, False, False, False, False, False, False],
        [False, False, False, False, False, False, False]])
```

### weight
```
 tensor([[0.2500, 0.2500, 0.2500, 0.2500, 0.2500, 0.0000, 0.0000],
        [0.2500, 1.0000, 1.0000, 0.2500, 0.2500, 0.0000, 0.0000],
        [0.2500, 1.0000, 1.0000, 0.2500, 0.2500, 0.0000, 0.0000],
        [0.2500, 0.2500, 0.2500, 0.0000, 0.0000, 0.0000, 0.0000],
        [0.2500, 0.2500, 0.2500, 0.2500, 0.2500, 0.0000, 0.0000],
        [0.2500, 0.2500, 0.2500, 0.2500, 0.2500, 0.0000, 0.0000],
        [0.2500, 0.2500, 1.0000, 0.2500, 1.0000, 0.2500, 0.0000],
        [0.2500, 0.2500, 0.2500, 0.2500, 0.2500, 0.0000, 0.0000],
        [0.2500, 0.2500, 0.2500, 0.2500, 0.2500, 0.0000, 0.0000],
        [0.2500, 0.2500, 1.0000, 0.2500, 1.0000, 0.2500, 0.0000],
        [0.2500, 0.2500, 1.0000, 0.2500, 1.0000, 0.2500, 0.0000],
        [0.2500, 0.2500, 1.0000, 0.2500, 1.0000, 0.2500, 0.0000],
        [0.2500, 0.2500, 0.2500, 0.2500, 0.2500, 0.2500, 0.0000]])
```

danpovey:
For the weights, here, we only actually care about the weights in the masked regions, so those 0.25's can be set to 0.  They were introduced for training purposes (the task is nontrivial only because we randomize the symbols in the non-masked regions, with some small probability).

csukuangfj:
Thanks, will change it.

csukuangfj:
I am able to use a masked conformer lm model to rescore a Nbest object and combine the resulting score with
attention score and acoustic score.

Will do some tests to see how it affects the WER.

---

By the way, the `conformer_ctc/decode.py` and `icefall/decode.py` become increasingly complicated and
more difficult to understand now.  I think we need another refactoring for the decoding scripts.

icefall/decode.py
```diff
@@ -889,3 +895,198 @@ def rescore_with_attention_decoder(
...
+        )
+    else:
+        expanded_memory_key_padding_mask = None
+
+    # remove axis corresponding to states.
+    tokens_shape = nbest.fsa.arcs.shape().remove_axis(1)
+    tokens = k2.RaggedTensor(tokens_shape, nbest.fsa.tokens)
+    tokens = tokens.remove_values_leq(0)
+
+    alignment = compute_alignment(tokens, nbest.shape)
```
csukuangfj, line 990: Code for conformer LM rescoring starts here.

icefall/lm/rescore.py
```diff
@@ -0,0 +1,392 @@
...
+# limitations under the License.
+
+"""
+This file contains rescoring code for NN LMs, e.g., conformer LM.
+
+Support an utterance has 3 paths:
+    (a, b, c)
+and we want to use a masked conformer LM to assign a likelihood to each path.
+
+The following shows the steps:
```
csukuangfj, line 24: @danpovey 
This shows how the masked LM model is used to assign a score to each path.

icefall/decode.py
```diff
@@ -889,3 +897,225 @@ def rescore_with_attention_decoder(
...
+                src_key_padding_mask,
+            )
+
+            # nll means negative log-likelihood
+            # ll means log-likelihood
+            tgt_ll = -1 * (tgt_nll * tgt_weights).sum(dim=-1)
+
+            tgt_ll_list.append(tgt_ll)
+
+        #  tgt_ll = tgt_ll_list[1] - tgt_ll_list[0] # wer: 2.61
```
danpovey, line 1046: This is a little suspicious to me; seems to me it should be hyp - ref = [1] - [0].  If the sign is wrong I think we should determine where.

csukuangfj, line 1046: Suppose there is only one utterance and there are three paths. Let us call the three paths "a", "b", "c".

`tgt_ll_list[0]` contains
- "ab_self", use "a" as src and its shifted version as "tgt", the alignment is computed using the pair (a, b), where a is used as a reference and b as hypothesis
- "ac_self"
- "ba_self"
- "bc_self"
- "ca_self"
- "cb_self"

That is `[ab_self, ac_self, ba_self, bc_self, ca_self, cb_self]`



`tgt_ll_list[1]` contains
`[ab_other, ac_other, ba_other, bc_other, ca_other, cb_other]`

`tgt_ll = tgt_ll_list[0] - tgt_ll_list[1]`, which is
```
[ab_self - ab_other, ac_self - ac_other, ba_self - ba_other, bc_self - bc_other, ca_self - ca_other, cb_self - cb_other]
```

`ragged_tgt_ll` is
```
[ [ab_self - ab_other, ac_self - ac_other], 
  [ba_self - ba_other, bc_self - bc_other], 
  [ca_self - ca_other, cb_self - cb_other]
]
```

`ab_other - ab_self` can be interpreted as 
```
log p(b) - log p (a)
```
But I found that `ab_self - ab_other` has a lower WER. @danpovey

icefall/lm/rescore.py
```diff
@@ -0,0 +1,392 @@
...
+    (i) Compute the alignment between "a" and "b"
+    (ii) Use the computed alignment as `masked_src,`
+    (iii) Use "a" as "src" and its shifted version as "tgt" (of course, we need
+          to add bos and eos) and we can get a log-likelihood value (after
+          negating the negative log-likelihood). Let us call this value as
+          "ab_self"
+    (iv)  Use "b" as "src" and its shifted version as "tgt".
+          We can get another likelihood value, denoted as "ab_other"
+
+So for the path pair (a, b), (a, c), we can get the following log-likelihood
```
csukuangfj, line 42:
```suggestion
So for the path pair (a, b), (a, c), (b, a), (b, c), (c, a), and (c, b), we can get the following log-likelihood
```

icefall/decode.py
```diff
@@ -889,3 +893,206 @@ def rescore_with_attention_decoder(
...
+    if attention_scale is None:
+        attention_scale_list = [0.01, 0.05, 0.08]
+        attention_scale_list += [0.1, 0.3, 0.5, 0.6, 0.7, 0.9, 1.0]
+        attention_scale_list += [1.1, 1.2, 1.3, 1.5, 1.7, 1.9, 2.0]
+        attention_scale_list += [2.1, 2.2, 2.3, 2.5, 3.0, 4.0, 5.0]
+    else:
+        attention_scale_list = [attention_scale]
+
+    if masked_lm_scale is None:
+        masked_lm_scale_list = [0.01, 0.05, 0.08]
```
danpovey, line 1075: Perhaps, now that there are so many scales, we should start pruning the scales to the more frequently seen values.

icefall/lm/rescore.py
```diff
@@ -0,0 +1,338 @@
...
+        )
+        map_tensor_list.append(map_tensor)
+
+    return torch.cat(map_tensor_list)
+
+
+def compute_alignment(
+    tokens: k2.RaggedTensor,
+    shape: k2.RaggedShape,
+) -> k2.Fsa:
```
danpovey, line 225: I think it would be especially helpful to have some documentation for this, esp. about the input and output formats.  Would be easier to follow what the function is doing.  Same with make_hyp_to_ref_map.

icefall/decode.py
csukuangfj: Yes, I agree. It takes hours to compute the final WERs with so many scales.

PR #100, opened 2021-11-01
giga speech dataloader

glynpu:
Hi Dan, @danpovey 
Is this pr what you want about gigaspeech dataloader?

Example usage:
```
import argparse                                                                 
from gigaspeech_datamodule import GigaSpeechAsrDataModule                       
                                                                                
def get_parser():                                                               
    parser = argparse.ArgumentParser(                                           
        formatter_class=argparse.ArgumentDefaultsHelpFormatter                  
    )                                                                           
    return parser                                                               
                                                                                
parser = get_parser()                                                           
GigaSpeechAsrDataModule.add_arguments(parser)                                   
args = parser.parse_args()                                                      
gigaspeech = GigaSpeechAsrDataModule(args)                                      
train_dl = gigaspeech.train_dataloaders()                                       
valid_dl = gigaspeech.valid_dataloaders()                                       
batch=next(iter(train_dl))                                                      
print(batch["supervisions"]["text"][0:5])                                       
print(batch["inputs"].shape)                                                    
batch=next(iter(train_dl))                                                      
print(batch["supervisions"]["text"][0:5])                                       
print(batch["inputs"].shape)


```

Expected output:

```
['HERE ARE JUST A FEW OF THE SEEMINGLY UNREAL <COMMA> ALBEIT <COMMA> ENTIRELY TRUE HIGHLIGHTS FROM HIS RESUME <PERIOD>', "HE'S A FORMER MTV BREAK DANCER IN TAIWAN <COMMA>", 'THE ONE TIME NATIONAL CHINESE KICKBOXING CHAMPION <COMMA>', 'THE FIRST AMERICAN TO HOLD THE GUINNESS WORLD RECORD IN TANGO <COMMA> A HORSEBACK ARCHER <COMMA> PRINCETON UNIVERSITY GUEST LECTURER <COMMA> AND ANGEL INVESTOR WHO HAS RACKED UP A SERIOUS STRING OF ENTREPRENEURIAL MAGA HITS <COMMA>', 'INCLUDING UBER <COMMA> FACEBOOK <COMMA> TWITTER <COMMA> AND ALIBABA <PERIOD>']
torch.Size([1, 274823, 80])
['HERE ARE JUST A FEW OF THE SEEMINGLY UNREAL <COMMA> ALBEIT <COMMA> ENTIRELY TRUE HIGHLIGHTS FROM HIS RESUME <PERIOD>', "HE'S A FORMER MTV BREAK DANCER IN TAIWAN <COMMA>", 'THE ONE TIME NATIONAL CHINESE KICKBOXING CHAMPION <COMMA>', 'THE FIRST AMERICAN TO HOLD THE GUINNESS WORLD RECORD IN TANGO <COMMA> A HORSEBACK ARCHER <COMMA> PRINCETON UNIVERSITY GUEST LECTURER <COMMA> AND ANGEL INVESTOR WHO HAS RACKED UP A SERIOUS STRING OF ENTREPRENEURIAL MAGA HITS <COMMA>', 'INCLUDING UBER <COMMA> FACEBOOK <COMMA> TWITTER <COMMA> AND ALIBABA <PERIOD>']
torch.Size([1, 274823, 80])

```

### DIFF
--- a/egs/librispeech/ASR/conformer_ctc/gigaspeech_datamodule.py
+++ b/egs/librispeech/ASR/conformer_ctc/gigaspeech_datamodule.py
@@ -1,0 +1,335 @@
+#!/usr/bin/env python3
+# Copyright    2021  Xiaomi Corp.        (authors: Liyong Guo)
+#
+# See ../../../../LICENSE for clarification regarding multiple authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import argparse
+import logging
+from functools import lru_cache
+from pathlib import Path
+from typing import List, Union
+
+from lhotse import CutSet, Fbank, FbankConfig, load_manifest
+from lhotse.dataset import (
+    BucketingSampler,
+    CutConcatenate,
+    CutMix,
+    K2SpeechRecognitionDataset,
+    PrecomputedFeatures,
+    SingleCutSampler,
+    SpecAugment,
+)
+from lhotse.dataset.input_strategies import OnTheFlyFeatures
+from torch.utils.data import DataLoader
+
+from icefall.dataset.datamodule import DataModule
+from icefall.utils import str2bool
+
+
+class GigaSpeechAsrDataModule(DataModule):
+    """
+    DataModule for k2 ASR experiments.
+    It assumes there is always one train and valid dataloader,
+
+    It contains all the common data pipeline modules used in ASR
+    experiments, e.g.:
+    - dynamic batch size,
+    - bucketing samplers,
+    - cut concatenation,
+    - augmentation,
+    - on-the-fly feature extraction
+
+    This class should be derived for specific corpora used in ASR tasks.
+    """
+
+    @classmethod
+    def add_arguments(cls, parser: argparse.ArgumentParser):
+        super().add_arguments(parser)
+        group = parser.add_argument_group(
+            title="ASR data related options",
+            description="These options are used for the preparation of "
+            "PyTorch DataLoaders from Lhotse CutSet's -- they control the "
+            "effective batch sizes, sampling strategies, applied data "
+            "augmentations, etc.",
+        )
+        group.add_argument(
+            "--full-giga",
+            type=str2bool,
+            default=False,
+            help="When enabled, use XL part of GigaSpeech. "
+            "Otherwise, use XS subset.",
+        )
+        group.add_argument(
+            "--feature-dir",
+            type=Path,
+            default=Path("data/fbank"),
+            help="Path to directory with train/valid/test cuts.",
+        )
+        group.add_argument(
+            "--max-duration",
+            type=int,
+            default=200.0,
+            help="Maximum pooled recordings duration (seconds) in a "
+            "single batch. You can reduce it if it causes CUDA OOM.",
+        )
+        group.add_argument(
+            "--bucketing-sampler",
+            type=str2bool,
+            default=True,
+            help="When enabled, the batches will come from buckets of "
+            "similar duration (saves padding frames).",
+        )
+        group.add_argument(
+            "--num-buckets",
+            type=int,
+            default=30,
+            help="The number of buckets for the BucketingSampler"
+            "(you might want to increase it for larger datasets).",
+        )
+        group.add_argument(
+            "--concatenate-cuts",
+            type=str2bool,
+            default=False,
+            help="When enabled, utterances (cuts) will be concatenated "
+            "to minimize the amount of padding.",
+        )
+        group.add_argument(
+            "--duration-factor",
+            type=float,
+            default=1.0,
+            help="Determines the maximum duration of a concatenated cut "
+            "relative to the duration of the longest cut in a batch.",
+        )
+        group.add_argument(
+            "--gap",
+            type=float,
+            default=1.0,
+            help="The amount of padding (in seconds) inserted between "
+            "concatenated cuts. This padding is filled with noise when "
+            "noise augmentation is used.",
+        )
+        group.add_argument(
+            "--on-the-fly-feats",
+            type=str2bool,
+            default=False,
+            help="When enabled, use on-the-fly cut mixing and feature "
+            "extraction. Will drop existing precomputed feature manifests "
+            "if available.",
+        )
+        group.add_argument(
+            "--shuffle",
+            type=str2bool,
+            default=True,
+            help="When enabled (=default), the examples will be "
+            "shuffled for each epoch.",
+        )
+        group.add_argument(
+            "--return-cuts",
+            type=str2bool,
+            default=True,
+            help="When enabled, each batch will have the "
+            "field: batch['supervisions']['cut'] with the cuts that "
+            "were used to construct it.",
+        )
+
+        group.add_argument(
+            "--num-workers",
+            type=int,
+            default=2,
+            help="The number of training dataloader workers that "
+            "collect the batches.",
+        )
+
+    def train_dataloaders(self) -> DataLoader:
+        logging.info("About to get train cuts")
+        cuts_train = self.train_cuts()
+
+        logging.info("About to create train dataset")
+        transforms = None
+        if self.args.concatenate_cuts:
+            logging.info(
+                f"Using cut concatenation with duration factor "
+                f"{self.args.duration_factor} and gap {self.args.gap}."
+            )
+            # Cut concatenation should be the first transform in the list,
+            # so that if we e.g. mix noise in, it will fill the gaps between
+            # different utterances.
+            transforms = [
+                CutConcatenate(
+                    duration_factor=self.args.duration_factor, gap=self.args.gap
+                )
+            ]
+
+        train = K2SpeechRecognitionDataset(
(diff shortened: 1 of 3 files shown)


danpovey:
Great-- thanks!

pzelasko:
I think you'd do best to copy my code from snowfall for GigaSpeech; there's a number of details that need to be addressed before you can actually run training, e.g. splitting into smaller utterances, lazy-reading the CutSet, etc.

danpovey:
OK.
In this case we actually won't be using the transcripts, as we'll be using it for unsupervised.  We don't really need the cut information either, at least I wasn't planning to use it.  I guess we can drop that somehow?

danpovey:
Hm, maybe it's better that we do what Piotr said.  Would you mind figuring this part out, Liyong?
The tricky thing is that we are co-training with Librispeech, so the epochs will be determined by the size of the Librispeech training data.  We should probably use the "L" or "XL" subsets, which are larger than Librispeech, although of course it's OK to test with a smaller subset.  What i am thinking we can do about the data loaders, is just recreate the Gigaspeech dataloaders whenever they run out of data.  This doesn't even have to be exactly synchronized between the training processes.

danpovey:
.. and it's OK with me to use the same cut information that we'd use if we were using the supervisions, although I don't plan to actually use the text.

glynpu:
Make a copy from Piotr's implementation with following modification:
1.  change snowfall depencies with icefall
2. change some interface to make compatible to latest lhotse
3. Add an interface called "inexhaustible_train_dataloaders"
4. remove specaugment transforms as required

>recreate the Gigaspeech dataloaders whenever they run out of data

The added "inexhaustible_train_dataloaders" will do this.

example usage(It will never stop since the dataloader is inexhaustible):
```
import argparse                                                                 
from gigaspeech_datamodule import GigaSpeechAsrDataModule                       
                                                                                
def get_parser():                                                               
    parser = argparse.ArgumentParser(                                           
        formatter_class=argparse.ArgumentDefaultsHelpFormatter                  
    )                                                                           
    return parser                                                               
                                                                                
if __name__ == '__main__':                                                      
    parser = get_parser()                                                       
    GigaSpeechAsrDataModule.add_arguments(parser)                               
    args = parser.parse_args()                                                  
    gigaspeech = GigaSpeechAsrDataModule(args)                                  
    train_dl = gigaspeech.inexhaustible_train_dataloaders()                     
    for idx, batch in enumerate(train_dl):                                      
        print(batch["inputs"].shape)                                            
        print(len(batch["supervisions"]["text"]))                               
        print(batch["supervisions"]["text"][0:2])
```

glynpu:
Now option conflicts between librispeech and gigaspeech (e.g. they both have --max-duration) are solved in a quite ugly way.

example usage:
```
python example_giga_dataloader.py --max-duration 10  giga --max-duration=150 --feature-dir="exp/data/"
```
The former --maxduration 10 is for librispeech; 
options after **_giga_** are for gigaspeech.
I believe there exits an elegant way for this problem
.

egs/librispeech/ASR/conformer_ctc/gigaspeech_datamodule.py
```diff
@@ -0,0 +1,403 @@
...
+        test_loaders = []
+        if not is_list:
+            cuts = [cuts]
+
+        for cuts_test in cuts:
+            logging.debug("About to create test dataset")
+            test = K2SpeechRecognitionDataset(
+                cuts_test,
+                input_strategy=(
+                    OnTheFlyFeatures(Fbank(FbankConfig(num_mel_bins=80)), num_workers=8)
```
pzelasko, line 337: Note: replacing this with `KaldifeatFbank` in latest Lhotse may speed up on-the-fly feature extraction by 20~30% (on CPU).

egs/librispeech/ASR/prepare_gigaspeech.py
```diff
@@ -0,0 +1,293 @@
...
+    sup: SupervisionSegment, oov_pattern=re.compile(r"<(SIL|MUSIC|NOISE|OTHER)>")
+) -> bool:
+    return oov_pattern.search(sup.text) is None
+
+
+def main():
+    args = get_parser().parse_args()
+    dataset_parts = [args.subset, "DEV", "TEST"]
+    if args.subset in ["L", "XL"]:
+        assert is_module_available("pyarrow"), (
```
pzelasko, line 143: This doesn't look right to me; `pyarrow` is long gone from Lhotse, which version of the recipe is that? I don't see this code in https://github.com/k2-fsa/snowfall/blob/master/egs/gigaspeech/asr/simple_v1/prepare.py

egs/librispeech/ASR/prepare_gigaspeech.py
```diff
@@ -0,0 +1,293 @@
...
+    )
+
+    ctx_suffix = get_context_suffix(args)
+
+    print("Feature extraction:")
+    extractor = Fbank(FbankConfig(num_mel_bins=80))
+    with get_executor() as ex:  # Initialize the executor only once.
+        for partition, manifests in gigaspeech_manifests.items():
+            # For L and XL partition we are going to store the manifest using pyarrow.
+            cuts_path_ext = "jsonl.gz" if partition not in ["L", "XL"] else "arrow"
```
pzelasko, line 184: This is also from an older version, please try to replace this with the recipe code on current master in snowfall.

BTW I originally skipped feature extraction for L and XL because it was too slow and took too much memory. But since then @csukuangfj added chunked feature extraction in kaldifeat, and I implemented CUDA extraction in Lhotse. You can safely compute features for everything with the new mechanisms. See the following method on CutSet with examples:

https://github.com/lhotse-speech/lhotse/blob/ce1437c8127d3545f7c45a5677c44bbb5b8d4f35/lhotse/cut.py#L3358-L3410

egs/librispeech/ASR/conformer_ctc/gigaspeech_datamodule.py
```diff
@@ -0,0 +1,403 @@
...
+            type=str2bool,
+            default=False,
+            help='Should we use only 1000 utterances for dev (speeds up training)'
+        )
+
+    def validate_args(self):
+        if self.args.giga_subset in ['L', 'XL']:
+            assert (
+                self.args.giga_shuffle == False
+            ), "For GigaSpeech L/XL, you must use --shuffle 0 to avoid eagerly reading pyarrow manifests."
```
pzelasko, line 184: As mentioned in other comment -- this file also seems to come from an older version of the recipe.

egs/librispeech/ASR/prepare_gigaspeech.py
glynpu: Will modify it to the master one.

I directly copy the file from this link https://github.com/k2-fsa/snowfall/pull/230/files 
The website defaultly show "the pyarrow version", which exactly not the same with the one in the master branch.

glynpu: If you search pyarrow from previous link, you would got the "pyarrow" version prepare.py .
https://github.com/k2-fsa/snowfall/pull/230/files#diff-10ff23c2be6cdc66d1461d8ff40b8acee607ab6c2e5d734c7ce03b65765289c3R143

I always thought github would show the latest version, but it seems not so.

pzelasko: I think I made follow-up PRs where I synchronized the changes with Lhotse and also improved GigaSpeech recipe.

glynpu: I know what happened; 
I was told that https://github.com/k2-fsa/snowfall/pull/230 is Piotr's implementation about gigaspeech.

But some files related are also modified in pr https://github.com/k2-fsa/snowfall/pull/238

That's why the older version is refered.

egs/librispeech/ASR/prepare_gigaspeech.py
glynpu: >took too much memory

How is this problem solved?
I think originally feats are already stored in HDF5 format.

glynpu: For gigaspeech, do you have any recommand values of num_workers and batch_duration when using compute_and_store_features_batch?
Now the default values are used, i.e. num_workers=4 and batch_duration=600.0

egs/librispeech/ASR/conformer_ctc/gigaspeech_datamodule.py
glynpu: This part is the same with snowfall master.  https://github.com/k2-fsa/snowfall/blob/master/egs/gigaspeech/asr/simple_v1/asr_datamodule.py#L183

Shall I removed the check of shuffle==False and bucketing_sampler==False? (I already removed the check of on_the_fly_feats)

egs/librispeech/ASR/prepare_gigaspeech.py
pzelasko: > > took too much memory
> 
> How is this problem solved? I think originally feats are already stored in HDF5 format.

It’s about the RAM memory, not disk usage. Basically for 1h+ waveforms naive feature extraction implementations use quadratic memory. Kaldifeat has a mechanism to prevent that. But anyway without CUDA it would have taken too much time to compute feats for 10k hours.

> For gigaspeech, do you have any recommand values of num_workers and batch_duration when using compute_and_store_features_batch?
Now the default values are used, i.e. num_workers=4 and batch_duration=600.0

It should work fine with defaults; you can try increasing these but I found that on some systems it generates too much I/O if you create too many workers and the DataLoader crashes.

egs/librispeech/ASR/conformer_ctc/gigaspeech_datamodule.py
pzelasko: You can remove the shuffle==False check (I added streaming pseudo shuffling in Lhotse) but I’m not sure about the BucketingSampler — I think I still haven’t added support for very large data for it (it will take long time, a few minutes maybe, to start the script if you turn it on). I can try to add it soon.

PR #111, opened 2021-11-09
Fix a bug in Nbest.compute_am_scores and Nbest.compute_lm_scores.

csukuangfj:
It does not affect the WER since `Nbest.fsa.scores` is not used after invoking `Nbest.compute_am_scores()`, but the implementation is incorrect. This PR fixes that.


csukuangfj:
By the way, the implementation of `compute_am_scores()` and `compute_lm_scores()` can be simplified in that FSAs in an Nbest object are linear in structure and we can just compute the sum of the scores of every arc of an FSA to get the `tot_scores`.

PR #118, opened 2021-11-11
Fix computing WERs for empty hypotheses

csukuangfj:


egs/librispeech/ASR/conformer_ctc/decode.py
```diff
@@ -467,16 +466,29 @@ def decode_dataset(
...
             this_batch = []
-            assert len(hyps) == len(texts)
-            for hyp_words, ref_text in zip(hyps, texts):
+            hyp_words = []
+            for ref_text in texts:
                 ref_words = ref_text.split()
                 this_batch.append((ref_words, hyp_words))
 
-            results[lm_scale].extend(this_batch)
+            for lm_scale in results.keys():
```
csukuangfj, line 488: This PR assumes that it will not lead to an empty lattice at the very beginning.
If later there are empty lattices, we just set the hypothesis to an empty list
and reuse the `key` from the non-empty `results`.

PR #63, opened 2021-09-29
WIP: Support multi-node multi-GPU training.

csukuangfj:
Test with two machines, each with 3 GPUs.

## Machine 1

```bash
export CUDA_VISIBLE_DEVICES="0,1,2"
cd egs/librispeech/ASR
./conformer_ctc/run-multi-node-multi-gpu.sh  --master-addr 10.177.74.141 --master-port 12356 --node-rank 0  --num-nodes 2 --full-libri 0
```

<details>
<summary>Click to view the log</summary>

```
CUDA_VISIBLE_DEVICES: 0,1,2
num_gpus: 3
master_addr: 10.177.74.141
+ python -m torch.distributed.launch --use_env --nproc_per_node 3 --nnodes 2 --node_rank 0 --master_addr 10.177.74.141 --master_port 12356 /ceph-fj/fa
ngjun/open-source/icefall-ctc-decoding/egs/librispeech/ASR/conformer_ctc/train.py --use-multi-node true --master-port 12356 --max-duration 200 --bucke
ting-sampler 1 --full-libri 0 --start-epoch 0 --num-epochs 50 --exp-dir conformer_ctc/exp2 --lang-dir data/lang_bpe_500
*****************************************
Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the varia
ble for optimal performance in your application as needed.
*****************************************
rank: 0, world_size: 6
rank: 2, world_size: 6
rank: 0, world_size: 6, local_rank: 0
rank: 2, world_size: 6, local_rank: 2
rank: 1, world_size: 6
rank: 1, world_size: 6, local_rank: 1
2021-09-30 01:04:30,684 INFO [train.py:659] (2/6) Training started
2021-09-30 01:04:30,684 INFO [train.py:660] (2/6) {'exp_dir': PosixPath('conformer_ctc/exp2'), 'lang_dir': PosixPath('data/lang_bpe_500'), 'best_train
_loss': inf, 'best_valid_loss': inf, 'best_train_epoch': -1, 'best_valid_epoch': -1, 'batch_idx_train': 0, 'log_interval': 10, 'reset_interval': 200,
'valid_interval': 3000, 'feature_dim': 80, 'subsampling_factor': 4, 'use_feat_batchnorm': True, 'attention_dim': 512, 'nhead': 8, 'num_decoder_layers'
[...]

Issue #35, opened 2021-09-04
logging not working for ddp training


pkufool:
Seems that logging module is not working when training with ddp for some particular pytorch version(I tried three versions, it works for torch 1.7.2, but not for 1.8.0, 1.8.1). 

I searched for a while and did not figure out how to fix it.

FYI.

csukuangfj:
Please see 
https://github.com/k2-fsa/snowfall/pull/158#issuecomment-886193986

We have encountered such a problem before. I would recommend you to switch to torch 1.7.1

pkufool:
Thanks, I did not notice the issue in snowfall.
I already switched to torch 1.7, just make others know.

csukuangfj:
I think if you change `logging.info` to `logging.warn`, you will see the log.

This works for me when using `torch.distributed.launch` for multi-node multi-gpu training, with torch==1.9.0

PR #120, opened 2021-11-14
GigaSpeech recipe

wgb14:
Features:
  - support BPE based lang
  - ~on-the-fly feature extraction~ chunked feature extraction with GPU by default

TODO:
 - [x] support phone based lang
   - [ ] ~cmudict2lexicon~
   - [ ] ~g2p~
   - [x] download pretrained lexicon and LM
 - [ ] ~language model pruning~
 - [x] debugging and running experiments


egs/gigaspeech/ASR/local/compute_fbank_gigaspeech.py
```diff
@@ -0,0 +1,248 @@
...
+                    # smaller cuts.
+                    # Note:
+                    # we support very efficient "chunked" feature reads with
+                    # the argument `storage_type=ChunkedLilcomHdf5Writer`,
+                    # but we don't support efficient data augmentation and
+                    # feature computation for long recordings yet.
+                    # Therefore, we sacrifice some storage for the ability to
+                    # precompute features on shorter chunks,
+                    # without memory blow-ups.
+                    cut_set = cut_set.compute_and_store_features(
```
csukuangfj, line 220: I thought Piotr was suggesting `compute_and_store_features_batch`
See https://github.com/lhotse-speech/lhotse/issues/452#issuecomment-961873479

pzelasko: Yes -- you'd do best to leverage the changes made by @glynpu here: https://github.com/k2-fsa/icefall/pull/100

wgb14: Does this method also work for XL set? And we no longer need on-the-fly feature extraction, nor pre-shuffling by default, right?

glynpu: I am testing the pre-computed-features interface with "L" and "XL" subset. As suggested by protr here https://github.com/lhotse-speech/lhotse/issues/452

pzelasko: This method should work for XL -- Lyong ran into some issues before but I suspect it was related to data prep, we should see soon.

You won't need on-the-fly extraction, but it still allows you to save a lot of disk space. You don't need pre-shuffling anymore though -- Lhotse samplers implement a streaming shuffle variant for lazily-opened CutSets (it works with zero code changes when `shuffle=True` is passed to a sampler).

wgb14: Here we do `compute_and_store_features_batch` if GPU card is detected, otherwise we go back to `compute_and_store_features`  or suggest `--precomputed-features False`.

wgb14: Also, I also got this errorwhile installing `kaldifeat` https://github.com/csukuangfj/kaldifeat/issues/12#issue-1030657358

csukuangfj: > I also got this errorwhile installing kaldifeat csukuangfj/kaldifeat#12 (comment)

So did you succeed in installing kaldifeat?

wgb14: > > I also got this errorwhile installing kaldifeat [csukuangfj/kaldifeat#12](https://github.com/csukuangfj/kaldifeat/issues/12) (comment)
> 
> So did you succeed in installing kaldifeat?

not yet. This error was resolved, but another error occurred:
```
...
    [ 50%] Linking CXX shared library ../../lib/libkaldifeat_core.so
    /usr/bin/ld: cannot find /export/c07/gwang50/miniconda3/envs/k2/lib: File format not recognized
    collect2: error: ld returned 1 exit status
    kaldifeat/csrc/CMakeFiles/kaldifeat_core.dir/build.make:223: recipe for target 'lib/libkaldifeat_core.so' failed
    make[3]: *** [lib/libkaldifeat_core.so] Error 1
    CMakeFiles/Makefile2:340: recipe for target 'kaldifeat/csrc/CMakeFiles/kaldifeat_core.dir/all' failed
    make[2]: *** [kaldifeat/csrc/CMakeFiles/kaldifeat_core.dir/all] Error 2
    CMakeFiles/Makefile2:427: recipe for target 'kaldifeat/python/csrc/CMakeFiles/_kaldifeat.dir/rule' failed
    make[1]: *** [kaldifeat/python/csrc/CMakeFiles/_kaldifeat.dir/rule] Error 2
    Makefile:225: recipe for target '_kaldifeat' failed
    make: *** [_kaldifeat] Error 2
...
```

csukuangfj: So it is the same error posted in
https://github.com/csukuangfj/kaldifeat/issues/15

Please see the discussions in the above issue.

If you cannot resolve it, I would recommend you to install it using the approach posted by Piotr in
https://github.com/csukuangfj/kaldifeat/issues/12#issue-1030657358

wgb14: I tried 
```
pip install kaldifeat
```
```
python setup.py install
```
and
```
cd build
cmake .. -DCMAKE_BUILD_TYPE=Release -DCUDNN_LIBRARY_PATH="/export/c07/gwang50/miniconda3/envs/k2/lib" -DCUDNN_INCLUDE_PATH="/export/c07/gwang50/miniconda3/envs/k2/include"
make  _kaldifeat
```
still got the same error.

wgb14: @pzelasko How did you install kaldifeat on CLSP grid? Actually I got 3 errors.
First is the version of `cmake` on grid is 3.7.2, and I installed a version of 3.22.0.
Second is `cannot find CUDNN`, resolved by specifying `DCUDNN_LIBRARY_PATH` and `DCUDNN_INCLUDE_PATH`.
Did you ever get the third error above`/usr/bin/ld: cannot find /export/c07/gwang50/miniconda3/envs/k2/lib: File format not recognized`?

csukuangfj: @wgb14 
Could you replace
```
-DCUDNN_LIBRARY_PATH="/export/c07/gwang50/miniconda3/envs/k2/lib"
```

with

```
-DCUDNN_LIBRARY_PATH="/export/c07/gwang50/miniconda3/envs/k2/lib/libcudnn.so"
```

I assume that `libcudnn.so` exists in `/export/c07/gwang50/miniconda3/envs/k2/lib`.
If not, please find the path to `libcudnn.so` and replace the above value.

Passing only the directory is not right, I think. We should pass the path to the `.so` file, not to its enclosing directory.

wgb14: cool, it really works. Thank you, Fangjun!

pzelasko: Cool! You're probably the first who installed it on the CLSP grid. If you could write down the steps in a gist, we could share it with our students...

wgb14: changed to chunked feature extraction  with GPU by default here, but got OOM error while running on grid
image
Any insights?

csukuangfj: Is it possible that the default value of --batch-duration, which is 600, is too large?

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech.py
```diff
@@ -134,78 +140,113 @@ def compute_fbank_gigaspeech(args):
...
+                context_direction=args.context_direction,
+            )
+
+            if args.precomputed_features:
+                # Extract the features after cutting large recordings into
+                # smaller cuts.
+                # Note:
+                # we support very efficient "chunked" feature reads with
+                # the argument `storage_type=ChunkedLilcomHdf5Writer`,
+                # but we don't support efficient data augmentation and
```
csukuangfj, line 222: >                 # but we don't support efficient data augmentation and
>                 # feature computation for long recordings yet.

Is it still the case with `kaldifeat`?

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech.py
```diff
@@ -134,78 +140,113 @@ def compute_fbank_gigaspeech(args):
...
+                # feature computation for long recordings yet.
+                # Therefore, we sacrifice some storage for the ability to
+                # precompute features on shorter chunks,
+                # without memory blow-ups.
+                if torch.cuda.is_available():
+                    logging.info("GPU detected, do the CUDA extraction.")
```
csukuangfj, line 228: I think this line can be moved to where `extractor` is created, i.e, to line 143.

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech.py
```diff
@@ -86,6 +80,19 @@ def get_parser():
...
         "-- use on-the-fly feature extraction in the training script instead.",
     )
+    parser.add_argument(
+        "--num-workers",
+        type=int,
+        default=4,
+        help="Number of dataloading workers used for reading the audio.",
+    )
+    parser.add_argument(
+        "--batch-duration",
```
csukuangfj, line 90: Also, there is another parameter:
https://github.com/lhotse-speech/lhotse/blob/master/lhotse/features/kaldifeat.py#L158
```python3
    chunk_size: Optional[int] = 1000
```

You can use
```
extractor = KaldifeatFbank(
            KaldifeatFbankConfig(device="cuda", chunk_size=<some_value>),
        )
```
to limit the chunk size. I am not sure whether your OOM error is related to these two parameters.

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech.py
danpovey: That is not about GPU memory, it's an error during fork, which is likely due to a limit of virtual memory.  This is probably about the number of dataloader processes (if that's an option), and about the overall size of the manifest.

pzelasko: Yeah it's your CPU memory that's blown up. You can try to decrease num workers (there is an option), but I think Dan is right about the manifest size. There is a way to make it load the manifest lazily, but I suggest you start with splitting the cut manifest into N splits and read->process->save each of them sequentially, and then recombine (you can even do that with bash commands gunzip, split, cat, gzip). I will add some code to support CUDA extraction with streaming manifest reads and writes later.

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech.py
pzelasko: No, `kaldifeat` solves the problem of feature computation, but data augmentation is still an issue. This may actually be another reason why @wgb14 observed OOM (I remember running into this before).

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech.py
pzelasko: 1000 is a reasonable default, you shouldn't get OOM with that value.

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech.py
pzelasko: ... actually, if you're using speed perturbation or other augmentations, ditching them might solve your issues. You can still use MUSAN, SpecAugment, etc. in Dataset later.

danpovey:
There are also commands in lhotse to split and recombine data.  I forget
the names/invocation but Piotr answered my question on this in either this
repo or lhotse's repo.

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech.py
danpovey: [posting on web as email unreliable] There are also commands in lhotse to split and recombine data.  I forget the names/invocation but Piotr answered my question on this in either this repo or lhotse's repo.

csukuangfj:
>  I forget the names/invocation but Piotr answered my question on this in either this
repo or lhotse's repo.

It is in https://github.com/lhotse-speech/lhotse/issues/452#issuecomment-962402670

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech.py
csukuangfj: It is in https://github.com/lhotse-speech/lhotse/issues/452#issuecomment-962402670

csukuangfj:
Looks like there is little progress in the GigaSpeech recipe.

I just made a PR https://github.com/wgb14/icefall/pull/1
to compute the features by splitting the manifests before extraction and combining them afterward.

It seems to resolve the OOM issue. The expected time to extract the features of the XL subset is about 2 days, I think.
(Note: After speed perturbing, the XL subset contains 30k hours of data. The GPU is idle most of the time, so I think computation is not the bottleneck).

csukuangfj:
The screenshot below compares the speed of feature extraction between CUDA and CPU on 1 of the 1000 pieces of the XL subset.

- CUDA:  3 minutes 13 seconds = 193 seconds
- CPU: 7 minutes 19 seconds = 439 seconds
- 439 / 193 = 2.2746

aa21

csukuangfj:
The following screenshot shows the memory assumption of extracting features of the XL subset on CUDA with `--num-workers=20 --batch-duration=600`. 

I believe there will be no OOM anymore. Otherwise, we have to use a larger splits number, e.g., 2000 instead of 1000.
(Note: A split value of 100 still causes OOM.)

---


Screen Shot 2021-11-28 at 15 33 45

csukuangfj:
Also, note that we don't need to limit the number of decoding threads  used by`ffmpeg`.

```diff
diff --git a/lhotse/audio.py b/lhotse/audio.py
index 2190dc9..ca623cf 100644
--- a/lhotse/audio.py
+++ b/lhotse/audio.py
@@ -1437,7 +1437,8 @@ def read_opus_ffmpeg(
     :return: a tuple of audio samples and the sampling rate.
     """
     # Construct the ffmpeg command depending on the arguments passed.
-    cmd = f"ffmpeg -threads 1"
+    #  cmd = f"ffmpeg -threads 1"
+    cmd = f"ffmpeg"
     sampling_rate = 48000
     # Note: we have to add offset and duration options (-ss and -t) BEFORE specifying the input
     #       (-i), otherwise ffmpeg will decode everything and trim afterwards...
@@ -1452,7 +1453,8 @@ def read_opus_ffmpeg(
         cmd += f" -ar {force_opus_sampling_rate}"
         sampling_rate = force_opus_sampling_rate
     # Read audio samples directly as float32.
-    cmd += " -f f32le -threads 1 pipe:1"
+    #  cmd += " -f f32le -threads 1 pipe:1"
+    cmd += " -f f32le pipe:1"
     # Actual audio reading.
     proc = run(cmd, shell=True, stdout=PIPE, stderr=PIPE)
     raw_audio = proc.stdout
```

@pzelasko  Shall we revert https://github.com/lhotse-speech/lhotse/pull/481/files

danpovey:
Great!
Regarding limiting ffmpeg threads, can we see whether it makes a difference to speed before reverting that?
Sometimes when you run something in multiple processes, having it use multiple  threads can be slower, depending on the mechanism it uses.

csukuangfj:
> Regarding limiting ffmpeg threads, can we see whether it makes a difference to speed before reverting that?

Will compare the speed with/without multiple decoding threads for ffmpeg.

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech_splits.py
```diff
@@ -0,0 +1,146 @@
...
+    num_digits = len(str(num_splits))
+
+    device = torch.device("cpu")
+    if torch.cuda.is_available():
+        device = torch.device("cuda", 0)
+    extractor = KaldifeatFbank(KaldifeatFbankConfig(device=device))
+    logging.info(f"device: {device}")
+
+    for i in range(num_splits):
+        idx = f"{i + 1}".zfill(num_digits)
```
wgb14, line 83: Got an error 
```
AssertionError: No such path: data/fbank/XL_split_1000/cuts_XL_raw.0001.jsonl.gz
```
So I modified here to `idx = f"{i + 1}"`

wgb14: So in my experiment it's `data/fbank/XL_split_1000/cuts_XL_raw.1.jsonl.gz`, but from @csukuangfj posted it's `data/fbank/XL_split_1000/cuts_XL_raw.0001.jsonl.gz`.
How does this inconsistency arise?

csukuangfj: > How did this inconsistency arise?

Please use the latest master of lhotse.

wgb14:
This time I didn't get OOM, but got a CUDA OOM error while processing 137/1000:
```
RuntimeError: CUDA out of memory. Tried to allocate 8.07 GiB (GPU 0; 22.41 GiB total capacity; 12.49 GiB already allocated; 7.77 GiB free; 14.00 GiB reserved in total by PyTorch)
```
I believe that 22GB of GPU memory is already higher than the average level of commonly used GPUs, so I'll reduce the value of some params. Which one is preferred, `--num-workers=20` or `--batch-duration=600`?

csukuangfj:
> This time I didn't get OOM, but got a CUDA OOM error while processing 137/1000:

Please install the latest kaldifeat. There was a bug in it. It was not using `chunk_size` in computing features, so it caused CUDA OOM for long utterances.

I just fixed it in https://github.com/csukuangfj/kaldifeat/pull/22

---

BTW, `./prepare.sh --stage 6` will continue the extraction from where it stopped.

csukuangfj:
> I believe that 22GB of GPU memory is already higher than the average level of commonly used GPUs

Because utterances in GigaSpeech are several hours long and`chunkwise` extraction was disabled before by mistake.

After that fix, it should not use that much GPU memory anymore.

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech.py
```diff
@@ -0,0 +1,289 @@
+#!/usr/bin/env python3
```
csukuangfj, line 1: This file is no longer needed and can be removed, I think.

pzelasko:
> > Regarding limiting ffmpeg threads, can we see whether it makes a difference to speed before reverting that?
> 
> Will compare the speed with/without multiple decoding threads for ffmpeg.

Even if you notice no speed difference, I want to avoid spawning num_cpu number of threads for every ffmpeg process, I think it might have been the reason why some of my large training jobs were complaining that I have exhausted system resources even though the memory was fine. These jobs were spawning a lot of ffmpeg subprocesses and I suspected it might be related to that.

wgb14:
image
image
Did you ever get this error?
```
ValueError: Requested more audio (42213.59s) than available (42213.5886875s)
```
I feel that this shouldn't be raised.

pzelasko:
Can you checkout this Lhotse PR? It allows you to set the tolerance threshold for mismatches like these. 

https://github.com/lhotse-speech/lhotse/pull/491

csukuangfj:
I would recommend splitting the manifests into 2000 pieces as it also causes OOM for some pieces if splitting into 1000 pieces.

Also, we can open several terminals on one machine or several machines and do
```
python3 ./local/compute_fbank_gigaspeech_splits.py --num-splits 2000 --start 0 --stop 100 --num-wokers 5
python3 ./local/compute_fbank_gigaspeech_splits.py --num-splits 2000 --start 100 --stop 200 --num-wokers 5
CUDA_VISIBLE_DEVICES=1 python3 ./local/compute_fbank_gigaspeech_splits.py --num-splits 2000 --start 200 --stop 300 --num-wokers 5
```
which can reduce the extraction time.

See https://github.com/wgb14/icefall/pull/2

csukuangfj:
> image image Did you ever get this error?
> 
> ```
> ValueError: Requested more audio (42213.59s) than available (42213.5886875s)
> ```
> 
> I feel that this shouldn't be raised.

The following change works for me with that PR:
```diff
diff --git a/egs/gigaspeech/ASR/local/compute_fbank_gigaspeech_splits.py b/egs/gigaspeech/ASR/local/compute_fbank_gigaspeech_splits.py
index 429168e..f06d1ea 100755
--- a/egs/gigaspeech/ASR/local/compute_fbank_gigaspeech_splits.py
+++ b/egs/gigaspeech/ASR/local/compute_fbank_gigaspeech_splits.py
@@ -28,6 +28,8 @@ from lhotse import (
     KaldifeatFbank,
     KaldifeatFbankConfig,
 )
+from lhotse.audio import set_audio_duration_mismatch_tolerance
+set_audio_duration_mismatch_tolerance(0.01) # seconds

 # Torch's multithreaded behavior needs to be disabled or
 # it wastes a lot of CPU and slow things down.
```

wgb14:
Yes, this also works for me. But It also muted my logger, so I commented out logging in `set_audio_duration_mismatch_tolerance`

wgb14:
By the way, as I checked GPU memory allocation during extraction, I can see about 22GB is allocated. Is this normal behavior? Do we release GPU after extracting each split?

csukuangfj:
> But It also muted my logger, so I commented out logging in set_audio_duration_mismatch_tolerance

You have to put it after setting up the logger.

csukuangfj:
>  Do we release GPU after extracting each split?

No. The allocated GPU memory is cached by PyTorch, I think.
I think you can limit the GPU memory by changing the chunk size while creating the feature extactor.

csukuangfj:
> By the way, as I checked GPU memory allocation during extraction, I can see about 22GB is allocated. Is this normal behavior? Do we release GPU after extracting each split?

I just had a CUDA OOM error:
Screen Shot 2021-12-01 at 07 17 29

I think we may need to empty the cached memory allocated by PyTorch after processing each split.

danpovey:
Since this is a PyTorch allocation error, surely it would free any cached memory if that would be helpful?
(Obviously if we still hold on to a reference to that memory somehow, it's a different matter).

csukuangfj:
From the error message:

> RuntimeError: CUDA out of memory. Tried to allocate 12.51 GiB (GPU 0; 31.75 GiB total capacity; 11.11 GiB already allocated; 10.74 GiB free; 19.64 GiB reserved in total by PyTorch) If reserved memory is >> allocated memory try setting max_split_size_mb to avoid fragmentation.  See documentation for Memory Management and PYTORCH_CUDA_ALLOC_CONF

I think the OOM is caused by fragmentation.  I will try to empty the cache after processing each split and see whether it will cause CUDA OOM again.

wgb14:
> > By the way, as I checked GPU memory allocation during extraction, I can see about 22GB is allocated. Is this normal behavior? Do we release GPU after extracting each split?
> 
> I just had a CUDA OOM error: Screen Shot 2021-12-01 at 07 17 29
> 
> I think we may need to empty the cached memory allocated by PyTorch after processing each split.

I got the same error, and reducing the `chunk_size` doesn't help.

csukuangfj:
Could you try the  following change:

```diff
@@ -123,6 +125,8 @@ def compute_fbank_gigaspeech_splits(args):
             batch_duration=args.batch_duration,
             storage_type=ChunkedLilcomHdf5Writer,
         )
+        if device.type == 'cuda':
+            torch.cuda.empty_cache()
```

The memory occupation reported by nvidia-smi before and after `torch.cuda.empty_cache()` are given below:

Screen Shot 2021-12-01 at 10 47 28 AM


Screen Shot 2021-12-01 at 10 47 47 AM

---

If it still does not help, I would suggest empty the cache after processing every batch.

wgb14:
> Could you try the following change:
> 
> ```diff
> @@ -123,6 +125,8 @@ def compute_fbank_gigaspeech_splits(args):
>              batch_duration=args.batch_duration,
>              storage_type=ChunkedLilcomHdf5Writer,
>          )
> +        if device.type == 'cuda':
> +            torch.cuda.empty_cache()
> ```
> 
> The memory occupation reported by nvidia-smi before and after `torch.cuda.empty_cache()` are given below:
> 
> Screen Shot 2021-12-01 at 10 47 28 AM
> 
> Screen Shot 2021-12-01 at 10 47 47 AM
> 
> If it still does not help, I would suggest empty the cache after processing every batch.

I tried, but got CUDA OOM at the same split where it stopped last time.
I'll try `empty_cache()` after each batch in `compute_and_store_features_batch`.

wgb14:
I tried this:
```
diff --git a/lhotse/cut.py b/lhotse/cut.py
index 9583eab..29d39cd 100644
--- a/lhotse/cut.py
+++ b/lhotse/cut.py
@@ -3838,6 +3838,8 @@ class CutSet(Serializable, Sequence[Cut]):
                     features = extractor.extract_batch(
                         waves, sampling_rate=cuts[0].sampling_rate
                     )
+                    if extractor.device.type == 'cuda':
+                        torch.cuda.empty_cache()

                 for cut, feat_mtx in zip(cuts, features):
                     if isinstance(cut, PaddingCut):

```
but got CUDA OOM in another batch in the same split.

csukuangfj:
Some information about the split that causes CUDA OOM for me is:
a21a


You can see that utterances in the split are 10 to 20 hours long.

I am trying to identify the utterance that is causing CUDA OOM.

danpovey:
How about working on the code that does GPU-based feature extraction, to
add some kind of wrapper that will split and then separately compute and
recombine very long recordings?

csukuangfj:
> How about working on the code that does GPU-based feature extraction, to
add some kind of wrapper that will split and then separately compute and
recombine very long recordings?

Yes, that is how we are doing currently. Please see
https://github.com/csukuangfj/kaldifeat/blob/d2652a2c493678f918eef69bbdb00ff9776b8a6c/kaldifeat/python/kaldifeat/offline_feature.py#L114-L127
```python3
            assert chunk_size > 0
            num_chunks = x.size(0) // chunk_size
            end = 0
            features = []
            for i in range(num_chunks):
                start = i * chunk_size
                end = start + chunk_size
                this_chunk = self.computer.compute_features(
                    x[start:end], vtln_warp
                )
                features.append(this_chunk)
            if end < x.size(0):
                last_chunk = self.computer.compute_features(x[end:], vtln_warp)
                features.append(last_chunk)
            features = torch.cat(features, dim=0)
```

The default chunk size used in lhotse is 1000.
See https://github.com/lhotse-speech/lhotse/blob/master/lhotse/features/kaldifeat.py#L158
```python3
    # This is an extra setting compared to kaldifeat FbankOptions:
    # by default, we'll ask kaldifeat to compute the feats in chunks
    # to avoid excessive memory usage.
    chunk_size: Optional[int] = 1000
```

danpovey:
I think it might make more sense to transfer to CPU after computing each chunk, and concatenate on CPU.
And also transfer only the chunks to CUDA.  Could use a much larger chunk size in that case, perhaps.

csukuangfj:
Thanks, will change it.

csukuangfj:
> I think it might make more sense to transfer to CPU after computing each chunk, and concatenate on CPU.

See https://github.com/lhotse-speech/lhotse/pull/498
and
https://github.com/csukuangfj/kaldifeat/pull/24

I think it should fix the CUDA OOM errors with kaldifeat v1.12.

csukuangfj:
Also, we don't need to empty the cache after each split. I just verified with the output of `nvidia-smi`, the GPU RAM consumption is rather stable, about 2 GB when the chunk size is 20 minutes.

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech_splits.py
```diff
@@ -100,6 +101,8 @@ def compute_fbank_gigaspeech_splits(args):
     extractor = KaldifeatFbank(KaldifeatFbankConfig(device=device))
     logging.info(f"device: {device}")
 
+    set_audio_duration_mismatch_tolerance(0.01)  # 10ms tolerance
```
csukuangfj, line 104: You may need to set it to 0.1

wgb14: Did you see any error if set 0.01? I just finished extraction by setting 0.01

csukuangfj:
I just realized that we have to disable the caching mechanism in lhotse when reading audio.

The default size of cached items is 512. For corpus like GigaSpeech where utterances are several hours long, it's going
to take lots of RAM and causes OOM easily. Also, as each audio file is read only once from disk and there is no point caching it, I think.

See https://github.com/lhotse-speech/lhotse/blob/master/lhotse/audio.py#L1064-L1065
```python3
@dynamic_lru_cache
def read_audio(
```

https://github.com/lhotse-speech/lhotse/blob/master/lhotse/caching.py#L37
```python3
    To disable/enable caching globally in Lhotse, call::
        >>> from lhotse import set_caching_enabled
        >>> set_caching_enabled(True)   # enable
        >>> set_caching_enabled(False)  # disable
    Currently it hard-codes the cache size at 512 items
    (regardless of the array size).
```

pzelasko:
Good point! Never noticed it because I was using on the fly features.

wgb14:
Finally, I finished feature extraction, and now I'm about to start the model training.
Any suggestions on parameters? Or I'll begin with scripts in `egs/librispeech/ASR/conformer_ctc/`, and utilize params in https://github.com/k2-fsa/icefall/blob/95af0397336ac840a5bfed1ae8de79dbddcdad71/egs/librispeech/ASR/RESULTS.md?
By the way, do we support training on multi-GPU multi-node now?

danpovey:
Cool!
I would start with the librispeech setup for now, although the optimal setup would probably have more layers and/or have larger d_model.
I believe Fangjun has experimented with multi-GPU training, although I'm not sure if we currently have support for it.

csukuangfj:
As for multi-node multi-GPU training, there is a PR about it. Please see https://github.com/k2-fsa/icefall/pull/63

The entrypoint is the file `egs/librispeech/ASR/conformer_ctc/run-multi-node-multi-gpu.sh` in that PR.

I would recommend you to use PyTorch >= 1.9.0 so that you can use a different number of GPUs on each node.

Please leave a message if you encounter any issues.

wgb14:
I got this warning if set `--small-dev True`. I think 28 is actually the number of recordings.
image

Also, if `--world-size` >1, `setup_logger()` would not work, and I can only see warnings.

csukuangfj:
> Also, if --world-size >1, setup_logger() would not work, and I can only see warnings. Did you ever see this bug

It depends on your PyTorch version. Please see https://github.com/k2-fsa/icefall/issues/35

One workaround is to do the following inside the `main()` function (or do it globally)
```
logging.info = logging.warning
```

csukuangfj:
> I got this warning if set --small-dev True. I think 28 is actually the number of recordings.

It is caused by
https://github.com/k2-fsa/icefall/blob/bea78f609445a407db2377304818da550268d79c/egs/gigaspeech/ASR/conformer_ctc/asr_datamodule.py#L375

1000 is too large for this `CutSet` and leads to the above warning.

wgb14:
> ```
> logging.info = logging.warning
> ```

Thanks, I can see the log from console now.


> > I got this warning if set --small-dev True. I think 28 is actually the number of recordings.
> 
> It is caused by
> 
> https://github.com/k2-fsa/icefall/blob/bea78f609445a407db2377304818da550268d79c/egs/gigaspeech/ASR/conformer_ctc/asr_datamodule.py#L375
> 
> 1000 is too large for this `CutSet` and leads to the above warning.

I copied this part from @pzelasko's recipe in snowfall . Here I thought 1000 refers to the numbers of utterances, since there are about 6000 lines of utterances in `cuts_DEV`, and we want to speed up validation. But now it seems that the option `first=1000` tends to return 1000 recordings.

I got CUDA OOM error even if specified `--max-duration` to 100, should I continue reducing this param? Or do we have a table about the relation between GPU memory and `max-duration` value?

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech_dev_test.py
```diff
@@ -0,0 +1,90 @@
...
+    for partition in subsets:
+        cuts_path = in_out_dir / f"cuts_{partition}.jsonl.gz"
+        if cuts_path.is_file():
+            logging.info(f"{cuts_path} exists - skipping")
+            continue
+
+        raw_cuts_path = in_out_dir / f"cuts_{partition}_raw.jsonl.gz"
+
+        logging.info(f"Loading {raw_cuts_path}")
+        cut_set = CutSet.from_file(raw_cuts_path)
```
pzelasko, line 64: @wgb14 you need to run `.trim_to_supervisions(keep_overlapping=False)` on this cut set in order to get data that is usable for validation (otherwise the cuts represent whole recordings). I see you have it for train cuts, but not for dev and test.

wgb14: Good catch, fixed

csukuangfj: Do you still have OOM after this fix?

csukuangfj:
> I got CUDA OOM error even if specified --max-duration to 100, should I continue reducing this param? Or do we have a table about the relation between GPU memory and max-duration value?

If are there are very very long utterances, e.g., several hours, in the CutSet, it is very likely that it will always cause OOM even if you use `--max-duration 1`.

https://github.com/k2-fsa/icefall/pull/120#discussion_r771685179 would help to fix the OOM issue due to long utterances.

---

To use a larger `--max-duration` in the training, I would recommend using https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/local/display_manifest_statistics.py
to get an idea of the duration distribution and use it to remove very long utterances from the CutSet, see
https://github.com/k2-fsa/icefall/blob/1d44da845b264b3a24fadcc6af577055d399220d/egs/librispeech/ASR/transducer/train.py#L616-L630

**Caution:** The threshold `20` is for LibriSpeech. You may need to change it for GigaSpeech. And you have to use `trim_to_supervisions()`  before using `display_manifest_statistics.py`.

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech_dev_test.py
wgb14: Yes, and I'll check with your statistic tool first.
image
Does this mean an utterance of 22s caused the CUDA OOM?

csukuangfj: > Does this mean an utterance of 22s caused the CUDA OOM?

I don't think so. You need to check the total duration, after padding, in the batch that is causing CUDA OOM.

wgb14: GigaSpeech has removed all the utterances longer than 20s, the longest utterance in CutSet would be about 22.22s after speed perturbation.

csukuangfj: So does it still cause OOM after removing long utterances and using a smaller `--max-duration`?

wgb14: Yes, although I have set `--max-duration` to 80. By the way, does `--num-workers` affect GPU memory usage?

egs/gigaspeech/ASR/conformer_ctc/asr_datamodule.py
```diff
@@ -0,0 +1,382 @@
...
+            default=True,
+            help="When enabled, each batch will have the "
+            "field: batch['supervisions']['cut'] with the cuts that "
+            "were used to construct it.",
+        )
+
+        group.add_argument(
+            "--num-workers",
+            type=int,
+            default=20,
```
csukuangfj, line 145: Oh, this value is too large, I think, it may cause CPU OOM.
We are using `2` in LibriSpeech. It was 4, which caused CPU OOM when there are limited CPU RAM.

egs/gigaspeech/ASR/conformer_ctc/asr_datamodule.py
```diff
@@ -0,0 +1,382 @@
...
+            "--max-duration",
+            type=int,
+            default=200.0,
+            help="Maximum pooled recordings duration (seconds) in a "
+            "single batch. You can reduce it if it causes CUDA OOM.",
+        )
+        group.add_argument(
+            "--bucketing-sampler",
+            type=str2bool,
+            default=False,
```
csukuangfj, line 85: Also, there is less chance to get CUDA OOM when enabling `bucketing sampler`.

danpovey: I wonder if it might help to print the exception and stack-trace from the out-of-memory batch, so we know what the total GPU memory was and how much was used.  E.g. that would make it clear if some other process was using GPU memory.

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech_splits.py
csukuangfj: Yes, I remember that I have to reset it to 0.1. But it is fine if 0.01 works for you.

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech_dev_test.py
csukuangfj: > By the way, does --num-workers affect GPU memory usage?

No, I don't think so. DataLoader uses only CPU memory.

egs/gigaspeech/ASR/conformer_ctc/asr_datamodule.py
wgb14: fixed

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech_dev_test.py
pzelasko: > > Does this mean an utterance of 22s caused the CUDA OOM?
> 
> I don't think so. You need to check the total duration, after padding, in the batch that is causing CUDA OOM.

It is possible that 1) your batch size is still too large 2) it can be larger if you drop longer utterances (eg filter at 18s or less) 3) you are not using bucketing and simply found a pessimistic batch with a lot of padding (in which case you have to use bucketing or filter long cuts otherwise your training will eventually fail)

egs/gigaspeech/ASR/conformer_ctc/asr_datamodule.py
pzelasko: I think he’s using lazy manifests in which case the RAM usage is greatly reduced with more workers. But it won’t play super well with bucketing (I am planning a “on the fly” bucketing mechanism in the future)

egs/gigaspeech/ASR/conformer_ctc/asr_datamodule.py
wgb14: I agree `bucketing sampler` would help. But  from https://github.com/k2-fsa/snowfall/blob/911198817edc7b306265f32447ef8a7dc5cfa8f2/egs/gigaspeech/asr/simple_v1/asr_datamodule.py#L179-L192, our lazily-loaded CutSet does not work efficiently with `bucketing sampler`. @pzelasko please confirm if it is still the case.

tz301:
> > I got CUDA OOM error even if specified --max-duration to 100, should I continue reducing this param? Or do we have a table about the relation between GPU memory and max-duration value?
> 
> If there are very very long utterances, e.g., several hours, in the CutSet, it is very likely that it will always cause OOM even if you use `--max-duration 1`.
> 
> [#120 (comment)](https://github.com/k2-fsa/icefall/pull/120#discussion_r771685179) would help to fix the OOM issue due to long utterances.
> 
> To use a larger `--max-duration` in the training, I would recommend using https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/local/display_manifest_statistics.py to get an idea of the duration distribution and use it to remove very long utterances from the CutSet, see
> 
> https://github.com/k2-fsa/icefall/blob/1d44da845b264b3a24fadcc6af577055d399220d/egs/librispeech/ASR/transducer/train.py#L616-L630
> 
> **Caution:** The threshold `20` is for LibriSpeech. You may need to change it for GigaSpeech. And you have to use `trim_to_supervisions()` before using `display_manifest_statistics.py`.

I also meet similiar problem using my large data (larger than 10000h). 
For feature extraction, after seen gigaspeech recipe, it's ok now to use lazy cuts. Also I have many subsets, with which max is ~1500h. I create cuts and features seperately, and merge to one cuts.jsonl.gz file finally.

But for training, I can only make it work by set --max-duration=50 (80 will have OOM). It's extremely slow, will need two weeks for one epoch using 4 GPU.
I see the duration distribution, 60% are less than 3s (min is 0.5s) and 2% are large than 15s (max is 20s). Maybe valid duration per batch is so small due to useless padding.

I try something:
(1)  Use BuckerSampler and set --max-duration=200, but it cost too much cpu memory. Each parallel will read all cuts into memory. Only --world-size=1 can be supported on my 256G memory machine.
(2) Reorder cuts in advance and save to file. I split cuts into buckets (1~2s, 2~3s), shuffing inside buckets and recombined orderly. So samples in one batch will not padding too much. Not sure if this will cost converge problem or wer degradation, still run now. Using SingleCutSampler and set --max-duration=200 & --world-size=4, speed is 60h for one epoch.

Hope there will be lazy BuckerSampler or better solution. @pzelasko

danpovey:
> I got CUDA OOM error even if specified --max-duration to 100, should I
> continue reducing this param? Or do we have a table about the relation
> between GPU memory and max-duration value?
>
> If there are very very long utterances, e.g., several hours, in the
> CutSet, it is very likely that it will always cause OOM even if you use --max-duration
> 1.
>
> #120 (comment)
> <https://github.com/k2-fsa/icefall/pull/120#discussion_r771685179> would
> help to fix the OOM issue due to long utterances.
>
> To use a larger --max-duration in the training, I would recommend using
> https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/local/display_manifest_statistics.py
> to get an idea of the duration distribution and use it to remove very long
> utterances from the CutSet, see
>
>
> https://github.com/k2-fsa/icefall/blob/1d44da845b264b3a24fadcc6af577055d399220d/egs/librispeech/ASR/transducer/train.py#L616-L630
>
> *Caution:* The threshold 20 is for LibriSpeech. You may need to change it
> for GigaSpeech. And you have to use trim_to_supervisions() before using
> display_manifest_statistics.py.
>
> I also meet similiar problem using my large data (larger than 10000h).
> For feature extraction, after seen gigaspeech recipe, it's ok now to use
> lazy cuts. Also I have many subsets, with which max is ~1500h. I create
> cuts and features seperately, and merge to one cuts.jsonl.gz file finally.
>
> But for training, I can only make it work by set --max-duration=50 (80
> will have OOM). It's extremely slow, will need two weeks for one epoch
> using 4 GPU.
>
It might be worth printing out some information on the failing batch to
determine the utterance size and num-utterances that is failing,
e.g. to see whether it's about very mismatched sizes, or one very long
utterance, or many short ones.
We're trying to get away from the transformer-decoder, which is a little
expensive in terms of memory.

pzelasko:
> I try something: (1) Use BuckerSampler and set --max-duration=200, but it cost too much cpu memory. Each parallel will read all cuts into memory. Only --world-size=1 can be supported on my 256G memory machine. (2) Reorder cuts in advance and save to file. I split cuts into buckets ([1s, 2s], [2s, 3s], ...), shuffing inside buckets and recombined orderly. So samples in one batch will not padding too much. Not sure if this will cost converge problem or wer degradation, still run now. Using SingleCutSampler and set --max-duration=200 & --world-size=4, speed is 60h for one epoch.
> 
> Hope there will be lazy BuckerSampler or better solution. @pzelasko

Check out the `DynamicBucketingSampler` here (added in https://github.com/lhotse-speech/lhotse/pull/517): https://github.com/lhotse-speech/lhotse/blob/master/lhotse/dataset/sampling/dynamic_bucketing.py

I am starting to look into your other issue with growing RAM to see if I can replicate it... efficient large-scale training is something we definitely want to support.

egs/gigaspeech/ASR/conformer_ctc/asr_datamodule.py
pingfengluo: @wgb14 how many CPU RAM it takes to run GigaSpeech recipe with your asr_datamodule.py setting? 

https://github.com/k2-fsa/icefall/pull/120/files/bea78f609445a407db2377304818da550268d79c#diff-13aabaed6dbf8573df7a9b01e6c22c716b4cff37948c98055b915a010404bd63

pzelasko: FYI @wgb14 @pingfengluo you can now improve the RAM usage (and I think by a lot) by doing the following:

1) open manifests lazily (i.e., `CutSet.from_jsonl_lazy(path)` or `lhotse.load_manifest_lazy(path)`, both are equivalent)

2) replace `BucketingSampler` with `DynamicBucketingSampler` (you might need to remove/change some arguments, but they are mostly the same)

pzelasko:
FYI for people tracking the RAM issues here: I added the "lazy" bucketing and described how to switch to it here: https://github.com/k2-fsa/icefall/pull/120#discussion_r776461928

If you're using precomputed features stored in HDF5 files, you might still notice growing CPU RAM usage. Unfortunately I can't find a way to disable HDF5 caches (and some people tell me it might be a memory leak in HDF5 itself). I created a different storage format that should be ready to use; if somebody wants to help me in testing its stability, there's an easy way to do so:

1) check out Lhotse at branch from PR: https://github.com/lhotse-speech/lhotse/pull/522

2) copy your existing cut manifest using the command: 

```bash
$ lhotse copy-feats -t lilcom_chunky <path-to-cuts.jsonl.gz> <path-to-copy.jsonl.gz> <new-feat-dir>
```

Example:

```bash
$ mkdir data/fbank_cpy
$ lhotse copy-feats -t lilcom_chunky data/fbank/cuts_train-other-500.jsonl.gz data/fbank_cpy/cuts_train-other-500.jsonl.gz data/fbank_cpy/train-other-500
```

(please notice the "l" letter in extension .jsonl.gz; for the output cut set it MUST be in `.jsonl` or `.jsonl.gz` format)

3) change paths in the `asr_datamodule.py` script to the new cutset and re-run the training

egs/gigaspeech/ASR/conformer_ctc/asr_datamodule.py
wgb14: @pingfengluo updated my `asr_datamodule.py` setting https://github.com/k2-fsa/icefall/pull/120/commits/6e5b189fc5509c76bc2d989710ee85e78ad6603b
I didn't get CPU OOM with DynamicBucketingSampler and lazy-loading training set.

pingfengluo: @pzelasko with DynamicBucketingSampler,  there is still OOM on WenetSpeech dataset  training on a 256G RAM machine, my asr_datamodule.py is https://github.com/pingfengluo/icefall/blob/wenetspeech/egs/wenetspeech/ASR/tdnn_lstm_ctc/asr_datamodule.py .
should i checkout this PR https://github.com/k2-fsa/icefall/pull/120#discussion_r776461928 ?

pzelasko: @pingfengluo can you use on the fly feature extraction and see if the issue persists?

pingfengluo: > @pingfengluo can you use on the fly feature extraction and see if the issue persists?

i have tried on-the-fly feature extraction, but there still was OOM.

pzelasko: You probably haven’t opened the CutSet lazily (`CutSet.from_jsonl_lazy(path)`)

pingfengluo: > You probably haven’t opened the CutSet lazily (`CutSet.from_jsonl_lazy(path)`)

I make sure  than using  `CutSet.from_jsonl_lazy(path)`  ,  asr_datamodule.py and log as  https://github.com/k2-fsa/icefall/pull/167/files#diff-8951e342449f8076c154f4382e8c56579448dd8bc41e021a1ddf9794e115a5c8
image

danpovey: Does memory usage seem to increase gradually?
I wonder whether it dies while processing an exceptionally long utterance?
Don't lose hope!  I'm sure we are getting closer to a solution.

danpovey: ... and please clarify every time you mention OOM whether it is CUDA or CPU, because the comments on this PR mention both types.

pzelasko: Is there any specific function call that makes it go OOM?

As a test, can you comment out the code for forward and backward pass, just to let it sample batches from the data pipeline, and see if you’re still getting OOM? It will help determine if it’s data pipeline related or not. I didn’t see anything obviously wrong in the data module.

pingfengluo: > Does memory usage seem to increase gradually? 
Yes. The CPU RAM  increased gradually until using up 256GB.

commands on WenetSpeech recipe are 
```
cd icefall/egs/wenetspeech/ASR && ./transducer_stateless/train.py \
  --bucketing-sampler True \
  --world-size 8 \
  --lang-dir data/lang_char \
  --num-epochs 40 \
  --start-epoch 0 \
  --exp-dir transducer_stateless/exp_char \
  --max-duration 160 \
  --lr-factor 3
```
code to test  is in this PR https://github.com/k2-fsa/icefall/pull/167

> I wonder whether it dies while processing an exceptionally long utterance? 
I will try to remove long utterance.

> I'm sure we are getting closer to a solution.
Great!

pingfengluo: > As a test, can you comment out the code for forward and backward pass, just to let it sample batches from the data pipeline, and see if you’re still getting OOM? 

Ok,  i will test it. thanks.

wgb14:
Also update results here:
The best WER, as of 2022-04-06, for the gigaspeech is below
(using HLG decoding + n-gram LM rescoring + attention decoder rescoring):

|     |  Dev  | Test  |
|-----|-------|-------|
| WER | 11.93 | 11.86 |

Scale values used in n-gram LM rescoring and attention rescoring for the best WERs are:
| ngram_lm_scale | attention_scale |
|----------------|-----------------|
|      0.3       |        1.5      |

Ready for review now.

egs/gigaspeech/ASR/conformer_ctc/asr_datamodule.py
```diff
@@ -0,0 +1,382 @@
...
+            sampler=sampler,
+            num_workers=self.args.num_workers,
+        )
+        return test_dl
+
+    @lru_cache()
+    def train_cuts(self) -> CutSet:
+        logging.info(f"About to get train_{self.args.subset} cuts")
+        path = self.args.manifest_dir / f"cuts_{self.args.subset}.jsonl.gz"
+        if self.args.subset in ["L", "XL"] and self.args.lazy_load:
```
pzelasko, line 364: Since you're using DynamicBucketingSampler anyway, it makes sense to always open manifests lazily regardless of the chosen split. I'd remove `self.args.lazy_load` option and hard-code it.

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech_splits.py
```diff
@@ -0,0 +1,170 @@
...
+        cut_set = CutSet.from_file(raw_cuts_path)
+
+        logging.info("Computing features")
+
+        cut_set = cut_set.compute_and_store_features_batch(
+            extractor=extractor,
+            storage_path=f"{output_dir}/feats_XL_{idx}",
+            num_workers=args.num_workers,
+            batch_duration=args.batch_duration,
+            storage_type=ChunkedLilcomHdf5Writer,
```
pzelasko, line 127: I recommend removing this line, the default storage_type is better suited for large datasets.

egs/gigaspeech/ASR/prepare.sh
```diff
@@ -0,0 +1,324 @@
...
+if [ $stage -le 4 ] && [ $stop_stage -ge 4 ]; then
+  log "Stage 4: Compute features for DEV and TEST subsets of GigaSpeech (may take 2 minutes)"
+  python3 ./local/compute_fbank_gigaspeech_dev_test.py
+fi
+
+if [ $stage -le 5 ] && [ $stop_stage -ge 5 ]; then
+  log "Stage 5: Split XL subset into ${num_splits} pieces (may take 30 minutes)"
+  split_dir=data/fbank/XL_split_${num_splits}
+  if [ ! -f $split_dir/.split_completed ]; then
+    lhotse split $num_splits ./data/fbank/cuts_XL_raw.jsonl.gz $split_dir
```
pzelasko, line 147: There is a new mode (if I remember correctly `lhotse split-lazy`) that may be faster and more memory friendly, where you specify num_items per split instead of num_splits.

pzelasko (commented):
Nice work! I left a couple of comments related to data part.

danpovey:
Cool!
BTW, if you have time, it might be worth trying the RNN-T setup, because it tends to shine when the amount of data is very large.

csukuangfj:
Please use the pruned RNN-T (https://github.com/k2-fsa/icefall/tree/master/egs/librispeech/ASR/pruned_transducer_stateless) if you want to try it.

pzelasko:
Good point -- when I was building the snowfall recipe a while ago, I wrote some code for scoring in the "official" way that you could use https://github.com/k2-fsa/snowfall/blob/911198817edc7b306265f32447ef8a7dc5cfa8f2/egs/gigaspeech/asr/simple_v1/mmi_att_transformer_decode.py#L602-L615

wgb14:
> @wgb14 Guanbo, as a reminder, I'm wondering if you have added text post-processing [here](https://github.com/SpeechColab/GigaSpeech/blob/main/utils/gigaspeech_scoring.py#L11). For spontaneous dataset evaluation, things like this is essential to ensure apple-to-apple comparisons (Other baselines in GigaSpeech paper all applied it to both references & hypotheses). With that added, I guess K2's number should be already as good as Espnet.

Thanks! I did forgot to do text post-processing. Will update results here.

chenguoguo:
And we have a GigaSpeech leaderboard here, I hope we'll soon be able to add the K2 numbers: https://github.com/SpeechColab/GigaSpeech#leaderboard

And thanks a lot for getting the initial numbers Guanbo!

wgb14:
Updated results here: (using HLG decoding + n-gram LM rescoring + attention decoder rescoring):

|     |  Dev  | Test  |
|-----|-------|-------|
| WER | 10.5 |  |

Scale values used in n-gram LM rescoring and attention rescoring for the best WERs are:
| ngram_lm_scale | attention_scale |
|----------------|-----------------|
|      0.3       |        1.5      |

Got a much better number. But it seems that this pair of scales is no longer the best. I'll redo the param scanning and update numbers in this PR.

chenguoguo:
This is great news. Guanbo, once you get all the numbers and clean up the recipe, you can update the leaderboard here: https://github.com/SpeechColab/GigaSpeech#leaderboard

wgb14:
Updating results:
Results using HLG decoding + n-gram LM rescoring + attention decoder rescoring:

|     |  Dev  | Test  |
|-----|-------|-------|
| WER | 10.47 | 10.58 |

Results using HLG decoding + whole lattice rescoring:

|     |  Dev  | Test  |
|-----|-------|-------|
| WER | 10.51 | 10.62 |

egs/gigaspeech/ASR/conformer_ctc/asr_datamodule.py
wgb14: Fixed

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech_splits.py
wgb14: I did some experiments between these 2 types: https://github.com/lhotse-speech/lhotse/pull/527 and https://github.com/lhotse-speech/lhotse/pull/522.
In my experiments, `ChunkedLilcomHdf5Writer` (the HDF5) works well, and memory usage was kept in the range between 150G~200G, with `--world-size 8 --num-workers 1 --max-duration 120`
While I use `LilcomChunkyWriter`, I got some large `jsonl.gz` files, one example:
`LilcomChunkyWriter`:
```
cuts_XL.1324.jsonl.gz 4.03 GB
feats_XL_1324.lca     17.3 GB
```
`ChunkedLilcomHdf5Writer`:
```
cuts_XL.1324.jsonl.gz 16.0 MB
feats_XL_1324.h5      18.2 GB
```
Is this large json file as expected?

pzelasko: I wouldn't say expected, but I guess it makes sense if there is a lot of really long recordings.. nevermind then.

wgb14: Yes, it has a super long array in `"storage_key"`. Anyway I'll use this instead if better suited for large datasets.

egs/gigaspeech/ASR/prepare.sh
wgb14: Great! I've tried `lhotse split-lazy`, and it's really memory efficient. but it generates manifests like `0.jsonl.gz`. Can we generate `cuts_XL_raw.0001.jsonl.gz` just as what `lhotse split` do?

wgb14: I mean adding suffix and starts from 0001

egs/gigaspeech/ASR/local/compute_fbank_gigaspeech_splits.py
wgb14: Fixed

egs/gigaspeech/ASR/prepare.sh
pzelasko: That makes sense, do you think you could submit a PR with that? I'm a bit overwhelmed atm.

dophist:
> Updating results: Results using HLG decoding + n-gram LM rescoring + attention decoder rescoring:
> 
> Dev	Test
> WER	10.47	10.58
> Results using HLG decoding + whole lattice rescoring:
> 
> Dev	Test
> WER	10.51	10.62

This is the best number I've ever seen for GigaSpeech, great.

egs/gigaspeech/ASR/prepare.sh
wgb14: Sure, I opened a PR here: https://github.com/lhotse-speech/lhotse/pull/664

danpovey:
Guanbo, if you want to run an RNN-T recipe, please now use the setup in pruned_transducer_stateless2 from librispeech.  This converges much faster than the old setup.
The only options you might want to change are:
    --lr-epochs (reduce from 6 to some number less than about half the number of epochs you plan to train, e.g. 2 or 3).
You might want to add --use-fp16=True to use half-precision to speed up training...
and you can adjust --max-duration accordingly.  --use-fp16=True allows, on our GPUs, to increase max-duration from 300 to 550.  (However it will converge slower near the start of training if the max-duration is too high, so this is a tradeoff).

csukuangfj:
@wgb14 

Could you also upload the training log `exp/log/log-xxxx-0` in the result directory? 
I want to compare the training time across batches as I find that it is slow to do on-the-fly feature extraction with GigaSpeech during training.

wgb14:
> @wgb14
> 
> Could you also upload the training log `exp/log/log-xxxx-0` in the result directory? I want to compare the training time across batches as I find that it is slow to do on-the-fly feature extraction with GigaSpeech during training.

[log-train-2022-04-01-02-04-11-0.txt](https://github.com/k2-fsa/icefall/files/8476095/log-train-2022-04-01-02-04-11-0.txt)

I used precomputed feature to train. It took about 5 minutes per 500 batch, with `--max-duration 120 --num-workers 1`

wgb14:
> Guanbo, if you want to run an RNN-T recipe, please now use the setup in pruned_transducer_stateless2 from librispeech. This converges much faster than the old setup. The only options you might want to change are: --lr-epochs (reduce from 6 to some number less than about half the number of epochs you plan to train, e.g. 2 or 3). You might want to add --use-fp16=True to use half-precision to speed up training... and you can adjust --max-duration accordingly. --use-fp16=True allows, on our GPUs, to increase max-duration from 300 to 550. (However it will converge slower near the start of training if the max-duration is too high, so this is a tradeoff).

Thanks for so many suggestions! I'll open a PR if I got resources to train a RNN-T model

csukuangfj:
Is it ready for merge?

chenguoguo:
I think we can use the cluster we borrowed from THU/SJTU. I'll make sure I
get some computation resources for you.

Guoguo

wgb14:
> Is it ready for merge?

Ready for merge now

egs/gigaspeech/ASR/prepare.sh
wgb14: fixed

egs/gigaspeech/ASR/conformer_ctc/gigaspeech_scoring.py
```diff
@@ -0,0 +1,115 @@
...
+                for line in fi:
+                    line = line.strip()
+                    if line:
+                        cols = line.split()
+                        text = asr_text_post_processing(" ".join(cols[0:-1]))
+                        uttid_field = cols[-1]
+                        print(f"{text} {uttid_field}", file=fo)
+
+    # GigaSpeech's uttid comforms to swb
+    os.system(f"sclite -r {REF} trn -h {HYP} trn -i swb | tee {RESULT}")
```
csukuangfj, line 115: Is there any documentation about how to install `sclite`?

I think k2 is using https://github.com/pzelasko/kaldialign to compute the edit distance and it also outputs different types of errors.

wgb14: We don't use sclite to do scoring in icefall, only using the funtion `asr_text_post_processing` in `decode.py`

danpovey: It seems the "-i swb" option means the "utterance-id type"... it seems this means:
"the utterance id is made up of a                                                                                                                                                                                   
speaker code, followed by a hyphen or  underscore,                                                                                                                                                                                        
followed by an utterance number.   "
I think this means that if we only care about the final WER, sclite is probably doing the same as kaldialign and may not be necessary.
Installation of sctk is not something super-trivial, so it would be good to get rid of this dependency.

csukuangfj: Thanks.

danpovey:
I'm OK to keep the sclite stuff in, for documentation purposes, as long as it's totally optional and this is made clear to the user.

egs/gigaspeech/ASR/conformer_ctc/gigaspeech_scoring.py
csukuangfj: I find that this code is only executed when this script is invoked from the command line. 
Its core part is `asr_text_post_processing`, which is imported in `decode.py`.

wgb14: Yes, I agree. The main funtion of this file isn't used in our script. This is just an option. I can delete these lines if they introduce any confusion.

csukuangfj:
Thanks! I am merging it.

PR #125, opened 2021-11-18
Associate a cut with token alignment (without repeats)

csukuangfj:
This PR depends on the master branch of lhotse


test/test_ali.py
```diff
@@ -44,180 +44,47 @@
...
-    lexicon = Lexicon(lang_dir)
-    max_token_id = max(lexicon.tokens)
-    num_classes = max_token_id + 1  # +1 for the blank
-    word_table = lexicon.word_table
-
-    HLG = k2.Fsa.from_dict(
-        torch.load(f"{lang_dir}/HLG.pt", map_location=device)
-    )
-
     for batch in dl:
```
csukuangfj, line 82: @pzelasko 
This line causes the following exception.

I think the interface for `K2SpeechRecognitionDataset` also needs to be updated to pass a `pad_value_dict`.

```
  File "test/test_ali.py", line 90, in <module>
    test()
  File "test/test_ali.py", line 82, in test
    for batch in dl:
  File "/ceph-fj/fangjun/py38/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 435, in __next__
    data = self._next_data()
  File "/ceph-fj/fangjun/py38/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 1085, in _next_data
    return self._process_data(data)
  File "/ceph-fj/fangjun/py38/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 1111, in _process_data
    data.reraise()
  File "/ceph-fj/fangjun/py38/lib/python3.8/site-packages/torch/_utils.py", line 428, in reraise
    raise self.exc_type(msg)
AssertionError: Caught AssertionError in DataLoader worker process 0.
Original Traceback (most recent call last):
  File "/ceph-fj/fangjun/py38/lib/python3.8/site-packages/torch/utils/data/_utils/worker.py", line 198, in _worker_loop
    data = fetcher.fetch(index)
  File "/ceph-fj/fangjun/py38/lib/python3.8/site-packages/torch/utils/data/_utils/fetch.py", line 46, in fetch
    data = self.dataset[possibly_batched_index]
  File "/ceph-fj/fangjun/open-source-2/lhotse-ali/lhotse/dataset/speech_recognition.py", line 105, in __getitem__
    inputs, _ = self.input_strategy(cuts)
  File "/ceph-fj/fangjun/open-source-2/lhotse-ali/lhotse/dataset/input_strategies.py", line 110, in __call__
    return collate_features(cuts, executor=_get_executor(self.num_workers))
  File "/ceph-fj/fangjun/open-source-2/lhotse-ali/lhotse/dataset/collation.py", line 127, in collate_features
    cuts = maybe_pad(
  File "/ceph-fj/fangjun/open-source-2/lhotse-ali/lhotse/dataset/collation.py", line 342, in maybe_pad
    return cuts.pad(
  File "/ceph-fj/fangjun/open-source-2/lhotse-ali/lhotse/cut.py", line 3130, in pad
    return CutSet.from_cuts(
  File "/ceph-fj/fangjun/open-source-2/lhotse-ali/lhotse/cut.py", line 2637, in from_cuts
    return CutSet(cuts=index_by_id_and_check(cuts))
  File "/ceph-fj/fangjun/open-source-2/lhotse-ali/lhotse/utils.py", line 481, in index_by_id_and_check
    for m in manifests:
  File "/ceph-fj/fangjun/open-source-2/lhotse-ali/lhotse/cut.py", line 3131, in <genexpr>
    cut.pad(
  File "/ceph-fj/fangjun/open-source-2/lhotse-ali/lhotse/cut.py", line 1096, in pad
    return pad(
  File "/ceph-fj/fangjun/open-source-2/lhotse-ali/lhotse/cut.py", line 4137, in pad
    assert pad_value_dict is not None and all(
AssertionError: Cut being padded has custom TemporalArray attributes: ['token_alignment']. We expected a 'pad_value_dict' argument with padding values for these attributes.
```

pzelasko: Gotcha. Yeah that assertion will be a pain in the butt. I'll change it to a warning and use a default value of 0 as you suggested.

egs/librispeech/ASR/conformer_ctc/ali.py
```diff
@@ -240,71 +308,66 @@ def main():
...
-        save_alignments(
-            alignments=alignments,
-            subsampling_factor=params.subsampling_factor,
-            filename=filename,
-        )
-        logging.info(
-            f"For dataset {name}, its alignments are saved to {filename}"
-        )
+
+    cut_set.to_json(out_manifest_filename)
```
pzelasko, line 364:
```suggestion
    cut_set.to_file(out_manifest_filename)
```

will make it easier if you ever want to switch to `jsonl` which enables lazy reading support

test/test_ali.py
csukuangfj: Thanks. No, it doesn't.

csukuangfj: One thing I don't quite understand about the exception is that the data loader does not read the custom fileds
contained in the CutSet, but why does it throw an exception related to custom fields?

pzelasko: That's because the assertion was inside the `cut.pad` function; the assertion was supposed to be a sanity check to let you know that you forgot to specify the pad value for custom fields (if any are present). I only realized later that in many use-cases you actually won't care about the custom attributes presence so it makes sense to use a default value instead.

pzelasko: (the K2 dataset is padding mini-batches and that triggered the assertion)

egs/librispeech/ASR/conformer_ctc/ali.py
```diff
@@ -223,21 +228,32 @@ def compute_alignments(
             use_double_scores=params.use_double_scores,
         )
 
-        ali_ids = get_alignments(best_path)
-        assert len(ali_ids) == len(cut_list)
-        for cut, ali in zip(cut_list, ali_ids):
-
-            cut.token_alignment = writer.store_array(
+        labels_ali = get_alignments(best_path, kind="labels")
+        aux_labels_ali = get_alignments(best_path, kind="aux_labels")
```
pzelasko, line 232: Does "labels" stand for BPEs/phones and "aux_labels" for words? If yes then it might be clearer to name these attrs for what they represent.

csukuangfj: It uses a CTC topology for decoding here. Labels are BPE tokens or phones, while aux labels are also BPE tokens or phones but without repeat.

For instance, if labels are

c c blk blk a a a blk t t blk

then aux labels are

c blk blk blk a blk blk blk t blk blk


The meaning of aux labels depends on the kind of FSA you are using.

If you are using a CTC graph or CTC topo, the aux labels are usually BPE tokens or phones.

If you are using an HLG graph, then the aux labels are usually words.

pzelasko: Gotcha, makes sense.

csukuangfj: The framewise alignment using BPE tokens without repeat can be used together with the embedding stuff to compute auxiliary loss(es).

csukuangfj:
The following code shows how to load pre-computed alignments

https://github.com/k2-fsa/icefall/blob/ad8a2900470807f061fc33498235b9849a4224e2/test/test_ali.py#L72-L84

pzelasko:
Note that in practice you’d rather have this code inside the dataset class, so the IO happens inside the dataloader.

csukuangfj:
> Note that in practice you’d rather have this code inside the dataset class, so the IO happens inside the dataloader.

Yes, I agree. This should be done in the `collate_fn` of the data loader.

pzelasko:
With Lhotse Datasets the batching is done inside the Dataset rather than `collate_fn`, because of cut mixing and other transforms that are best done on the manifest level. `collate_fn` is actually unused.

PR #139, opened 2021-12-02
Apply layer normalization to the output of each gate in LSTM/GRU.

csukuangfj:
- [x] Also add layer normalized GRU

---

The unit tests show that when we disable layer normalization, the customized LSTM behaves identically as `torch.nn.LSTM`

Note:
- bidirectional LSTM will not be implemented as it is not suitable for streaming
- will add dropout ~and projection~ if needed.


csukuangfj:
This is ready for review.

egs/librispeech/ASR/transducer/rnn.py
```diff
@@ -0,0 +1,652 @@
...
+        >>> cn.shape
+        torch.Size([2, 20])
+    """
+
+    def __init__(
+        self,
+        input_size: int,
+        hidden_size: int,
+        bias: bool = True,
+        ln: nn.Module = nn.LayerNorm,
```
danpovey, line 236: nn.LayerNorm is a class object, not a class instance, so I think it should be `ln: class`.
I'm not sure how we can test that these type annotations are correct.

csukuangfj: Ah, thanks.

There are some type checking tools (not mypy).

Will add it.

pzelasko: It's `Type[nn.Module]` [(`from typing import Type`)](https://docs.python.org/3.8/library/typing.html#typing.Type)

pzelasko: It's `Type[nn.Module]` [(`from typing import Type`)](https://docs.python.org/3.8/library/typing.html#typing.Type)

csukuangfj: thanks

csukuangfj:
I just borrowed the way of what espnet is doing: Use `typeguard.check_argument_types()` to check the type annotations.

The following screenshot is the error message from typeguard about the incorrect usage of the type hint.

Screen Shot 2021-12-03 at 23 58 38

---

Note I don't use `check_argument_types()` or `check_return_type()` in  the`forward()` method as torchscript reports the following error if it is used:

Screen Shot 2021-12-04 at 00 06 05

csukuangfj:
I am extending it to support packed sequence to avoid unnecessary computation for paddings in a batch.

danpovey:
OK, but what would worry me more than the total amount of computation, is having too many sequential kernels (I assume this is computed by a single kernel like regular LSTM is, I think...).  I'm not sure how the speed of this differs from regular LSTM...

csukuangfj:
Without disabling the cache, it also causes OOM on the GigaSpeech recipe even with splits on a machine with 60 GB RAM.

csukuangfj:
> OK, but what would worry me more than the total amount of computation, is having too many sequential kernels (I assume this is computed by a single kernel like regular LSTM is, I think...).

GEMM is done using two separate kernels in PyTorch, see
https://github.com/pytorch/pytorch/blob/c236247826bbf49d4f491d97bac6df1da6c1abe8/aten/src/ATen/native/RNN.cpp#L768

```cpp
    if (input.is_cuda()) {
      TORCH_CHECK(!pre_compute_input);
      auto igates = params.matmul_ih(input);
      auto hgates = params.matmul_hh(hx);
      auto result = at::_thnn_fused_lstm_cell(
          igates, hgates, cx, params.b_ih(), params.b_hh());
      // applying projections if w_hr is defined
      auto hy = params.matmul_hr(std::get<0>(result));
      // Slice off the workspace argument (it's needed only for AD).
      return std::make_tuple(std::move(hy), std::move(std::get<1>(result)));
    }
```

Pointwise addition and activation are fused into a single kernel. See
https://github.com/pytorch/pytorch/blob/c236247826bbf49d4f491d97bac6df1da6c1abe8/aten/src/ATen/native/cuda/RNN.cu#L150
```cpp
      ig = sigmoid(H2F(iig) + H2F(hig) + H2F(b1i) + H2F(b2i));
      fg = sigmoid(H2F(ifg) + H2F(hfg) + H2F(b1f) + H2F(b2f));
      cg = ::tanh(H2F(icg) + H2F(hcg) + H2F(b1c) + H2F(b2c));
      og = sigmoid(H2F(iog) + H2F(hog) + H2F(b1o) + H2F(b2o));
```

As we are using layer normalization before the activation operation, I am not sure how easy it is to fuse them into a single kernel.

PR #145, opened 2021-12-08
[POC] Add hf hub

patrickvonplaten:
Hey icefall team, 

This is a quick proof-of-concept in case you are interested in having a more direct way of downloading model checkpoints form the Hugging Face hub. 

It would allow you to directly download model checkpoints within Python and not have to rely on `git lfs`. 
If this is of interest, I would be more than happy to add tests, etc...


csukuangfj:
@danpovey What do you think?

danpovey:
I think it's a cool idea!
Would make it easier to share models, I guess..

pzelasko:
+1, thanks @patrickvonplaten

PR #147, opened 2021-12-10
Remove batchnorm

pkufool:
This pull request try to remove all the batch norm in our model,  we replace the batch normalization for features with a scaling (with a scale value 0.1), and we remove the batch normalization in `ConvolutionModule`.  The experiment result shows below:

| Setting | test-clean | test-other |
| -------|  ----| ---- |
| baseline (latest icefall model) | 2.42 | 5.73|
| replace feat_batchnorm with 0.1 scale and remove batchnorm in Convmodule| 2.47 | 5.62|

The default setting remains unchanged for backward compatibility，you can try this change by modifying `feat_batchnorm` to 0.1 in conformer_ctc/train.py

### DIFF
--- a/egs/librispeech/ASR/conformer_ctc/train.py
+++ b/egs/librispeech/ASR/conformer_ctc/train.py
@@ -17,7 +17,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 import argparse
 import logging
 from pathlib import Path
@@ -172,8 +171,12 @@ def get_params() -> AttributeDict:
 
         - subsampling_factor:  The subsampling factor for the model.
 
-        - use_feat_batchnorm: Whether to do batch normalization for the
-                              input features.
+        - use_feat_batchnorm: Normalization for the input features, can be a
+                              boolean idicating whether to do batch
+                              normalization, or a float which means just scaling
+                              the input features with this float value.
+                              If given a float value, we will remove batchnorm
+                              layer in `ConvolutionModule` as well.
 
         - attention_dim: Hidden dim for multi-head attention model.
 
--- a/egs/librispeech/ASR/conformer_ctc/train.py
+++ b/egs/librispeech/ASR/conformer_ctc/train.py
@@ -172,7 +172,7 @@ def get_params() -> AttributeDict:
         - subsampling_factor:  The subsampling factor for the model.
 
         - use_feat_batchnorm: Normalization for the input features, can be a
-                              boolean idicating whether to do batch
+                              boolean indicating whether to do batch
                               normalization, or a float which means just scaling
                               the input features with this float value.
                               If given a float value, we will remove batchnorm
--- a/egs/librispeech/ASR/conformer_ctc/transformer.py
+++ b/egs/librispeech/ASR/conformer_ctc/transformer.py
@@ -76,6 +76,7 @@ class Transformer(nn.Module):
         """
         super().__init__()
         self.use_feat_batchnorm = use_feat_batchnorm
+        assert isinstance(use_feat_batchnorm, (float, bool))
         if isinstance(use_feat_batchnorm, bool) and use_feat_batchnorm:
             self.feat_batchnorm = nn.BatchNorm1d(num_features)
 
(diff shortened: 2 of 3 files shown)


egs/librispeech/ASR/conformer_ctc/train.py
```diff
@@ -172,8 +171,12 @@ def get_params() -> AttributeDict:
 
         - subsampling_factor:  The subsampling factor for the model.
 
-        - use_feat_batchnorm: Whether to do batch normalization for the
-                              input features.
+        - use_feat_batchnorm: Normalization for the input features, can be a
+                              boolean idicating whether to do batch
```
danpovey, line 175: indicating

egs/librispeech/ASR/conformer_ctc/transformer.py
```diff
@@ -179,10 +180,15 @@ def forward(
...
-        if self.use_feat_batchnorm:
+        if (
+            isinstance(self.use_feat_batchnorm, bool)
+            and self.use_feat_batchnorm
+        ):
             x = x.permute(0, 2, 1)  # (N, T, C) -> (N, C, T)
             x = self.feat_batchnorm(x)
             x = x.permute(0, 2, 1)  # (N, C, T) -> (N, T, C)
+        if isinstance(self.use_feat_batchnorm, float):
+            x *= 0.1
```
pzelasko, line 191: I think you wanted 

x *= self.use_feat_batchnorm

pzelasko:
You could also use MVN https://github.com/lhotse-speech/lhotse/blob/1645a591f16f878318d56479c423a42ef47b57b3/lhotse/dataset/signal_transforms.py#L14

egs/librispeech/ASR/conformer_ctc/transformer.py
pkufool: Ah, yes.  Thanks.

egs/librispeech/ASR/conformer_ctc/transformer.py
```diff
@@ -72,10 +71,12 @@ def __init__(
             True to use vgg style frontend for subsampling.
           use_feat_batchnorm:
             True to use batchnorm for the input layer.
+            Float value to scale the input layer.
+            False to do nothing.
         """
         super().__init__()
         self.use_feat_batchnorm = use_feat_batchnorm
-        if use_feat_batchnorm:
+        if isinstance(use_feat_batchnorm, bool) and use_feat_batchnorm:
```
csukuangfj, line 79: How about
```python3
        if use_feat_batchnorm is True:
```

egs/librispeech/ASR/conformer_ctc/transformer.py
```diff
@@ -179,10 +180,15 @@ def forward(
...
         """
-        if self.use_feat_batchnorm:
+        if (
+            isinstance(self.use_feat_batchnorm, bool)
+            and self.use_feat_batchnorm
+        ):
             x = x.permute(0, 2, 1)  # (N, T, C) -> (N, C, T)
             x = self.feat_batchnorm(x)
             x = x.permute(0, 2, 1)  # (N, C, T) -> (N, T, C)
+        if isinstance(self.use_feat_batchnorm, float):
```
csukuangfj, line 190:
```python3
if self.use_feat_batchnorm:
            x = x.permute(0, 2, 1)  # (N, T, C) -> (N, C, T)
            x = self.feat_batchnorm(x)
            x = x.permute(0, 2, 1)  # (N, C, T) -> (N, T, C)
else:
            x *= self.use_feat_batchnorm
```

Also, we can add a check in the constructor
```python3
assert  isinstance(use_feat_batchnorm, (bool, float))
```

pkufool: I think `if self.use_feat_batchnorm` returns true if `self.use_feat_batchnorm` is a non-zero float. I will add the assertion and keep the if statements the same as before, it is more clear.

csukuangfj: I meant
```python
if self.use_feat_batchnorm is True
```

pkufool: We scale the features only when `use_feat_batchnorm` is a float. we have to test  whether `use_feat_batchnorm` is a float or not (i.e. it can be False).

PR #150, opened 2021-12-15
[WIP] A lip-reading recipe (using the dataset GRID, grid recipe) based on icefall

luomingshuang:
I try to build a lip-reading recipe with the dataset GRID (https://zenodo.org/record/3625687#.Ybn7HagzY2w, the paper: An audio-visual corpus for speech perception and automatic speech recognition) based on icefall. 

Of course, this grid recipe is just a simple lip-reading example based on icefall. I think the visual speech recognition (VSR, is the same to lip reading) can borrow much experimence from ASR. Maybe the icefall can be a good platform which can migrate some common and useful things from ASR to VSR.

In this recipe, I use character as modeling unit (token). And I place the processed data for using easily and quickly. I just use a simple model architecture called lipnet (https://arxiv.org/abs/1611.01599). Also, I borrow some codes form Dalu Feng (https://github.com/Fengdalu/LipNet-PyTorch, thanks him.). 
Now, the results (WER on val, epoch=25, avg=10) are as follows:
```
--------------------------------
    decoding-method     |  WER(%)
--------------------------------
          1best         |  15.68
--------------------------------
          nbest         |  15.68
--------------------------------
whole-lattice-rescoring |  15.76
--------------------------------
```
In GRID, the texts among different samples are very similar, such as `BIN RED IN Z EIGHT PLEASE`, `BIN GREEN AT N FOUR NOW` and so on. As the above results show, the LM (G_4_gram.fst.txt, which is built based on word unit.) seems no use for decoding, even misleads the model decoding.  So I think you don't be surprised that the whole-lattice-rescoring result is worse than the 1best result.

Some other things planned TO DO:
- [x] Build a simple VSR recipe on GRID.
- [ ] Use phone as tokens.
- [ ] Use a different model.
- [ ] Build a ASR recipe on GRID.
- [ ] Build a AVSR recipe on GRID.


pzelasko:
This is really cool! I was thinking about video support in Lhotse for quite some time but never had the time to actually do it. Trying to preserve consistency of frame rates between audio and video is challenging in general (in case you want an early fusion model, for late fusion models, especially non-streaming, this might be easier). If you'd be willing to consider contributing sth across these lines to Lhotse, I have some ideas that I can share..

luomingshuang:
> This is really cool! I was thinking about video support in Lhotse for quite some time but never had the time to actually do it. Trying to preserve consistency of frame rates between audio and video is challenging in general (in case you want an early fusion model, for late fusion models, especially non-streaming, this might be easier). If you'd be willing to consider contributing sth across these lines to Lhotse, I have some ideas that I can share..

luomingshuang:
Cool! If you have some ideas about loading the video data with Lhotse, it will be very good. Of course, the process of video data is also a complex and huge task. Because it includes many different methods and features. So how to give users enough freedom and convenience to process video based on Lhotse is a thing worth considering.

Actually, I considered loading the video data with Lhotse before, but I am not familiar with Lhotse. So I don't do it. First, I think we can just consider the non-streaming case for VSR or a fusion task. About the consistency of frame rates between audio and video, a simple way is to copy video frames to match audio frames after their respective encoding. 
  
> This is really cool! I was thinking about video support in Lhotse for quite some time but never had the time to actually do it. Trying to preserve consistency of frame rates between audio and video is challenging in general (in case you want an early fusion model, for late fusion models, especially non-streaming, this might be easier). If you'd be willing to consider contributing sth across these lines to Lhotse, I have some ideas that I can share..

luomingshuang:
@csukuangfj , I think this PR for grid recipe is ready to merge. In this recipe, I place four independent models for three tasks. The visualnet refers to the model for encoding the visual data instead of referring to some specific model, likewise audionet and combinenet. The visualnet is different from visualnet2.  
- visualnet_ctc_vsr for VSR, the WER on visual test data is 15.68%.
- visualnet2_ctc_vsr for VSR, the WER on visual test data is 13.63%.
- audionet_ctc_asr for ASR, the WER on audio test data is 2.35%. 
- combinenet_ctc_avsr for AVSR, the WER on audio-visual test data is 1.71%.

The docs and colab codes will be added after merging.

egs/grid/AVSR/local/dataset_audio.py
```diff
@@ -0,0 +1,120 @@
...
+        return len(self.data)
+
+    def _load_aud(self, filename):
+        opts = kaldifeat.FbankOptions()
+        opts.device = "cpu"
+        opts.frame_opts.dither = 0
+        opts.frame_opts.snip_edges = False
+        opts.frame_opts.samp_freq = self.sample_rate
+        opts.mel_opts.num_bins = self.feature_dim
+        fbank = kaldifeat.Fbank(opts)
```
csukuangfj, line 101: Please construct `fbank` in the constructor , i.e., in `__init__` and use `self.fbank` here.

egs/grid/AVSR/local/dataset_audio.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def _load_anno(self, name):
+        with open(name, "r") as f:
+            lines = [line.strip().split(" ") for line in f.readlines()]
+            txt = [line[2] for line in lines]
+            txt = list(filter(lambda s: not s.upper() in ["SIL", "SP"], txt))
+            txt = " ".join(txt)
+        return txt
+
+    def _padding(self, array, length):
```
csukuangfj, line 115: Please document the methods in this file.

egs/grid/AVSR/local/dataset_audio.py
```diff
@@ -0,0 +1,120 @@
...
+        array = [array[_] for _ in range(array.shape[0])]
+        size = array[0].shape
+        for i in range(length - len(array)):
+            array.append(np.zeros(size))
+        return np.stack(array, axis=0)
```
csukuangfj, line 120: Please also add tests for this class.

csukuangfj (changes requested):

pzelasko:
> Cool! If you have some ideas about loading the video data with Lhotse, it will be very good. Of course, the process of video data is also a complex and huge task. Because it includes many different methods and features. So how to give users enough freedom and convenience to process video based on Lhotse is a thing worth considering.
> 
> Actually, I considered loading the video data with Lhotse before, but I am not familiar with Lhotse. So I don't do it. First, I think we can just consider the non-streaming case for VSR or a fusion task. About the consistency of frame rates between audio and video, a simple way is to copy video frames to match audio frames after their respective encoding.
> 
> > This is really cool! I was thinking about video support in Lhotse for quite some time but never had the time to actually do it. Trying to preserve consistency of frame rates between audio and video is challenging in general (in case you want an early fusion model, for late fusion models, especially non-streaming, this might be easier). If you'd be willing to consider contributing sth across these lines to Lhotse, I have some ideas that I can share..

Sorry, I forgot about it, and only just found it in my notes to think about the video use case. The high-level idea would be to extend `Recording` with a `def load_video(self, offset: Seconds = 0, duration: Optional[Seconds] = None)` method. Internally, we might implement it by adding a `VideoSource` type and make Recording hold a list of `sources: List[Union[AudioSource, VideoSource]]`. `load_audio()` would also be able to read them from video files (and `load_video` would possibly also have a flag `with_audio: bool` to read video + audio efficiently in one pass). We would then also extend `MonoCut` and `MixedCut` with a `load_video()` method (`MixedCut` would only support "mixing" with zero-padding of videos in these cases).

... an alternative is adding a `Video` manifest, but I think it would result in much more changes in the codebase; a lot of the logic would be shared with `Recording`, and making `Cut` work with that might be a bit painful.

I think we would want to wait for torchaudio to finish implementing their ffmpeg integration as that might possibly make it much, much easier to do the actual video + audio stream reads. See: https://github.com/pytorch/audio/issues/2138

mthrok:
Regarding the ffmpeg integration in torchaudio. I am planning to land the interface in coming weeks.

The goal of this API is to provide support for video ASR and streaming inference. (and add more formats that torchaudio can support.)

Here is the detailed tutorial on the new API. If you have time, please take a look and give us feedback.
I appreciate your feedback.
https://525482-90321822-gh.circle-artifacts.com/0/docs/tutorials/streaming_api_tutorial.html#example

luomingshuang:
Cool! Will have a look and test.
> Regarding the ffmpeg integration in torchaudio. I am planning to land the interface in coming weeks.
> 
> The goal of this API is to provide support for video ASR and streaming inference. (and add more formats that torchaudio can support.)
> 
> Here is the detailed tutorial on the new API. If you have time, please take a look and give us feedback. I appreciate your feedback. https://525482-90321822-gh.circle-artifacts.com/0/docs/tutorials/streaming_api_tutorial.html#example

PR #152, opened 2021-12-18
Minor fixes to the RNN-T Conformer model

csukuangfj:
- Disable weight decay
- Remove BatchNorm layers
- Replace the BatchNorm in Conformer with LayerNorm
- Replace `relu` with `tahn` in the joiner
- Decrease the number of decoder layers from 4 to 2-layer LSTM
- Remove SOS. We only need to keep the blank symbol

I just started an experiment with this PR. Will report the results once they are available.


csukuangfj:
With the changes listed in https://github.com/k2-fsa/icefall/pull/152#issue-1083738230

The WERs of this PR is

|     | test-clean | test-other |
|-----|------------|------------|
| WER | 3.07       | 7.51       |

using `--epoch 34 --avg 11` with **greedy search**.

I have updated the [pre-trained model](https://huggingface.co/csukuangfj/icefall-asr-librispeech-transducer-bpe-500-2021-12-23) on hugging face and its usage in the [colab notebook](https://colab.research.google.com/drive/1_u6yK9jDkPwG_NLrZMN2XK7Aeq4suMO2?usp=sharing)

egs/librispeech/ASR/transducer/decode.py
```diff
@@ -401,7 +398,6 @@ def main():
 
     # <blk> and <sos/eos> are defined in local/train_bpe_model.py
     params.blank_id = sp.piece_to_id("<blk>")
```
danpovey, line 400: So is the idea that we use blank id for sos and eos as well, in the decoder?
BTW, I think the comment a line above may be wrong.

csukuangfj: > So is the idea that we use blank id for sos and eos as well, in the decoder?

I think there is only one extra blank symbol in transducer. There are no SOS and EOS.

---

> BTW, I think the comment a line above may be wrong.

Yes, you're right. I am fixing it.

PR #156, opened 2021-12-24
[Not for merge] Replace torchaudio rnnt_loss to k2 pruned rnnt loss

pkufool:
Make a pull request to show how to use k2 pruned rnnt loss (https://github.com/k2-fsa/k2/pull/891) to train a model. So that we can discuss and update experiment results here.


pkufool:
see #170

Issue #163, opened 2021-12-29
shared/make_kn_gram.py Error when sum_z1_f_z equals (or overflows to) 1.0


teowenshen:
When make_kn_gram.py analyses my transcript_token.txt, either **ZeroDivisionError** occurs at line 244 or **ValueError**: Math domain error occurs at line 353, where sum_z1_f_z became >= 1.0 due to overflow.

> Line 244: `counts_for_hist.word_to_bow[w] = (1.0 - sum_z1_f_a_z) / (1.0 - sum_z1_f_z)`
> Line 353: `line += '\t{0}'.format('%.7f' % math.log10(bow))`

I added an if statement in this program to default my bow to 1e99 if sum_z1_f_z >= 1.0. Is this fail-safe not recommended in general, or is this outlier case simply too rare to be considered ? 

```
>                         if (sum_z1_f_z < 1.0): 
>                             counts_for_hist.word_to_bow[w] = (1.0 - sum_z1_f_a_z) / (1.0 - sum_z1_f_z)
>                         else:
>                             counts_for_hist.word_to_bow[w] = 1e99
```

Background:
I am attempting to create an Icefall recipe for Japanese based on Kaldi-CSJ, Kaldi-LaboroTV and Icefall-AISHELL, so I suspect that the smaller phonetic inventory and stricter phonotactics of Japanese might have caused the polarised backoff weights.

pingfengluo:
maybe your text is empty?

teowenshen:
Actually, on the contrary, my transcript_token.txt contains around 150M worth of text, with only 41 unique tokens. I think that is what caused sum_z1_f_z to become >= 1.0 due to overflow.

danpovey:
I think sum_z1_f_z must be exactly 1.0, perhaps there were no 1-count bigrams so the discounting constant D ends up being 
 0.  This will possibly lead to unexpected things happening due to infinite backoff weights.  I think the easiest fix may be, where it does:
self.d.append(n1 * 1.0 / (n1 + 2 * n2))
we could do
 self.d.append(max(0.01, n1 * 1.0 / (n1 + 2 * n2)))
This should not affect the results measurably since the discounting constant is usually larger than 1.
We should comment that we are doing this to avoid zero discounting constant D, which could happen
if the number of symbols is small and can cause division by zero later on.

@huangruizhe,  perhaps we should make this change in Kaldi as well?

huangruizhe:
I agree with Dan.
[...]

PR #168, opened 2022-01-02
Update make_kn_lm.py

huangruizhe:
Fixed issue #163

### DIFF
--- a/icefall/shared/make_kn_lm.py
+++ b/icefall/shared/make_kn_lm.py
@@ -243,7 +243,10 @@ class NgramCounts:
                         for u in a_counts_for_hist.word_to_count.keys():  # Should be careful here: what is Z1
                             sum_z1_f_z += _counts_for_hist.word_to_f[u]
 
-                        if 1.0 - sum_z1_f_z == 0:
+                        # I assume the following to be true:
+                        # assert sum_z1_f_z <= 1.0
+                        # assert sum_z1_f_a_z <= 1.0
+                        if sum_z1_f_z < 1:
                             counts_for_hist.word_to_bow[w] = (1.0 - sum_z1_f_a_z) / (1.0 - sum_z1_f_z)
                         else:
                             counts_for_hist.word_to_bow[w] = None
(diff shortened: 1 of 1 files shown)


csukuangfj:
Is it ready for merge?

huangruizhe:
> Is it ready for merge?

Not yet. Will let you know when it is done. I think @teowenshen  will test it on his dataset.

teowenshen:
Hi @huangruizhe , thanks for the tag. 

The algorithm is stuck again at /i/. I think the conditions should be the other way round, i.e. assigning None when the denominator is zero.

Also, from my experience, some sum_z1_f_z summed up to more than 1.0, maybe due to Python's rounding, and yielded a large negative number. This caused print_as_arpa to throw a Math Domain Error at Line 353:

> `line += '\t{0}'.format('%.7f' % math.log10(bow))`

So maybe we should assign None to both when sum_z1_f_z is greater than or equal to 1.0?

icefall/shared/make_kn_lm.py
```diff
@@ -241,7 +243,10 @@ def cal_bow(self):
                         for u in a_counts_for_hist.word_to_count.keys():  # Should be careful here: what is Z1
                             sum_z1_f_z += _counts_for_hist.word_to_f[u]
 
-                        counts_for_hist.word_to_bow[w] = (1.0 - sum_z1_f_a_z) / (1.0 - sum_z1_f_z)
+                        if 1.0 - sum_z1_f_z == 0:
```
danpovey, line 246: Shouldn't this be < 0 instead of == 0?

huangruizhe: Yes. Sorry for the mistake.

huangruizhe:
@teowenshen Sorry... I made a mistake when making the PR. You are correct.

Regarding "some sum_z1_f_z summed up to more than 1.0", I agree with your guess.
I think the best way to guarantee the correctness (i.e. a valid probability distribution) is to check whether .

teowenshen:
Hi @huangruizhe , I have just implemented your fix in my code. It ran smoothly and produced this end result, which seems to be as expected. I attach the end result here too for your inspection. 
 
[P.arpa.txt](https://github.com/k2-fsa/icefall/files/7801164/P.arpa.txt)

PR #171, opened 2022-01-07
WIP: Begin to use WFST in RNN-T beam search decoding

csukuangfj:
The basic idea is when we generate a new token, we also get its score from an LG graph. Note the `olabel` of LG is not used.

It depends on https://github.com/csukuangfj/kaldifst, which provides Python APIs for OpenFST.

### DIFF
--- a/egs/librispeech/ASR/local/compile_lg.py
+++ b/egs/librispeech/ASR/local/compile_lg.py
@@ -1,0 +1,139 @@
+#!/usr/bin/env python3
+# Copyright    2021  Xiaomi Corp.        (authors: Fangjun Kuang)
+#
+# See ../../../../LICENSE for clarification regarding multiple authors
+#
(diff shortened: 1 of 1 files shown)


pzelasko:
I’m wondering why do you need openfst in addition to k2 — is it because you want to avoid having to implement some algorithms like minimization to k2?

csukuangfj:
The following compares the resulting LG with and without `minimization`:

|                                               | num_states | num_arcs | file size |
|-----------------------------------------------|------------|----------|----------:|
| detemrinize + rmepsilon                       | 5750356    | 18962186 | 255 MB    |
| determinizestar + minimizeencoded + rmepsilon | 2174063    | 9256181  | 129 MB    |

minimization is missing in k2 currently, so I use OpenFST here.

csukuangfj:
Closed via #208

PR #162, opened 2021-12-28
Use optimized_transducer to compute transducer loss.

csukuangfj:
Use https://github.com/csukuangfj/optimized_transducer to compute the transducer loss and remove the dependency on torchaudio 0.10.0.

https://github.com/csukuangfj/optimized_transducer is based on torchaudio and implements the optimization techniques
proposed in "Improving RNN Transducer Modeling for End-to-End Speech Recognition" (https://arxiv.org/abs/1909.12415)

---

[EDITED]: Need to test
- [ ] training speed
- [ ] memory consumption
- [ ] WER

PR #177, opened 2022-01-18
Refactor beam search and update results.

csukuangfj:
## Refactor beam search

As we are using a stateless decoder, the output of the decoder depends only on 
the current input, so we can use the current input as a key to the cache to improve the hit rate
and reduce computation.

## Update RESULTS.md
After continuing training for more epochs from https://github.com/k2-fsa/icefall/pull/162, we can get a lower WER on both test-clean and test-other.

### Before

|                           | test-clean | test-other | comment                                  |
|---------------------------|------------|------------|------------------------------------------|
| greedy search             | 2.77       | 7.07       | --epoch 30, --avg 13, --max-duration 100 |
| beam search (beam size 4) | 2.76       | 6.97       | --epoch 30, --avg 13, --max-duration 100|

### This PR

|                           | test-clean | test-other | comment                                  |
|---------------------------|------------|------------|------------------------------------------|
| greedy search             | 2.69       | 6.81       | --epoch 71, --avg 15, --max-duration 100 |
| beam search (beam size 4) | 2.68       | 6.72       | --epoch 71, --avg 15, --max-duration 100 |

A pre-trained model is uploaded to https://huggingface.co/csukuangfj/icefall-asr-librispeech-transducer-stateless-bpe-500-2022-01-10

You can also find decoding logs in that repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ The best WER using beam search with beam size 4 is:
 
 |     | test-clean | test-other |
 |-----|------------|------------|
-| WER | 2.76       | 6.97       |
+| WER | 2.68       | 6.72       |
 
 Note: No auxiliary losses are used in the training and no LMs are used
 in the decoding.
--- a/egs/librispeech/ASR/RESULTS.md
+++ b/egs/librispeech/ASR/RESULTS.md
@@ -13,8 +13,8 @@ The WERs are
 
 |                           | test-clean | test-other | comment                                  |
 |---------------------------|------------|------------|------------------------------------------|
-| greedy search             | 2.77       | 7.07       | --epoch 30, --avg 13, --max-duration 100 |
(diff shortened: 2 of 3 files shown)


danpovey:
Cool!
So the key is how many symbols of left-context, one or two, in this configuration?
Any idea why the 1-best numbers are slightly different?

csukuangfj:
> So the key is how many symbols of left-context, one or two, in this configuration?

In the current configuration with `--context-size 2`, the key contains two symbols.

---

> Any idea why the 1-best numbers are slightly different?

I have implemented the missing prefix-search step in beam search (see https://github.com/k2-fsa/icefall/pull/178), 
but it does not seem to be helpful.

I've no idea why the WER of test-clean is nearly the same for beam-search and greedy-search.

PR #179, opened 2022-01-20
Use modified transducer loss in training.

csukuangfj:
In **modified transducer**, we limit the maximum number of symbols per frame to 1. The following
figure compares the formula for forward and backward procedures between standard transducer
and modified transducer.

**Note**: Modified transducer is proposed independently by @danpovey .
We were later informed that the idea already existed in
[Recurrent Neural Aligner: An Encoder-Decoder Neural Network Model for
Sequence to Sequence Mapping](https://www.isca-speech.org/archive_v0/Interspeech_2017/pdfs/1705.PDF)

---

[EDITED]:

It requires [optimized_transducer](https://github.com/csukuangfj/optimized_transducer) >= v1.3
```bash
pip install --verbose optimized_transducer>=1.3
```


egs/librispeech/ASR/transducer_stateless/beam_search.py
```diff
@@ -348,47 +352,70 @@ def modified_beam_search(
...
         # fmt: off
         current_encoder_out = encoder_out[:, t:t+1, :]
+        # current_encoder_out is of shape (1, 1, encoder_out_dim)
         # fmt: on
-        A = B
+        A = list(B)
         B = HypothesisList()
 
-        joint_cache: Dict[str, torch.Tensor] = {}
+        ys_log_probs = torch.cat([hyp.log_prob.reshape(1, 1) for hyp in A])
```
csukuangfj, line 373: With the modifier transducer we can remove the inner loop in the beam search as the maximum number of symbols per frame is 1. This simplifies the decoding process significantly. 

Also, we process all hypotheses in parallel here.

csukuangfj:
Here are the results for this PR so far:

|               |                      | test-clean | test-other |                             comment |
|---------------|----------------------|------------|------------|------------------------------------:|
| this PR       | greedy search        | 2.69       | 6.75       | --epoch 57 --avg 13                 |
| this PR       | modified beam search | 2.66       | 6.66       | --epoch 57 --avg 13 --beam-size 4   |
| master branch | greedy search        | 2.69       | 6.81       | --epoch 71, --avg 15                |
| master branch | beam search          | 2.68       | 6.72       | --epoch 71, --avg 15, --beam-size 4 |

The training command for this PR is
```
export CUDA_VISIBLE_DEVICES="0,1,2,3"

./transducer_stateless/train.py \
  --world-size 4 \
  --num-epochs 60 \
  --start-epoch 0 \
  --exp-dir transducer_stateless/exp-full\
  --full-libri 1 \
  --max-duration 300 \
  --lr-factor 5 \
  --bpe-model data/lang_bpe_500/bpe.model
```
It uses modified transducer loss 25% of the time, i.e., uses `--modified-transducer-prob=0.25`.
By `modified beam search`, it means it limits the maximum number of symbols per frame to 1.

danpovey:
Great!
How about the --max-sym-per-frame 1 option in test time, I think it would be nice to have that comparison because the reason for the modified transducer is so that we can use that option to make GPU decoding faster and easier.

csukuangfj:
> Great!
> 
> How about the --max-sym-per-frame 1 option in test time, I think it would be nice to have that comparison because the reason for the modified transducer is so that we can use that option to make GPU decoding faster and easier.

The "modified beam search" in this PR uses "max symbol per frame" implicitly.

I need to check the result of greedy search of this PR used "--max-sym-per-frame=1" or not.

Caution: --max-sym-per-frame is used only by "greedy search".
"modified beam search" hardcodes its value to 1 in its implementation (you can see that there is only one loop in the time axis)

csukuangfj:
## Latest results

|                                     | test-clean | test-other | comment                                  |
|-------------------------------------|------------|------------|------------------------------------------|
| greedy search (max sym per frame 1) | 2.68       | 6.71       | --epoch 61, --avg 18, --max-duration 100 |
| greedy search (max sym per frame 2) | 2.69       | 6.71       | --epoch 61, --avg 18, --max-duration 100 |
| greedy search (max sym per frame 3) | 2.69       | 6.71       | --epoch 61, --avg 18, --max-duration 100 |
| modified beam search (beam size 4)  | 2.67       | 6.64       | --epoch 61, --avg 18, --max-duration 100 |

## Pre-trained model and decoding logs

Please see https://huggingface.co/csukuangfj/icefall-asr-librispeech-transducer-stateless-bpe-500-2022-02-07

## Colab notebook

Please see https://colab.research.google.com/drive/1CO1bXJ-2khDckZIW8zjOPHGSKLHpTDlp?usp=sharing

I am merging it.

PR #180, opened 2022-01-20
[WIP] Update Aishell with k2 pruned rnnt loss

pkufool:
The pruned loss would use the output of encoder & prediction network to get pruning range, and there is a requirement that the last dimension size of the output of the encoder & prediction network MUST be greater than or equal to vocab-size. I made some small changes to the model structure to satisfy this constrain. Will update the results when they are available.


egs/aishell/ASR/transducer_stateless/joiner.py
```diff
@@ -36,15 +38,8 @@ def forward(
...
-        assert encoder_out.ndim == decoder_out.ndim == 3
-        assert encoder_out.size(0) == decoder_out.size(0)
-        assert encoder_out.size(2) == decoder_out.size(2)
-
-        encoder_out = encoder_out.unsqueeze(2)
-        # Now encoder_out is (N, T, 1, C)
-
-        decoder_out = decoder_out.unsqueeze(1)
-        # Now decoder_out is (N, 1, U, C)
+        assert encoder_out.ndim == decoder_out.ndim == 4
```
csukuangfj, line 41: Please also update the doc of the function args.

egs/aishell/ASR/transducer_stateless/joiner.py
```diff
@@ -19,10 +19,12 @@
 
 
 class Joiner(nn.Module):
-    def __init__(self, input_dim: int, output_dim: int):
+    def __init__(self, input_dim: int, inner_dim: int, output_dim: int):
         super().__init__()
 
-        self.output_linear = nn.Linear(input_dim, output_dim)
+        self.output_linear = nn.Sequential(
+            nn.Linear(input_dim, inner_dim), nn.Linear(inner_dim, output_dim)
```
csukuangfj, line 26: There are no non-linear layers between the two `nn.Linear` layers. Is there a reason to introduce an extra `nn.Linear` layer instead of using a single `nn.Linear` layer here?

egs/aishell/ASR/transducer_stateless/model.py
```diff
@@ -38,6 +32,9 @@ def __init__(
         encoder: EncoderInterface,
         decoder: nn.Module,
         joiner: nn.Module,
+        prune_range: int = 5,
```
csukuangfj, line 35: please also update the doc for fucntions args.

egs/aishell/ASR/transducer_stateless/decoder.py
```diff
@@ -95,4 +96,5 @@ def forward(self, y: torch.Tensor, need_pad: bool = True) -> torch.Tensor:
                 assert embeding_out.size(-1) == self.context_size
             embeding_out = self.conv(embeding_out)
             embeding_out = embeding_out.permute(0, 2, 1)
+        embeding_out = self.output_linear(embeding_out)
```
csukuangfj, line 99: Both `self.conv` and `self.output_linear` can be considered as affine transforms.
Do you need to add some non-linear layers between them?

egs/aishell/ASR/transducer_stateless/joiner.py
pkufool: Using two `nn.Linear` here to reduce the num of parameters.

pkufool:
See #244

PR #184, opened 2022-01-21
[WIP] swbd + fisher recipe

pzelasko:
I only tested it on Switchboard alone, but ran the data prep for Fisher and I expect it to run. We don't have eval2000 data prep yet so it's not comparable to any numbers out there yet, but seems to converge on SWBD dev set to ~11% WER, and may be better with added Fisher data, some tweaking, and more LM work.

I'm not sure if I'll have the time to tweak it further anytime soon, but in case somebody wants to pick it up or finds it useful, I'll leave the WIP PR here.


ngoel17:
@pzelasko I am really confused about multithreading here. I ran the command (12 core 1 GPU) 
./conformer_ctc/train.py   --lr-factor 1.25   --max-duration 300   --num-workers 64   --lang-dir data/lang_bpe_500   --num-epochs 40
  and the total CPU utilization is about 50%. If I change the num-workers to 32, still the utilization is only in 50% range.
What bothers me more is the really small GPU utilization. When I run the K2 recipe for librispeech, the features are not on-the-fly, but the GPU utilization is 100% or close to that, and then training finishes on 4 GPUs in just a few hours. 
Here, I cannot fully load 1 GPU.

pzelasko:
Is it possible that you have slow disks/NFS? Reading audio possibly requires much more I/O than reading features so you might get bottlenecked by this. On the grid where I ran it, it was fast with full utilization. The solution for you might be to pre-compute the features.

I'm not sure which memory "top" displays but it might be virtual memory, in that case it will be much higher than the actual memory used. In htop that's VIRT, and RES (roughly) shows the actual occupied memory.

ngoel17:
Do you think you can get full utilization out of 2 or 4 GPUs as well? If I do it on the fly, the perturbations can be richer, though I am not sure how much real impact that will have. Fisher will have long recordings, with small segments. Will that play a role in the real-time IO?

pzelasko:
> Do you think you can get full utilization out of 2 or 4 GPUs as well?

On yet another grid (a good one) I was training GigaSpeech model with 4xV100 and on the fly features with maximum GPU usage. If you have the IO bandwidth, it does scale. 

> Fisher will have long recordings, with small segments. Will that play a role in the real-time IO?

For most audio codecs we support reading from an offset, so it even though the recordings are long, we’ll only read as much audio as needed. Shouldn’t impact IO too much.

danpovey:
Piotr is probably right; there can be a very wide range of speeds of disks, networks, etc.

ngoel17:
I ran into this: 
  File "/home/ngoel/lhotse/lhotse/audio.py", line 1136, in __init__
    assert (
AssertionError: To perform mix, energy must be non-zero and non-negative (got 0.0)
 [extra info] When calling: MixedCut.load_audio(args=(MixedCut(id='ec085737-4d9c-4443-af28-f1ea1aafa598', tracks=[MixTrack(cut=MonoCut(id='ec085737-4d9c-4443-af28-f1ea1aafa598', start=308.08, duration=4.42, channel=1, supervisions=[SupervisionSegment(id='sw2167B-ms98-a-0093', recording_id='sw02167', start=0.0, duration=4.42, channel=1, text='SUSPENSION FROM PLAYING ON UH TV AND THEN GAMES', language='English', speaker='sw02167A', gender=None, custom=None, alignment=None)], features=None, recording=Recording(id='sw02167', sources=[AudioSource(type=file, channels=[0, 1], source=/home/ngoel/fisher_k2/egs/fisher_swbd/ASR/download/LDC97S62/swb1_d1/data/sw02167.sph)], sampling_rate=16000, num_samples=7013640, duration=438.3525, transforms=[{'name': 'Resample', 'kwargs': {'source_sampling_rate': 8000, 'target_sampling_rate': 16000}}]), custom=None), offset=0.0, snr=None), MixTrack(cut=MonoCut(id='speech-librivox-0033-933-0', start=0.0, duration=5.033375, channel=0, supervisions=[], features=None, recording=Recording(id='speech-librivox-0033', sources=[AudioSource(type=file, channels=[0], source=/home/ngoel/fisher_k2/egs/fisher_swbd/ASR/download/musan/speech/librivox/speech-librivox-0033.wav)], sampling_rate=16000, num_samples=10225371, duration=639.0856875, transforms=None), custom=None), offset=0.0, snr=18.447565055203285)]),) kwargs={})

Technically it's a bug in the LDC data. Channel[1] is the 2nd channel, that got muted about 30 seconds into the conversation. Although the "echo" of the Channel[1] speaker is audible in Channel[0]. So muted channel is all 0s lead to 
0 in the source energy.  I am surprised that you could train and did not see this bug. How should we address this? Technically, the bad data should be removed from the source, but maybe we could do some logic to remove the data if it has non-NULL supervision but energy is zero?

pzelasko:
I didn't run into it but somebody else did recently. There's a [WIP PR in Lhotse](https://github.com/lhotse-speech/lhotse/pull/595), I'll either ask Ondra if he can to finish it or maybe I will.

ngoel17:
What will be the simplest way to extract features that mimic one epoch
worth of features in advance mimicking the recipe in asr_datamodule of
fisher recipe? I am thinking of using the pre-computed features initially
and then later we may switch to on-the-fly features if that helps. It's
possible that the CPUs/disks that I have are very slow.

pzelasko:
> What will be the simplest way to extract features that mimic one epoch worth of features in advance mimicking the recipe in asr_datamodule of fisher recipe? I am thinking of using the pre-computed features initially and then later we may switch to on-the-fly features if that helps. It's possible that the CPUs/disks that I have are very slow.

I'd say copy the training script and replace the training loop contents with sth that attaches the feature arrays to cuts, and writes both to disk. For example (some modifications might be needed, not sure):

```python
from lhotse import Features, LilcomHdf5Writer, CutSet

with LilcomHdf5Writer("feats.h5") as feats_writer, CutSet.open_writer("cuts_with_feats.jsonl.gz") as cuts_writer:
    for idx, cut in enumerate(batch["supervisions"]["cut"]):
        feats = batch["inputs"][idx]
        cut.features = Features(
            type="fbank", 
            num_frames=feats.shape[1], 
            num_features=feats.shape[0],
            frame_shift=0.01, 
            sampling_rate=cut.sampling_rate, 
            start=0, 
            duration=cut.duration, 
            storage_type=feats_writer.name, 
            storage_path=str(feats_writer.storage_path), 
            storage_key=feats_writer.write(cut.id, feats.numpy())
        )
        cuts_writer.write(cut)
```

PR #185, opened 2022-01-21
Fix using `lang_phone` in conformer CTC training

pzelasko:
Note: this doesn't support the attention decoder (I think maybe it could run, but might be buggy as lang_phone does not have a `<sos/eos>` token).


egs/librispeech/ASR/conformer_ctc/train.py
```diff
@@ -577,12 +586,32 @@ def run(rank, world_size, args):
...
-    )
+    if "lang_bpe" in params.lang_dir:
+        graph_compiler = BpeCtcTrainingGraphCompiler(
+            params.lang_dir,
+            device=device,
+            sos_token="<sos/eos>",
+            eos_token="<sos/eos>",
+        )
+    elif "lang_phone" in params.lang_dir:
+        assert params.att_rate == 0, (
```
csukuangfj, line 597: We can also set num_decoder_layers to 0 so that there will be no decoders.
It will reduce the number of model parameters.

pzelasko: Good point. Please check the revised version.

csukuangfj:
+2

PR #191, opened 2022-01-25
Modified attention.

csukuangfj:
Before taking the softmax of the matrix from the dot-product, we multipy it by a matrix of shape (num_heads, num_heads).


egs/librispeech/ASR/transducer_stateless/conformer.py
```diff
@@ -780,6 +789,19 @@ def multi_head_attention_forward(
             bsz * num_heads, tgt_len, -1
```
pkufool, line 789: I think we don't need this line anymore.

pzelasko:
Interesting, what’s the motivation?

csukuangfj:
> Interesting, what’s the motivation?

Maybe Dan has more to say about it.

danpovey:
more general model, make attention product full rank.

PR #192, opened 2022-01-27
Rnnt with backward decoder

pkufool:


egs/librispeech/ASR/transducer_stateless/model.py
```diff
@@ -102,24 +107,56 @@ def forward(
...
-            logits=logits,
-            targets=y_padded,
-            logit_lengths=x_lens,
-            target_lengths=y_lens,
-            blank=blank_id,
-            reduction="sum",
+        # backward loss
+        assert self.backward_decoder is not None
+        assert self.backward_joiner is not None
+        backward_decoder_out = self.backward_decoder(sos_y_padded)
```
pkufool, line 147: I think I shouldn't use `sos_y_padded` here. It should be `y[1:] + eos`

egs/librispeech/ASR/transducer_stateless/beam_search.py
```diff
@@ -294,9 +293,11 @@ def beam_search(
 
             cached_key += f"-t-{t}"
             if cached_key not in joint_cache:
-                logits = model.joiner(current_encoder_out, decoder_out)
+                logits = model.joiner(
+                    current_encoder_out, decoder_out.unsqueeze(1)
+                )
 
-                # TODO(fangjun): Ccale the blank posterior
+                # TODO(fangjun): Cache the blank posterior
```
csukuangfj, line 300: It should be `Scale`.

egs/librispeech/ASR/transducer_stateless/train.py
```diff
@@ -228,12 +261,13 @@ def get_encoder_model(params: AttributeDict):
     return encoder
 
 
-def get_decoder_model(params: AttributeDict):
+def get_decoder_model(params: AttributeDict, backward: bool = False):
```
danpovey, line 264: We should clarify the return type of this function, e.g. is it ```-> torch.Module``` ?

egs/librispeech/ASR/transducer_stateless/decoder.py
```diff
@@ -86,13 +88,20 @@ def forward(self, y: torch.Tensor, need_pad: bool = True) -> torch.Tensor:
...
-                )
+                if self.backward:
+                    assert self.context_size == 2
+                    embeding_out = F.pad(
+                        embeding_out, pad=(0, self.context_size - 1)
+                    )
+                else:
+                    embeding_out = F.pad(
+                        embeding_out, pad=(self.context_size - 1, 0)
+                    )
```
csukuangfj, line 99: I don't understand why it can use the next two symbols to predict the current one just by changing the padding direction.

pkufool: If the input is [sos, a, b, c, d] and output is [a, b, c, d, eos], padding left and using kernel-size 2, it uses left context. If the input is [a, b, c, d, eos] and output is [sos, a, b, c, d], padding right and using kernel-size 2, I think it uses right context.

csukuangfj: Maybe @danpovey can have a look at this implementation. As the kernel size is 2, the i-th output symbol always sees only the (i-1)-th and i-th symbols. It never uses its next neighbors if the input sequence is not reversed.

egs/librispeech/ASR/transducer_stateless/decoder.py
```diff
@@ -82,17 +84,30 @@ def forward(self, y: torch.Tensor, need_pad: bool = True) -> torch.Tensor:
...
-        embeding_out = self.embedding(y)
+        embedding_out = self.embedding(y)
         if self.context_size > 1:
-            embeding_out = embeding_out.permute(0, 2, 1)
+            embedding_out = embedding_out.permute(0, 2, 1)
             if need_pad is True:
-                embeding_out = F.pad(
-                    embeding_out, pad=(self.context_size - 1, 0)
-                )
+                # If the input is [sos, a, b, c, d] and output is
```
danpovey, line 91: This  comment is not very clear.  When you say "input" and "output", please clarify what variables you are talking about.  It would be nice to show what the arrays are supposed to look like after padding.

egs/librispeech/ASR/transducer_stateless/decoder.py
danpovey: @csukuangfj I think it takes care of at least part of the shift needed when he does:
```eos_y_padded = F.pad(eos_y_padded[:, 1:], pad=(0, 1), value=blank_id)```
... i.e. the 1st element of the output (y) is discarded.  I'm not crazy about the overall structure or documentation
of this code (I know we pulled it from somewhere).  Kangwei, also I'm not quite sure what you mean by "input" or "output" here.

File: egs/librispeech/ASR/transducer_stateless/model.py [hunk 6]
```diff
@@ -102,24 +128,63 @@ def forward(
...
+            encoder_out, decoder_out, ranges
+        )
+        logits = self.joiner(am_pruned, lm_pruned)
+        pruned_loss = k2.rnnt_loss_pruned(
+            logits, y_padded.to(torch.int64), ranges, blank_id, boundary
         )
 
-        return loss
+        eos_y = add_eos(y, eos_id=blank_id)
+        eos_y_padded = eos_y.pad(mode="constant", padding_value=blank_id)
```
danpovey, line 165: ... I'm not sure why we need this add_eos in RNN-T, especially since we're using blank and not real EOS.
In language modeling you need an extra EOS because it's predicted and it's part of the loss function, but
there isn't supposed to be any EOS in RNN-T's and I think the blank would just behave like extra unnecessary padding.
All this code is quite mysterious to me, IMO it needs documentation at each stage of what the matrices look like.

egs/librispeech/ASR/transducer_stateless/decoder.py
pkufool: The "input" means the sequences feeding into the decoder, "output" means the labels used to calculate loss value.  Sorry for the confusion, will update the code and documention.

egs/librispeech/ASR/transducer_stateless/decoder.py
```diff
@@ -82,17 +87,36 @@ def forward(self, y: torch.Tensor, need_pad: bool = True) -> torch.Tensor:
...
-                embeding_out = F.pad(
-                    embeding_out, pad=(self.context_size - 1, 0)
-                )
+                # Regarding the left or right context we are using,
+                # if we feed sequence [sos, a, b, c, d] to this decoder, and
+                # want to predict the sequence [a, b, c, d]. After padding to
+                # the left with context_size==2, the fed in sequence changes to
+                # [pad, sos, a, b, c, d], and we use `pad,sos` to predict `a`,
+                # `sos,a` to predict `b` ..., that is left context.
+                # if we feed sequence [b, c, d, blk, blk] to this decoder,
```
danpovey, line 100: Clarify before the "if" that you are now talking about the backwards version of the decoder.
E.g.
`For the "backwards" RNN-T: `

PR #170, opened 2022-01-07
[WIP] Replace torchaudio rnnt_loss with k2 pruned rnnt loss

pkufool:
After a month of experiments，I finally manage to make the pruned rnnt loss work. Thank for Daniel's help! 

By using the pruned version loss, we can train with larger batch size (max-duration = 300), and it only costs 2 hours to finish one epoch on full librispeech dataset (using 4 GPUs). Below are some experiment results, the WER is little worse than the model trained with torch rnnt loss, we will continue tuning the model to make it better.

For more details of the pruned rnnt loss, please see https://github.com/k2-fsa/k2/pull/891.

best WER for (greedy search) | baseline (trained with torch rnnt loss) <br> max-duration = 200 <br> 4 GPUs <br> test-clean \|\| test-other | k2 pruned rnnt loss (range = 3)<br> max-duration = 300 <br> 3 GPUs <br> test-clean \|\| test-other   | k2 pruned rnnt loss (range = 5) <br> max-duration = 300 <br> 4 GPUs<br> test-clean \|\| test-other  | k2 pruned rnnt loss (range = 8) <br> max-duration = 300 <br> 4 GPUs<br> test-clean \|\| test-other 
-- | -- | -- | -- | --
epoch 4 |   | 5.86 \|\| 14.13 | 6.75 \|\| 15.91 | 6.72 \|\| 15.34
epoch 5 |   | 5.09 \|\| 12.57 | 5.6 \|\| 13.65 | 5.59 \|\| 13.52
epoch 6 | 4.24 \|\| 10.6 | 4.83 \|\| 11.87 | 4.99 \|\| 12.38 | 5.07 \|\| 12.23
epoch 7 | 3.95 \|\| 9.87 |   | 4.65 \|\| 11.3 | 4.46 \|\| 11.5
epoch 8 | 3.64 \|\| 9.29 | 4.11 \|\| 9.23 | 4.38 \|\| 10.58 | 4.34 \|\| 10.53
epoch 9 | 3.57 \|\| 8.99 |   | 4.12 \|\| 10.02 | 4.06 \|\| 10.14
epoch 10 | 3.49 \|\| 8.69 | 3.65 \|\| 9.23 | 3.9 \|\| 9.66 | 3.82 \|\| 9.66
epoch 11 |   |   |   |  
epoch 12 | 3.24 \|\| 8.17 | 3.38 \|\| 8.56 | 3.61 \|\| 9.01 | 3.53 \|\| 8.89
[...]

PR #194, opened 2022-01-27
Use k2 pruned transducer loss to train conformer-transducer model

pkufool:
Almost the same as https://github.com/k2-fsa/icefall/pull/170, except that it adds some linear transforms in `decoder.py` and `joiner.py` to make it work for any vocab-size. The pruned loss requires the output of encoder/predictior has the shape of [B, T/S, vocab-size].

You can see the previous experiment results here https://github.com/k2-fsa/icefall/pull/170.

I need to re-check the result of this PR before merging it.


egs/librispeech/ASR/pruned_transducer_stateless/joiner.py
```diff
@@ -0,0 +1,51 @@
...
+
+import torch
+import torch.nn as nn
+
+
+class Joiner(nn.Module):
+    def __init__(self, input_dim: int, inner_dim: int, output_dim: int):
+        super().__init__()
+
+        self.inner_linear = nn.Linear(input_dim, inner_dim)
```
pkufool, line 25: A Linear layer here, to avoid a big matrix of (vocab-size, vocab-size) if we use large vocab-size.

egs/librispeech/ASR/pruned_transducer_stateless/decoder.py
```diff
@@ -0,0 +1,100 @@
...
+        if context_size > 1:
+            self.conv = nn.Conv1d(
+                in_channels=embedding_dim,
+                out_channels=embedding_dim,
+                kernel_size=context_size,
+                padding=0,
+                groups=embedding_dim,
+                bias=False,
+            )
+        self.output_linear = nn.Linear(embedding_dim, vocab_size)
```
pkufool, line 73: A Linear layer to transform output dim of decoder to vocab-size.

egs/librispeech/ASR/pruned_transducer_stateless/decode.py
```diff
@@ -0,0 +1,476 @@
...
+        }
+    )
+    return params
+
+
+def get_encoder_model(params: AttributeDict) -> nn.Module:
+    # TODO: We can add an option to switch between Conformer and Transformer
+    encoder = Conformer(
+        num_features=params.feature_dim,
+        output_dim=params.vocab_size,
```
pkufool, line 157: Change the output_dim of encoder to vocab_size.

pkufool:
The best result of this PR is as follow, we only trained 30 epochs as our GPUs are busy, you can see that the added linear layers won't affect the WERs.


Settings | WERs (test-clean) | WERs (test-other) | Notes
-- | -- | -- | --
Baseline (model without linear layers) max-duration=300 | 2.89 | 7.07 | --epoch=30, --avg=12
Model with linear layers max-duration=300 | 2.85 | 6.98 | --epoch=28, --avg=15
</div>

danpovey:
So do you think now is the right time to merge it?

pkufool:
> So do you think now is the right time to merge it?

I think it is ready now.

csukuangfj:
@pkufool 

Could you upload your pre-trained model to hugging face and provide a colab notebook to demonstrate how to use your pre-trained model?

pkufool:
OK, I think it is the same as your stateless transducer, it only changes the loss part.

desh2608:
@pkufool do you have the tensorboard for this training by any chance? I am training a pruned RNNT model on a different dataset (SPGISpeech; 5k hours) and was wondering what kind of loss values the model usually converges to?

Issue #201, opened 2022-02-07
SpecAugment change


danpovey:
This is something I'd like someone to try (perhaps someone outside our team, as our GPUs are very busy).
It consists of modifying SpecAugment behavior, which would likely require copy-and-modify of the SpecAugment 
module from Lhotse (and use the new version in asr_datamodule.py).

Right now there is a parameter max_frames_mask_fraction == 0.2, which limits how much proportion of the
utterance *each* time-masked  region can take.  I'd like this to be changed with a new default max_frames_mask_fraction = 0.4,
and made so that it limits the *total* proportion of the utterance that *all* the time-masked regions can take.
In addition, I'd like it so that, for each utterance, we first decrease num_frame_masks so that max_frames_mask_fraction
is *nearly* respected, and only then decrease max_mask_frames.

Currently the code reads (lhotse, signal_transforms.py:234):
```
            for _ in range(self.num_frame_masks):
                max_mask_frames = min(
                    self.frames_mask_size,
                    self.max_frames_mask_fraction * features.size(0),
                )
                features = mask_along_axis(
                    features.unsqueeze(0),
                    mask_param=max_mask_frames,
                    mask_value=mean,
                    axis=1,
                ).squeeze(0)
```
I'd like to modify this code by doing:
```
  _max_tot_mask_frames = self.max_frames_mask_fraction * features.size(0)
  num_frame_masks = min(self.num_frame_masks, math.ceil(_max_tot_mask_frames / self.max_mask_frames)
  max_mask_frames = min(self.max_mask_frames, _max_tot_masked_frames // num_frame_masks)
  [and then use num_frame_masks and max_mask_frames]...
```
... And also, I'd like the value we use for num_frame_masks to be increased to a largish value, like 10. 
So we'd be masking many more regions for longer utterances.  This may cause convergence problems if
it's a setup that wasn't converging very easily; #199 makes it converge way faster so combining with that
may make this easier to test, but obviously compare against the correct baseline.

luomingshuang:
Form the following codes, there is a parameter `self.max_mask_frames`, it means we should increase a new high-parameter?
``` 
  _max_tot_mask_frames = self.max_frames_mask_fraction * features.size(0)
  num_frame_masks = min(self.num_frame_masks, math.ceil(_max_tot_mask_frames / self.max_mask_frames)
  max_mask_frames = min(self.max_mask_frames, _max_tot_masked_frames // num_frame_masks)
[...]

PR #205, opened 2022-02-08
Experiments based on SpecAugment change

luomingshuang:
As a response for #201 ,  in the new asr_datamodule.py, I set `num_frame_masks` to 10 and `max_frames_mask_fraction` as 0.4 according to Dan's suggestions. Also, I plan to use conformer_ctc as the baseline. For decreasing the training time, I plan to verify this idea based on test-clean 100h data for training.

### DIFF
--- a/egs/librispeech/ASR/tdnn_lstm_ctc/asr_datamodule.py
+++ b/egs/librispeech/ASR/tdnn_lstm_ctc/asr_datamodule.py
@@ -28,7 +28,6 @@ from lhotse.dataset import (
     K2SpeechRecognitionDataset,
     PrecomputedFeatures,
     SingleCutSampler,
-    SpecAugment,
 )
 from lhotse.dataset.input_strategies import OnTheFlyFeatures
 from torch.utils.data import DataLoader
@@ -219,10 +218,11 @@ class LibriSpeechAsrDataModule:
             input_transforms.append(
                 SpecAugment(
                     time_warp_factor=self.args.spec_aug_time_warp_factor,
-                    num_frame_masks=2,
+                    num_frame_masks=10,
                     features_mask_size=27,
                     num_feature_masks=2,
                     frames_mask_size=100,
+                    max_frames_mask_fraction=0.4,
                 )
             )
         else:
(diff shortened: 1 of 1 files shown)


egs/librispeech/ASR/tdnn_lstm_ctc/asr_datamodule.py
```diff
@@ -383,3 +383,212 @@ def test_clean_cuts(self) -> CutSet:
...
+    """
+
+    def __init__(
+        self,
+        time_warp_factor: Optional[int] = 80,
+        num_feature_masks: int = 1,
+        features_mask_size: int = 13,
+        num_frame_masks: int = 1,
+        frames_mask_size: int = 70,
+        max_frames_mask_fraction: float = 0.2,
```
danpovey, line 416: You can probably change the defaults here too, since the behavior/meaning of those options has changed.
And make sure to correctly document the new meaning of the options, e.g.:
clarify frames_mask_size is per masked region, num_frame_masks is the maximum (it's also constrained by max_frames_mask_fraction), and that max_frames_mask_fraction is the total masked fraction including all masked regions (the expected/mean fraction will be less than half of this, as we randomly select the mask sizes between 0 and the computed maximum mask lengths).

danpovey:
You may be able to use a Jupyter notebook to test this and display the features, Lhotse has good interoperation with Jupyter (see if the lhotse page has any links to such notebooks).

luomingshuang:
OK.
> You may be able to use a Jupyter notebook to test this and display the features, Lhotse has good interoperation with Jupyter (see if the lhotse page has any links to such notebooks).

luomingshuang:
I test a wav sample from librispeech and display the feature with the colab url (https://colab.research.google.com/drive/1mwZJDkub5fRSTTQC53hXBQJrvTXBwMBG?usp=sharing). 

The heatmap of the original fbank feature shows as follows:
image


The heatmap of the old specaugment version (num_frame_masks: int = 1, max_frames_mask_fraction: float = 0.2, also with 
```
max_mask_frames = min(
                    self.frames_mask_size,
                    self.max_frames_mask_fraction * features.size(0),
                )
```
shows as follows:
image


The heatmap of the modified specaugment version (num_frame_masks: int = 10, max_frames_mask_fraction: float = 0.4, also with 
```
_max_tot_mask_frames = self.max_frames_mask_fraction * features.size(0)
num_frame_masks = min(self.num_frame_masks, math.ceil(_max_tot_mask_frames / self.frames_mask_size))
max_mask_frames = min(self.frames_mask_size, _max_tot_mask_frames // num_frame_masks)
```
shows as folows:
image

danpovey:
OK.  I think perhaps you can reduce max_frames_mask_fraction to 0.3; 0.4 seems to give too much masking.
BTW, that implementation does not scale efficiently for large num_frame_masks, because it copies the entire 
matrix for each mask.  It would be better to accumulate a boolean 1-d tensor and apply it as a mask later.
That would require reorganizing the code a little.
Perhaps you could work on making it more efficient, since the GPUs are very busy right now anyway?

luomingshuang:
Em, I see. In the original version, the self.num_frame_masks is 1, so it won't influence the running efficiency. When increasing the num_frame_masks, it will influence the efficiency. 
It seems that the `for` loop is not necessary. Can I remove the  `for` loop? And change 
```
            for _ in range(self.num_feature_masks):
                features = mask_along_axis(
                    features.unsqueeze(0),
                    mask_param=self.features_mask_size,
                    mask_value=mean,
                    axis=2,
                ).squeeze(0)
            for _ in range(self.num_frame_masks):
                _max_tot_mask_frames = self.max_frames_mask_fraction * features.size(0)
                num_frame_masks = min(self.num_frame_masks, math.ceil(_max_tot_mask_frames / self.frames_mask_size))
                max_mask_frames = min(self.frames_mask_size, _max_tot_mask_frames // num_frame_masks)
                
                features = mask_along_axis(
                    features.unsqueeze(0),
                    mask_param=max_mask_frames,
                    mask_value=mean,
                    axis=1,
                ).squeeze(0)
```
to 
```
            features = mask_along_axis(
                    features.unsqueeze(0),
                    mask_param=self.features_mask_size,
                    mask_value=mean,
                    axis=2,
            ).squeeze(0)
           
            _max_tot_mask_frames = self.max_frames_mask_fraction * features.size(0)
            num_frame_masks = min(self.num_frame_masks, math.ceil(_max_tot_mask_frames / self.frames_mask_size))
            max_mask_frames = min(self.frames_mask_size, _max_tot_mask_frames // num_frame_masks)
                
            features = mask_along_axis(
                    features.unsqueeze(0),
                    mask_param=max_mask_frames,
                    mask_value=mean,
                    axis=1,
             ).squeeze(0)
```
?
> OK. I think perhaps you can reduce max_frames_mask_fraction to 0.3; 0.4 seems to give too much masking. BTW, that implementation does not scale efficiently for large num_frame_masks, because it copies the entire matrix for each mask. It would be better to accumulate a boolean 1-d tensor and apply it as a mask later. That would require reorganizing the code a little. Perhaps you could work on making it more efficient, since the GPUs are very busy right now anyway?

luomingshuang:
Oh, it seems the `for` loop can not remove.
> Em, I see. In the original version, the self.num_frame_masks is 1, so it won't influence the running efficiency. When increasing the num_frame_masks, it will influence the efficiency. It seems that the `for` loop is not necessary. Can I remove the `for` loop? And change
> 
> ```
>             for _ in range(self.num_feature_masks):
>                 features = mask_along_axis(
>                     features.unsqueeze(0),
>                     mask_param=self.features_mask_size,
>                     mask_value=mean,
>                     axis=2,
>                 ).squeeze(0)
>             for _ in range(self.num_frame_masks):
>                 _max_tot_mask_frames = self.max_frames_mask_fraction * features.size(0)
>                 num_frame_masks = min(self.num_frame_masks, math.ceil(_max_tot_mask_frames / self.frames_mask_size))
>                 max_mask_frames = min(self.frames_mask_size, _max_tot_mask_frames // num_frame_masks)
>                 
>                 features = mask_along_axis(
>                     features.unsqueeze(0),
>                     mask_param=max_mask_frames,
>                     mask_value=mean,
>                     axis=1,
>                 ).squeeze(0)
> ```
> 
> to
> 
> ```
>             features = mask_along_axis(
>                     features.unsqueeze(0),
>                     mask_param=self.features_mask_size,
>                     mask_value=mean,
>                     axis=2,
>             ).squeeze(0)
>            
>             _max_tot_mask_frames = self.max_frames_mask_fraction * features.size(0)
>             num_frame_masks = min(self.num_frame_masks, math.ceil(_max_tot_mask_frames / self.frames_mask_size))
>             max_mask_frames = min(self.frames_mask_size, _max_tot_mask_frames // num_frame_masks)
>                 
>             features = mask_along_axis(
>                     features.unsqueeze(0),
>                     mask_param=max_mask_frames,
>                     mask_value=mean,
>                     axis=1,
>              ).squeeze(0)
> ```
> 
> ?
> 
> > OK. I think perhaps you can reduce max_frames_mask_fraction to 0.3; 0.4 seems to give too much masking. BTW, that implementation does not scale efficiently for large num_frame_masks, because it copies the entire matrix for each mask. It would be better to accumulate a boolean 1-d tensor and apply it as a mask later. That would require reorganizing the code a little. Perhaps you could work on making it more efficient, since the GPUs are very busy right now anyway?

danpovey:
It would be necessary to reorganize the code.  The for loop could make modifications to a 1-d bool tensor.
it's OK to do experiments with the current code though, it may not really affect efficiency much.

luomingshuang:
Ok, I am trying it.
> It would be necessary to reorganize the code. The for loop could make modifications to a 1-d bool tensor. it's OK to do experiments with the current code though, it may not really affect efficiency much.

luomingshuang:
It seems that the modified specaugment has a positive effect on the performance. I run the conformer-ctc based on the modified specaugment as this pr. 

At present, there is a comparison between the modified version and the original version when both trained for 15 epochs:
```
-----------------------------------------------
ctc-decoding, epoch=14, avg=1
-----------------------------------------------
      method\dataset |  test-clean(WER%) | test-other(WER%)
-----------------------------------------------
modified specaugment |     14.05         |  32.47
-----------------------------------------------
original specaugment |     15.15         |  33.76
-----------------------------------------------
```
Will continue train some epochs and have a see forward.

danpovey:
Cool!
BTW, I tested after 10 epochs using a librispeech-100h setup I have, training with 1 GPU.
It's a nonstandard setup, with relu on the attention and warm-steps=8000.
It was slightly worse after 10 epochs (--epoch 9 --avg 3), with WER (9.3,24.6 -> 9.85, 25.3), 
but the validation probs look better.  I will test again after more epochs, it could be it makes it
worse for the first few epochs.
I was using 0.3 for max_mask_fraction.

danpovey:
I am also, separately, testing the effect of changing p=0.5 to p=0.9 while using the original SpecAugment code.  It is only on epoch 0, I'll let you know when I have some results.

luomingshuang:
OK, and the above my results are based on max_mask_fraction=0.6.

pzelasko:
Cool! If you can report the best configuration and implementation at the end of this excercise, we could update Lhotse to reflect those.

danpovey:
I updated my comment above
https://github.com/k2-fsa/icefall/pull/205#issuecomment-1033394646
... just the baseline SpecAug but with p=0.5 changed to p=0.9 gave even more improvement.
I'm working on a combination but as I mentioned, I'm having problems with convergence right now
(the particular setup I'm using is tricky).

luomingshuang:
Yes, in the modified specaugment, when I use num_frame_masks=10, max_frames_mask_fraction=0.4, p=0.8, the model can't converge. But when p=0.5, the model can converge.

luomingshuang:
I am doing a experiment with the original specaugment but with num_frame_masks=10, max_frames_mask_fraction=0.4 and p=0.5. I want to see if there is an improvement when modifying the two parameters for the original specaugment.

danpovey:
I don't think those settings make much sense for the original SpecAugment.  num_frame_masks=10 will completely mask the signal, for signals less than about 5 seconds or so long.  And max_frames_mask_fraction is interpreted per masked region, so it will never have an effect unless the signal is less than 2.5 seconds long.

egs/librispeech/ASR/tdnn_lstm_ctc/asr_datamodule.py
```diff
@@ -383,3 +383,212 @@ def test_clean_cuts(self) -> CutSet:
...
+                features = mask_along_axis(
+                    features.unsqueeze(0),
+                    mask_param=self.features_mask_size,
+                    mask_value=mean,
+                    axis=2,
+                ).squeeze(0)
+            for _ in range(self.num_frame_masks):
+                _max_tot_mask_frames = self.max_frames_mask_fraction * features.size(0)
+                num_frame_masks = min(self.num_frame_masks, math.ceil(_max_tot_mask_frames / self.frames_mask_size))
+                max_mask_frames = min(self.frames_mask_size, _max_tot_mask_frames // num_frame_masks)
```
danpovey, line 523: @luomingshuang this code is not right.  You are setting num_frame_masks here but not doing anything with it: the code iterates by self.num_frame_masks.

luomingshuang: Em, you mean I should do changes as follows:
```
_max_tot_mask_frames = self.max_frames_mask_fraction * features.size(0)
num_frame_masks = min(self.num_frame_masks, math.ceil(_max_tot_mask_frames / self.frames_mask_size))
max_mask_frames = min(self.frames_mask_size, _max_tot_mask_frames // num_frame_masks)

for _ in range(num_frame_masks):
                features = mask_along_axis(
                    features.unsqueeze(0),
                    mask_param=max_mask_frames,
                    mask_value=mean,
                    axis=1,
                ).squeeze(0)
```

egs/librispeech/ASR/conformer_ctc/asr_datamodule.py
```diff
@@ -1 +1,594 @@
-../tdnn_lstm_ctc/asr_datamodule.py
\ No newline at end of file
+# Copyright      2021  Piotr Żelasko
```
danpovey, line 1: I think you changed this file from a soft link to a file.

luomingshuang: Yes. I changed it to a file.

egs/librispeech/ASR/conformer_ctc/asr_datamodule.py
```diff
@@ -1 +1,594 @@
...
+        logging.info("About to get test-clean cuts")
+        return load_manifest(self.args.manifest_dir / "cuts_test-clean.json.gz")
+
+    @lru_cache()
+    def test_other_cuts(self) -> CutSet:
+        logging.info("About to get test-other cuts")
+        return load_manifest(self.args.manifest_dir / "cuts_test-other.json.gz")
+
+
+import math
```
csukuangfj, line 386: Why not to move it to a separate file and then import it ?

luomingshuang: Of course. A separate file is also a good choose. Will notice it.

danpovey: This may create issues with needing to have too many soft links, if the file is in the recipes dir.
Eventually  Piotr plans to take our changes into Lhotse, that may be the easiest way...

luomingshuang:
After correcting the SpecAugment, the result has a significant progress comparing with the former or the the baseline. Will continue to do experiments based on this correct specaugment.
```
-----------------------------------------------------------------------
                  ctc-decoding, epoch=14, avg=1
-----------------------------------------------------------------------
        method\dataset       |  test-clean(WER%) | test-other(WER%)
-----------------------------------------------------------------------
modified specaugment-correct |     13.10         |  31.49
-----------------------------------------------------------------------
modified specaugment-wrong   |     14.05         |  32.47
-----------------------------------------------------------------------
     original specaugment    |     15.15         |  33.76
-----------------------------------------------------------------------
```

danpovey:
Great progress!
We can expect a little less improvement on the full Librispeech I think, but still some...

danpovey:
@luomingshuang I have updated my table above, again; I now have results with max_mask_fraction=0.3, p=0.9, but disappointingly, it is worse than the baseline code with p=0.9.  I am trying a couple other configurations.

danpovey:
Yes, I think 0.4 is too big.  You could try 0.2, see if that converges, maybe you will be able to replicate my result.

danpovey:
I have added more results to the table above.  So far my best results are with mask_fraction=0.15, p=0.9; I got 8.05,20.92, versus our original SpecAug implementation with p=0.5: 9.3/24.55

pzelasko:
Did you ever try p=1.0? I think in the original paper (https://arxiv.org/pdf/1904.08779v1.pdf) they never specified the probability of transform (`p` in the paper has - quite misleading - meaning of `mask_fraction` in this discussion), which to me implies they just applied the transform on all data.

danpovey:
> Did you ever try p=1.0? I think in the original paper (https://arxiv.org/pdf/1904.08779v1.pdf) they never specified the probability of transform (`p` in the paper has - quite misleading - meaning of `mask_fraction` in this discussion), which to me implies they just applied the transform on all data.

I tried p=0.95,time_mask_fraction=0.15.  For some reason it was substantially worse than p=0.9, time_mask_fraction=0.15, see the table above.  I am not sure how much confidence I have in this result though.  I am trying p=0.85 as well.

luomingshuang:
I update some results:
image

luomingshuang:
Comparing with the baseline (original), I just modify the SpecAugment and change the value of max_mask_fraction and p.

danpovey:
You could also test this on tedlium, if you have enough GPUs.
I am worried that your experiments and my experiments are in disagreement about whether the
max_frames_mask_fraction should be 0.3 or 0.15. We need to explore that range, preferably
on other data.

luomingshuang:
OK. Tedlium3 will be trained over more quickly than the full librispeech.

luomingshuang:
The decoding results for some epochs:
image

The above results show that the modified specaugment can accelerate the convergence of the model comparing to the original.
 
Some results based on 100h and full data:
image
image

The best result trained with full data based on modified specaugment (max_mask_fraction=0.15, p=0.9) is `2.54 | 5.24`, while the public result is `2.42 | 5.73`.

In the other hand,  I also test it with tedlium3 transducer-stateless recipe. The results show as follows (not trained for many epochs):
image

I think it's ok for us to use max-mask-fraction=0.15, p=0.9 as the fault value in the modified specaugment.

egs/librispeech/ASR/tdnn_lstm_ctc/asr_datamodule.py
```diff
@@ -383,3 +398,213 @@ def test_clean_cuts(self) -> CutSet:
...
+
+        :param time_warp_factor: parameter for the time warping; larger values mean more warping.
+            Set to ``None``, or less than ``1``, to disable.
+        :param num_feature_masks: how many feature masks should be applied. Set to ``0`` to disable.
+        :param features_mask_size: the width of the feature mask (expressed in the number of masked feature bins).
+            This is the ``F`` parameter from the SpecAugment paper.
+        :param num_frame_masks: how many frame (temporal) masks should be applied. Set to ``0`` to disable.
+        :param frames_mask_size: the width of the frame (temporal) masks (expressed in the number of masked frames).
+            This is the ``T`` parameter from the SpecAugment paper.
+        :param max_frames_mask_fraction: limits the size of the frame (temporal) mask to this value times the length
```
danpovey, line 445: Please properly document the current meaning of this option, and of num-frame_masks!

danpovey:
Oh OK, yes, you seem to have a lot of experiments; I agree that (max_frames_mask_fraction=0.15, p=0.9) are probably reasonable defaults.

@pzelasko what do you think we should do on the Lhotse side?

danpovey:
BTW, if we want to publish about this, we'll also need the baseline results for p=1.0, which corresponds to the conventional SpecAug setting.

luomingshuang:
OK, will try this setting.

PR #208, opened 2022-02-08
WIP: Use shallow fusion in modified beam search.

csukuangfj:


danpovey:
Nice!  
Something to consider here is how to handle backoff transitions and epsilons.
One way to do this is to treat backoff (e.g. epsilon) transitions as "failure" transitions, so that if there is no matching
arc, we take the epsilon and then continue the search for a matching arc.
The other way is to allow nondeterministic FSAs and either remove the epsilons, or follow the epsilons but consider them
a separate path (this may expand the number of paths a bit).
Likely the "failure" approach would be the most efficient.  It limits us to using backoff n-gram language models
for this type of decoding, but eventually we'll have an on-device decoding method that will be more flexible anyway,
so this should be fine for now.

csukuangfj:
Thanks! I like the following approach and am trying to implement it

> Something to consider here is how to handle backoff transitions and epsilons.
One way to do this is to treat backoff (e.g. epsilon) transitions as "failure" transitions, so that if there is no matching
arc, we take the epsilon and then continue the search for a matching arc. This has the advantage that the FSA becomes, in effect, deterministic, so there is no branching arising from the language model rescoring.

The current approach is using
https://github.com/k2-fsa/icefall/blob/954b4efff3ee970d22d861ba6b3930e7cdc763e7/egs/librispeech/ASR/local/compile_lg.py#L97
https://github.com/k2-fsa/icefall/blob/954b4efff3ee970d22d861ba6b3930e7cdc763e7/egs/librispeech/ASR/local/compile_lg.py#L109
https://github.com/k2-fsa/icefall/blob/954b4efff3ee970d22d861ba6b3930e7cdc763e7/egs/librispeech/ASR/local/compile_lg.py#L115

I find that it removes backoff arcs. The consequence is that at some frame it has nowhere to go since there are no out-going arcs in the current state set with a label equal to the current expanded token. As a workaround, we just skip this frame:
https://github.com/k2-fsa/icefall/blob/954b4efff3ee970d22d861ba6b3930e7cdc763e7/egs/librispeech/ASR/transducer_stateless/beam_search.py#L470-L473

But it leads to more deletions errors.

---

I just tried to use the following small corpus to visualize what the resulting `LG` looks like.

(Note: tokens.txt, words.txt, and bpe.model is from data/lang_bpe_500)

## corpus
```
HELLO WORLD
HELLOA WORLDER
HELLOA WORLDER HELLO
HELLOA WORLDER
```

## 3-gram arpa
```
\data\
ngram 1=6
ngram 2=8
ngram 3=6

\1-grams:
-99.0000000     <s>     -0.2430380
-0.6020600      HELLO   0.1549020
-0.9030900      WORLD   0.0579919
-0.4259687      </s>
-0.9030900      HELLOA  -0.0881361
-0.9030900      WORLDER 0.2798407

\2-grams:
-1.1461280      <s> HELLO       -0.1091445
-0.2430380      <s> HELLOA      -0.5070845
-0.8450980      HELLO WORLD     -0.0299632
-0.8450980      HELLO </s>
-0.5440680      WORLD </s>
-0.5440680      HELLOA WORLDER  -0.2060545
-0.8450980      WORLDER </s>
-0.8450980      WORLDER HELLO   -0.1091445

\3-grams:
-0.4771213      <s> HELLO WORLD
-0.4771213      HELLO WORLD </s>
-0.1091445      <s> HELLOA WORLDER
-0.3521825      HELLOA WORLDER </s>
-0.9542425      HELLOA WORLDER HELLO
-0.4771213      WORLDER HELLO </s>

\end\
```

## G with aux_labels being removed
G

## det_LG = determinize(compose(L_disambig, G))
det_LG

I think this graph is ready to use for shallow fusion. We can treat transitions with labels `#0`, `#1`, etc, as epsilon transitions. Before each iteration, we expand the current state set by following such transitions.


---

The current code sets all disambig IDs and #0 to 0 and applies an epsilon removal after that, which is incorrect I think since it also removes backoff transitions. The resulting LG is given below. You can see that it is easy to get stuck.
rm_epsilon_det_LG

danpovey:
RE setting backoff arcs to 0 and then removing epsilons: it isn't really incorrect, because the backoff paths are still available.  It's just that it was intended to work with much less aggressive pruning than we use in the RNN-T decoding.

csukuangfj:
| errors for test-clean                                | WER  |  Comment                                      |
|------------------------------------------------------|------|-----------------------------------------------|
|[errs-test-clean-beam_4-epoch-96319-avg-1-beam-4.txt](https://github.com/k2-fsa/icefall/files/8065926/errs-test-clean-beam_4-epoch-96319-avg-1-beam-4.txt)| 2.67 | no LG, beam size = 4                          |
|[errs-test-clean-beam_4-epoch-8-avg-1-beam-4.txt](https://github.com/k2-fsa/icefall/files/8065929/errs-test-clean-beam_4-epoch-8-avg-1-beam-4.txt)     | 2.92 | with LG, ngram lm scale = 0.1, beam size = 4  |
|[errs-test-clean-beam_4-epoch-9-avg-1-beam-4.txt](https://github.com/k2-fsa/icefall/files/8065930/errs-test-clean-beam_4-epoch-9-avg-1-beam-4.txt)     | 2.96 | with LG, ngram lm scale = 0.25, beam size = 4 |
|[errs-test-clean-beam_4-epoch-10-avg-1-beam-4.txt](https://github.com/k2-fsa/icefall/files/8065931/errs-test-clean-beam_4-epoch-10-avg-1-beam-4.txt)   | 3.61 | with LG, ngram lm scale = 1.0, beam size = 4  |

[words.txt](https://github.com/k2-fsa/icefall/files/8065925/words.txt)

Note: Don't be confused by `epoch-96319`, `epoch-8`, `epoch-9`, and `epoch-10`. They are symlinks to the same file.
The first row uses modified beam search for decoding, i..e, with `--max-sym-frame=1`, without an LG.
The remaining rows use modified beam search + LG for decoding.

The WERs with LG is a bit worse than the one without using LG. You can find the transcribed results by downloading it.

I find that there are many out-of-vocabulary words in `test-clean`. If we use LG, then it cannot handle OOV words.
The degradation in WER with LG decoding is caused mostly by OOV words.


To avoid the case that all active states die when encountering an OOV word, a backoff arc is added to each LM state connecting to the start state. The scores on these arcs are currently set to -20 since I find that `LG.scores.min()` is about `-16.6`.

danpovey:
That's interesting, thanks!

danpovey:
Eventually, we will want to subtract the neural LM scores when using scores from LG.
This will be possible in cases where we trained with the pruned loss, because in that case we have the LM scores separate from the acoustic scores (and the "lmonly" option forces them to be well normalized).

csukuangfj:
> What's the OOV rate of the LM on test-clean?

The following code is used to get the statistics of OOV information on test-clean.
```python3
#!/usr/bin/env python3

from collections import defaultdict

import k2
from lhotse import load_manifest


def main():
    cut_path = "./data/fbank/cuts_test-clean.json.gz"
    word_txt = "./data/lang_bpe_500/words.txt"

    word_table = k2.SymbolTable.from_file(word_txt)
    cut_set = load_manifest(cut_path)

    oovs = defaultdict(int)
    num_oovs = 0
    total_words = 0
    num_sentences = 0

    for c in cut_set:
        assert len(c.supervisions) == 1
        words = c.supervisions[0].text.strip().split()
        num_sentences += 1
        total_words += len(words)
        for w in words:
            if w not in word_table:
                oovs[w] += 1
                num_oovs += 1

    oovs = list(oovs.items())
    oovs = sorted(oovs, key=lambda wc: wc[1], reverse=True)
    for wc in oovs:
        print(wc[0], wc[1])
    print(f"num of oovs: {num_oovs}")
    print(f"num of words: {total_words}")
    print(f"num of sentences: {num_sentences}")
    print("oov rate: {:.4f}%".format(num_oovs * 100.0 / total_words))


if __name__ == "__main__":
    main()
```

The output is
```
BOOLOOROO 12
MONTFICHET 9
GAMEWELL 7
SERVADAC 7
FITZOOTH 7
THEL 5
BATTLEAX 5
MARGOLOTTE 5
ANDELLA 4
GHIP 4
GHISIZZLE 4
STUTELEY 4
MAINHALL 4
CHELFORD 4
PINKIES 3
ESTAFANIA 3
HOLLAN 3
BRITANNULA 3
KAFFAR 3
DOBRYNA 3
KAFFAR'S 2
RADIE 2
WHITTAWS 2
INNERLOCHY 2
LECOMPTE 2
LEOCADIA'S 2
WARRENTON'S 2
ZOOF'S 2
BEHAVIOURIST 2
CEASD 1
SMILD 1
MERCHISTON 1
TINTINT 1
THORLEIF 1
HAZEWRAPPED 1
ZOOF 1
TOLEDANS 1
RODOLFO'S 1
VALLEYED 1
WOODBEGIRT 1
CITADELLED 1
CRESSWELLS 1
PASTNESS 1
SHEWD 1
EMERG'D 1
CYN 1
HEADLONGS 1
CRAMPNESS 1
LEFRANK 1
LILLYS 1
SAILD 1
BILLYGOAT 1
REWEIGHED 1
MEADOWCROFT'S 1
SNUBNOSED 1
ANGOR 1
MEALYBACK 1
CONQUERIN 1
UNAVERRED 1
CENDENARIES 1
ADONA 1
IRWINE 1
SOUTHBRIDGE 1
GALLOWSNESS 1
CRASWELLER 1
BADAUDERIE 1
UNCASING 1
OCEANOGRAPHER 1
CRASWELLERS 1
DISINCORPORATED 1
LEOCADI 1
TEACHERY 1
MANICAMP 1
DROPIDAS 1
TENABILITY 1
MONTFICHET'S 1
LIEDENBROCK 1
EDISONIA 1
OVERLEAN 1
COMBASH 1
ARONNAX 1
ROSSETER 1
MILLION'D 1
SHARP'ST 1
INCERTAINTY 1
COMPOSSER 1
SCAROONS 1
POTFULS 1
OTHERWORLDLY 1
POOS 1
GAMEWELL'S 1
SPLENDET 1
IDIOSYNCRATICALLY 1
SCEVRA 1
SCEURA 1
VANDERPOOLS 1
ATHOLEMEN 1
CHRISTAIN 1
INTOLERANCY 1
QUINSON 1
WHITTAW 1
TREDDLESTON 1
BRITANNULISTS 1
SHOWRING 1
HORNFUL 1
MOMBI 1
SEMON'S 1
BRACTON'S 1
WHITTAWD 1
DEDALOS 1
WESTMERE 1
REBUK'D 1
SHABATA 1
NUNKIE 1
ELECTROPLATING 1
FITZOOTH'S 1
CHIAROSCURISTS 1
BEEDER 1
CHIAROSCURIST 1
FLUFFINOSE 1
ANSWERD 1
EASTERLY'S 1
TUPPENY 1
TOOMS 1
SERVADAC'S 1
GREEING 1
PEGRE 1
PEGRENNE 1
ROERER 1
TIMASCHEFF'S 1
MONCEUX 1
PIPT 1
num of oovs: 220
num of words: 52576
num of sentences: 2620
oov rate: 0.4184%
```

> (on my data) and I am wondering whether I should try LG shallow fusion.

I think it depends on the error patterns in your transcribed results.

Issue #212, opened 2022-02-11
Error when Return None as the resulting lattice is too large


wgb14:
posted decoding log here:
```
2022-02-10 22:12:52,350 INFO [decode.py:734] num_arcs before pruning: 465314
2022-02-10 22:12:52,350 INFO [decode.py:737] This OOM is not an error. You can ignore it. If your model does not converge well, or --max-duration is too large, or the input sound file is difficult to decode, you will meet this exception.
2022-02-10 22:12:52,465 INFO [decode.py:747] num_arcs after pruning: 67064
2022-02-10 22:12:52,763 INFO [decode.py:733] Caught exception:
CUDA error: out of memory
CUDA kernel errors might be asynchronously reported at some other API call,so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1.

2022-02-10 22:12:52,763 INFO [decode.py:734] num_arcs before pruning: 67064
2022-02-10 22:12:52,764 INFO [decode.py:737] This OOM is not an error. You can ignore it. If your model does not converge well, or --max-duration is too large, or the input sound file is difficult to decode, you will meet this exception.
2022-02-10 22:12:52,896 INFO [decode.py:747] num_arcs after pruning: 67030
2022-02-10 22:12:53,247 INFO [decode.py:733] Caught exception:
CUDA error: out of memory
CUDA kernel errors might be asynchronously reported at some other API call,so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1.

2022-02-10 22:12:53,247 INFO [decode.py:734] num_arcs before pruning: 67030
2022-02-10 22:12:53,247 INFO [decode.py:737] This OOM is not an error. You can ignore it. If your model does not converge well, or --max-duration is too large, or the input sound file is difficult to decode, you will meet this exception.
2022-02-10 22:12:53,399 INFO [decode.py:747] num_arcs after pruning: 67030
2022-02-10 22:12:53,724 INFO [decode.py:733] Caught exception:
CUDA error: out of memory
CUDA kernel errors might be asynchronously reported at some other API call,so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1.

2022-02-10 22:12:53,724 INFO [decode.py:734] num_arcs before pruning: 67030
[...]

PR #214, opened 2022-02-16
Incremental pruning threshold

wgb14:
closing #212

### DIFF
--- a/icefall/decode.py
+++ b/icefall/decode.py
@@ -736,8 +736,7 @@ def rescore_with_whole_lattice(
             logging.info(f"Caught exception:\n{e}\n")
             if loop_count >= max_loop_count:
                 logging.info(
-                    "Return None as the resulting lattice "
-                    "is too large"
+                    "Return None as the resulting lattice is too large."
                 )
                 return None
             logging.info(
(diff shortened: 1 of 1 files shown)


icefall/decode.py
```diff
@@ -740,16 +748,15 @@ def rescore_with_whole_lattice(
                 "is too large, or the input sound file is difficult to "
                 "decode, you will meet this exception."
             )
-
-            # NOTE(fangjun): The choice of the threshold 1e-9 is arbitrary here
-            # to avoid OOM. You may need to fine tune it.
-            inv_lattice = k2.prune_on_arc_post(inv_lattice, 1e-9, True)
+            inv_lattice = k2.prune_on_arc_post(
+                inv_lattice,
+                prune_th_list[loop_count],
```
csukuangfj, line 753: When `loop_count` is 0, it will cause access-out-of boundary error as `prune_th_list` has only 10 elements.

wgb14: I have moved `loop_count += 1` to the end of the while loop. https://github.com/wgb14/icefall/blob/ff88356ee41f900e470dafa7fe1cc9b5bbaf95a3/icefall/decode.py#L759
And when `loop_count` gets to 10, `return None` would be called before this line.
https://github.com/wgb14/icefall/blob/ff88356ee41f900e470dafa7fe1cc9b5bbaf95a3/icefall/decode.py#L737-L741
So only `prune_th_list[0:9]` are accessed.

csukuangfj: Ah, I see.

csukuangfj:
Thanks

PR #215, opened 2022-02-18
update aishell-1 recipe with k2.rnnt_loss

pingfengluo:
update aishell-1 recipe with k2.rnnt_loss and a better result (5.7 -> 5.4 ) with same number epochs compare to torchaudio.functional.rnnt_loss.
@csukuangfj  ready for review, thanks.

### DIFF
--- a/egs/aishell/ASR/transducer_stateless/model.py
+++ b/egs/aishell/ASR/transducer_stateless/model.py
@@ -120,11 +120,6 @@ class Transducer(nn.Module):
             "Please install a version >= 0.10.0"
         )
 
-        loss = k2.rnnt_loss(
-            logits,
-            y_padded,
-            blank_id,
-            boundary
-        )
+        loss = k2.rnnt_loss(logits, y_padded, blank_id, boundary)
 
         return torch.sum(loss)
--- a/egs/aishell/ASR/transducer_stateless/model.py
+++ b/egs/aishell/ASR/transducer_stateless/model.py
@@ -108,18 +108,23 @@ class Transducer(nn.Module):
         # Note: y does not start with SOS
         y_padded = y.pad(mode="constant", padding_value=0)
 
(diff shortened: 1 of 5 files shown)


csukuangfj:
Thanks! Is there a link to the pre-trained model so that others can download it?

csukuangfj:
Thanks

egs/aishell/ASR/transducer_stateless/model.py
```diff
@@ -108,18 +108,18 @@ def forward(
...
-            targets=y_padded,
-            logit_lengths=x_lens,
-            target_lengths=y_lens,
-            blank=blank_id,
-            reduction="sum",
-        )
+        loss = k2.rnnt_loss(logits, y_padded, blank_id, boundary)
 
-        return loss
+        return torch.sum(loss)
```
pkufool, line 125: Just to make you know that, the default reduction for `k2.rnnt_loss` is "mean" (the same as pytorch rnnt_loss), so you don't need a `torch.sum` here.

danpovey: I thought we were summing the loss??
If not, surely the normalization by "frames" (which we do in the diagnostics) would be wrong.

pkufool: We used reduction=="sum" in our expts.

csukuangfj: The returned `loss` is a scalar if `reduction=sum` is used, so we should change `return torch.sum(loss)` to `return loss`, even though `return torch.sum(loss)` is equivalent to `return loss` when `loss` is a scalar.

pingfengluo: Yes,  torch.sum is unnecessary here, will fix it. thanks

PR #228, opened 2022-02-28
Diagnostics code... needs refactor.

danpovey:


danpovey:
I am putting this up in the hope that someone will have time to refactor this slightly.
The key thing here is the diagnostics code, which is very useful to see what's going on in a neural net.  It should be put somewhere shared, not in the local dir, and it would be nice to add the related options to the train.py.

I think the slight refactoring in conformer.py is also safe to use: it simplifies some of the code (makes it easier to debug/change), and removes a feature that we weren't using, without affecting any compatibility.

The changes to SpecAugment, we can consider separately when and if Piotr bumps the lhotse version, so we can test that.

PR #229, opened 2022-02-28
Randomly combining intermediate layers in RNN-T training

danpovey:
This PR demonstrates how to do something like "iterated loss" using intermediate layers, but with only one
loss-function evaluation.  It does this by randomly interpolating different combinations of different layers, with
linear "adapter layers" for all but the last layer, and with the interpolation weights different per frame.

There is a significant WER improvement from this: on test-clean-100 on this setup i have 7.58/20.36 with greedy search, whereas with similar setups I was more usually getting 8.xx / 22.xx or so.
(This is hard to have an exact baseline for because the baseline didn't converge on 100 hours).

I am hoping someone could test this with a current setup somehow.


danpovey:
BTW, this will of course break compatibility with older models, so it may be necessary to introduce an option for it.
(The compatibility issue wouldn't actually affect decoding though, I think, since it's for parts that are only used during training).

csukuangfj:
I can test it with the multiple datasets setup, which does converge on the 100h subset.

danpovey:
BTW, for things like this and the diagnostics, I'd really like to have them also applied to the pruned recipe.
Right now I'm experimenting with smaller models, but with the non-pruned recipe, unfortunately i can't reduce the batch size without memory being exhausted, which is a pain (too much hassle right now to move my code over to the pruned recipe).

PR #230, opened 2022-02-28
Add diagnostics

yaozengwei:
@danpovey I have refactored the diagnostics codes. ref to #228


icefall/diagnostics.py
```diff
@@ -0,0 +1,284 @@
...
+from typing import Tuple, List
+
+
+class TensorDiagnosticOptions(object):
+    """
+    Options object for tensor diagnostics:
+
+     Args:
+        memory_limit: the maximum number of bytes per tensor (limits how many copies
+                of the tensor we cache).
```
csukuangfj, line 15: Please also update documentation about the arguments.

danpovey: Regarding print_pos_ratio: I think we can just hardcode this to true and remove the configuration.  So remove the stats_types() function and just use `["mean-abs", "pos-ratio"]`.

yaozengwei: > Regarding print_pos_ratio: I think we can just hardcode this to true and remove the configuration. So remove the stats_types() function and just use `["mean-abs", "pos-ratio"]`.
Could I modify the stats_types(self,) function to make it just return `["mean-abs", "pos-ratio"]`. I think it is safe since the stats types are fixed and could not be changed in this way. 

I am also confused about the condition `size != 31` defined in the dim_is_summarized(self, size:int) function.

danpovey: It's a heuristic about whether we want to print percentiles of the dimensions, for dimensions that don't have a "meaning" (e.g. relu, or attention heads), or whether we want to actually print the values one by one, for dimensions where the position has some kind of meaning, as in convolutional layers.

yaozengwei: > It's a heuristic about whether we want to print percentiles of the dimensions, for dimensions that don't have a "meaning" (e.g. relu, or attention heads), or whether we want to actually print the values one by one, for dimensions where the position has some kind of meaning, as in convolutional layers.

I see.. So the condition `size != 31` will be `False` and we'll print the convolutional kernel values one by one, since the   convolutional kernel size is set to 31 in Conformer.

icefall/diagnostics.py
```diff
@@ -0,0 +1,328 @@
+import random
```
csukuangfj, line 1: Please add copyright information here.

yaozengwei: Ok, it is done.

icefall/diagnostics.py
```diff
@@ -1,3 +1,21 @@
+# Copyright      2021  Xiaomi Corp.        (authors: Daniel Povey
```
csukuangfj:
```suggestion
# Copyright      2022  Xiaomi Corp.        (authors: Daniel Povey
```

yaozengwei: > 
Corrected.

PR #235, opened 2022-03-02
Add stateless transducer tutorial.

csukuangfj:
A preview is available at https://icefal1.readthedocs.io/en/doc-transducer/recipes/aishell/stateless_transducer.html

Screen Shot 2022-03-02 at 6 08 30 PM


danpovey:
Nice!
Where it says "A Conv1d layer with a default kernel size 2" we could expand to "A Conv1d layer with a default kernel size 2 (i.e. it sees 2 symbols of left-context by default)."

PR #243, opened 2022-03-09
Modify for the embedding layer in decoder

luomingshuang:
A modify for the embedding layer in decoder. Will have some experiments for it.

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless/decoder.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless/decoder.py
@@ -38,6 +38,7 @@ class Decoder(nn.Module):
         embedding_dim: int,
         blank_id: int,
         context_size: int,
+        modify_embedding_layer: bool,
     ):
         """
         Args:
@@ -50,6 +51,8 @@ class Decoder(nn.Module):
           context_size:
             Number of previous words to use to predict the next word.
             1 means bigram; 2 means trigram. n means (n+1)-gram.
+          modify_embedding_layer:
+            Do or not do modify the decoder embedding layer.
         """
         super().__init__()
         self.embedding = nn.Embedding(
@@ -57,6 +60,16 @@ class Decoder(nn.Module):
             embedding_dim=embedding_dim,
             padding_idx=blank_id,
         )
+        self.modify_embedding_layer = modify_embedding_layer
+        if self.modify_embedding_layer:
+            self.embedding.weight = nn.Parameter(
+                (
+                    torch.randn(vocab_size, embedding_dim)
+                    * (embedding_dim ** -0.5)
+                ).detach()
+            )
+
+        self.embedding_dim = embedding_dim
         self.blank_id = blank_id
 
         assert context_size >= 1, context_size
@@ -84,6 +97,9 @@ class Decoder(nn.Module):
           Return a tensor of shape (N, U, embedding_dim).
         """
         embedding_out = self.embedding(y)
+        if self.modify_embedding_layer:
+            embedding_out = self.embedding(y) * (self.embedding_dim ** 0.5)
+
         if self.context_size > 1:
             embedding_out = embedding_out.permute(0, 2, 1)
             if need_pad is True:
--- a/egs/librispeech/ASR/pruned_transducer_stateless/decoder.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless/decoder.py
@@ -52,7 +52,7 @@ class Decoder(nn.Module):
             Number of previous words to use to predict the next word.
             1 means bigram; 2 means trigram. n means (n+1)-gram.
           modify_embedding_layer:
-            Decide to modify the decoder embedding layer or not.
+            Whether modify the decoder embedding layer or not.
         """
         super().__init__()
         self.embedding = nn.Embedding(
@@ -61,12 +61,18 @@ class Decoder(nn.Module):
             padding_idx=blank_id,
         )
         self.modify_embedding_layer = modify_embedding_layer
-        self.embedding_dim_inv_sqrt = embedding_dim ** 0.5
+        self.embedding_dim_sqrt = embedding_dim ** 0.5
+        embedding_weight = torch.randn(vocab_size, embedding_dim) * (
+            1 / self.embedding_dim_sqrt
+        )
         if self.modify_embedding_layer:
-            embedding_weight = torch.randn(vocab_size, embedding_dim) * (
-                1 / self.embedding_dim_inv_sqrt
+            self.embedding = nn.Embedding(
+                num_embeddings=vocab_size,
+                embedding_dim=embedding_dim,
+                padding_idx=blank_id,
+                _weight=embedding_weight,
             )
-            self.embedding.weight = nn.Parameter(embedding_weight)
+        self.embedding._fill_padding_idx_with_zero()
 
         self.embedding_dim = embedding_dim
         self.blank_id = blank_id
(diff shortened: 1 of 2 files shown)


egs/librispeech/ASR/pruned_transducer_stateless/decoder.py
```diff
@@ -50,13 +51,25 @@ def __init__(
...
+        self.modify_embedding_layer = modify_embedding_layer
+        if self.modify_embedding_layer:
+            self.embedding.weight = nn.Parameter(
+                (
+                    torch.randn(vocab_size, embedding_dim)
+                    * (embedding_dim ** -0.5)
+                ).detach()
+            )
+
+        self.embedding_dim = embedding_dim
```
csukuangfj, line 72:
```suggestion
        self.embedding_dim_inv_sqrt = embedding_dim ** -0.5
        embedding_weight = torch.randn(vocab_size, embedding_dim) * self.embedding_dim_inv_sqrt
        self.embedding = nn.Embedding(
            num_embeddings=vocab_size,
            embedding_dim=embedding_dim,
            padding_idx=blank_id,
            _weight=embedding_weight
        )
        self.embedding._fill_padding_idx_with_zero()
```

And change line 99 to
```python3
embedding_out = self.embedding(y) * self.embedding_dim_inv_sqrt
```

egs/librispeech/ASR/pruned_transducer_stateless/decoder.py
```diff
@@ -38,6 +38,7 @@ def __init__(
         embedding_dim: int,
         blank_id: int,
         context_size: int,
+        modify_embedding_layer: bool,
```
csukuangfj, line 41: Please give it a default value `False` for backward compatibility.

csukuangfj (changes requested):

egs/librispeech/ASR/pruned_transducer_stateless/decoder.py
luomingshuang: I think is self.embedding(y) * (embedding**0.5). The self.embedding_dim_inv_sqrt is embedding**-0.5

csukuangfj: Oh, yes, you're right. But please make it a member of `self` so that it is computed only once.

egs/librispeech/ASR/pruned_transducer_stateless/decoder.py
```diff
@@ -52,7 +52,7 @@ def __init__(
             Number of previous words to use to predict the next word.
             1 means bigram; 2 means trigram. n means (n+1)-gram.
           modify_embedding_layer:
-            Do or not do modify the decoder embedding layer.
+            Decide to modify the decoder embedding layer or not.
```
csukuangfj, line 55:
```suggestion
            Whether to modify the decoder embedding layer or not.
```

egs/librispeech/ASR/pruned_transducer_stateless/decoder.py
```diff
@@ -61,13 +61,12 @@ def __init__(
             padding_idx=blank_id,
         )
         self.modify_embedding_layer = modify_embedding_layer
+        self.embedding_dim_inv_sqrt = embedding_dim ** 0.5
```
csukuangfj, line 64:
```suggestion
        self.embedding_dim_sqrt = embedding_dim ** 0.5
```

egs/librispeech/ASR/pruned_transducer_stateless/decoder.py
```diff
@@ -50,13 +51,24 @@ def __init__(
...
             embedding_dim=embedding_dim,
             padding_idx=blank_id,
         )
+        self.modify_embedding_layer = modify_embedding_layer
+        self.embedding_dim_inv_sqrt = embedding_dim ** 0.5
+        if self.modify_embedding_layer:
+            embedding_weight = torch.randn(vocab_size, embedding_dim) * (
+                1 / self.embedding_dim_inv_sqrt
+            )
+            self.embedding.weight = nn.Parameter(embedding_weight)
```
csukuangfj, line 69: Please see my previous comment about this.

Please pass it as a parameter while creating `self.embedding_weight`
and call `self.embedding._fill_padding_idx_with_zero()` afterwards.

csukuangfj (changes requested):

egs/librispeech/ASR/pruned_transducer_stateless/decoder.py
luomingshuang: Em, _weight is ok. And the self.embedding.__fill_padding_idx_with_zero() is necessary?

csukuangfj: In our current setup, the embedding for padding idx is always 0, and your current code changes that behavior.

luomingshuang: OK, I see.

luomingshuang:
@danpovey , I did some experiments based on the PR codes trained with librispeech 100 hours test-clean. And there are some results:
When setting modify-embedding=False, the tensorboard and training log:
image
[log-train-2022-03-10-11-38-52.txt](https://github.com/k2-fsa/icefall/files/8228825/log-train-2022-03-10-11-38-52.txt)


When setting modify-embedding=True, the tensorboard and training log:
image
[log-train-2022-03-10-11-39-22.txt](https://github.com/k2-fsa/icefall/files/8228831/log-train-2022-03-10-11-39-22.txt)

According to the above results, the training durations for one epoch between two methods are same (54 minutes), and it seems the training loss converges difficultly when set modify-embedding=True.
Are there some errors in my codes or disadvantages for the idea?

csukuangfj:
Do you have decoding results and diagnostic results?

csukuangfj:
The lr factor of 5.0 for single GPU training is a bit high. The pruned loss continues to decrease as can be seen from the training log, though a little slower.

BTW: The implementation is not efficient and contains unnecessary computation when using modified embedding.

danpovey:
Can you try training with --lr-factor 1.5?

luomingshuang:
OK, I will have a try based on lr-factor=1.5.

csukuangfj:
> OK, I will have a try based on lr-factor=1.5.

Also, please use a larger max duration, i.e., change 250 to 300.

luomingshuang:
When using lr-factor=1.5, the experiment based on modify-embedding=True can converge. Will continue to run some epochs.

luomingshuang:
But my question is why the experiment with a big lr-factor=5.0 based on modify-embedding=False can converge.

danpovey:
The embedding layer is one of the components that is sensitive to the learning rate because the scale matters (not canceled out later by layer norm).  Often changing the model like this will affect the range of learning rates that it converges for.

luomingshuang:
I did two experiments (modify-embedding=False and modify-embedding=True) which are both based on lr-factor=1.5. The tensorboard log shows in this url https://tensorboard.dev/experiment/eYHNN79eT5GtBtJcOQ9a6Q/#scalars, which includes two experiments's tensorboard logs. 
image

According the tensorboard log, the current loss in the experiment based on modify-embedding=True converges slightly faster than the other one (modify-embedding=False).
I test the WER for two experiments, refer the modify-embedding=True as True, and modify-embedding=False.
Greedy_search:
```
epoch: 28, avg: 15, True: 7.24 | 19.71,  False: 7.11 | 19.19
epoch: 28, avg: 20, True: 7.28 | 19.90
epoch: 29, avg: 20, True: 7.23 | 19.82
epoch: 29, avg: 21, True: 7.27 | 19.82
```
It seems the performance in False is better than in True.

danpovey:
OK, we can clos this.

PR #244, opened 2022-03-09
[WIP] Use pruned loss on Aishell

pkufool:


pkufool:
There are re-worked conformer models on Aishell, close this one.

PR #248, opened 2022-03-12
Add modified beam search for pruned rnn-t.

csukuangfj:
Training command
```bash
./pruned_transducer_stateless/train.py \
  --world-size 8 \
  --num-epochs 60 \
  --start-epoch 0 \
  --exp-dir pruned_transducer_stateless/exp \
  --full-libri 1 \
  --max-duration 300 \
  --prune-range 5 \
  --lr-factor 5 \
  --lm-scale 0.25
```

Tensorboard log https://tensorboard.dev/experiment/WKRFY5fYSzaVBHahenpNlA/

Decoding command
```bash
for epoch in 42; do
  for avg in 11; do
    ./pruned_transducer_stateless/decode.py \
      --epoch $epoch \
      --avg $avg \
      --exp-dir ./pruned_transducer_stateless/exp \
      --max-duration 100 \
      --decoding-method greedy_search \
      --max-sym-per-frame 1
  done
done

for epoch in 42; do
  for avg in 11; do
    ./pruned_transducer_stateless/decode.py \
      --epoch $epoch \
      --avg $avg \
      --exp-dir ./pruned_transducer_stateless/exp \
      --max-duration 100 \
      --decoding-method modified_beam_search \
      --beam-size 4
  done
done

for epoch in 42; do
  for avg in 11; do
    ./pruned_transducer_stateless/decode.py \
      --epoch $epoch \
      --avg $avg \
      --exp-dir ./pruned_transducer_stateless/exp \
      --max-duration 100 \
      --decoding-method beam_search \
      --beam-size 4
  done
done
```

Decoding results:

|  decoding method                                           | test-clean | test-other | comment                                                     |
|--------------------------------------------|------------|-----------|-----------------------------------------|
| greedy search (--max-sym-per-frame 1)     | 2.62            |  6.37         |--epoch 42 --avg 11 --max-duration 100 |
| greedy search (--max-sym-per-frame 2)      |   2.62         |  6.37        |--epoch 42 --avg 11 --max-duration 100 |
| greedy search (--max-sym-per-frame 3)      |   2.62         |  6.37         |--epoch 42 --avg 11 --max-duration 100 |
| modified beam search (--beam-size 4)      |  2.56            |     6.27     |--epoch 42 --avg 11 --max-duration 100 
| beam search (--beam-size 4)      |    2.57        |  6.27        |--epoch 42 --avg 11 --max-duration 100 |

Note:
1. The model is **not** trained using **modified** transducer.
[...]

PR #250, opened 2022-03-14
Add fast beam search decoding

pkufool:
This PR uses the GPU parallel decoding method in https://github.com/k2-fsa/k2/pull/926.

The decoding results and time used are as follows:

I used the same model as https://github.com/k2-fsa/icefall/pull/248. The decoding process can run with the max value of **max-duration=2000**, which means **several hundred streams** for not very long utterances.  You can see in the table, the decoding time seems not affect very much by the `beam`, `max-contexts`,  `max-states` (I think it is because we use a small decoding graph, i.e. k2.trivial_graph). So I suggest to use beam=4, max-contexts=4, max-states=8 here.

As for the WERs, it is not as good as the results in https://github.com/k2-fsa/icefall/pull/248 that decoded with `modified_beam_search`(which is 2.56||6.27).  I think one of the reasons is we use much larger batch-size, when I used `max-duration=1` to decode, I can got  WERs of 2.58 || 6.28.

Will try HLG graph to see if it can reduce the WERs any further.

beam | max-contexts | max-states | WERs for test-clean | WERs for test-other | time for test-clean(seconds) | time for test-other(seconds)
-- | -- | -- | -- | -- | -- | --
1 | -1 | -1 | 2.62 | 6.34 | 50 | 47
3 | -1 | -1 | 2.61 | 6.29 |   |  
5 | -1 | -1 | 2.61 | 6.29 | 50 | 49
7 | -1 | -1 | 2.61 | 6.29 |   |  
9 | -1 | -1 | 2.61 | 6.29 |   |  
12 | -1 | -1 | 2.61 | 6.29 | 50 | 48
15 | -1 | -1 | 2.61 | 6.29 |   |  
15 | 1 | 1 | 2.62 | 6.37 | 50 | 49
15 | 2 | 2 | 2.62 | 6.31 |   |  
15 | 2 | 4 | 2.62 | 6.31 |   |  
15 | 2 | 6 | 2.62 | 6.31 |   |  
15 | 3 | 3 | 2.61 | 6.29 |   |  
15 | 3 | 6 | 2.61 | 6.29 |   |  
15 | 3 | 9 | 2.61 | 6.29 |   |  
15 | 4 | 4 | 2.61 | 6.29 |   |  
15 | 4 | 8 | 2.61 | 6.29 |   |  
15 | 4 | 12 | 2.61 | 6.29 |   |  
15 | 5 | 5 | 2.61 | 6.29 |   |  
15 | 5 | 10 | 2.61 | 6.29 |   |  
15 | 5 | 15 | 2.61 | 6.29 |   |  
15 | 6 | 6 | 2.61 | 6.29 |   |  
15 | 6 | 12 | 2.61 | 6.29 |   |  
15 | 6 | 18 | 2.61 | 6.29 |   |  
15 | 8 | 8 | 2.61 | 6.29 |   |  
15 | 8 | 16 | 2.61 | 6.29 |   |  
15 | 8 | 24 | 2.61 | 6.29 |   |  
15 | 10 | 10 | 2.61 | 6.29 |   |  
15 | 10 | 20 | 2.61 | 6.29 |   |  
15 | 10 | 30 | 2.61 | 6.29 | 52 | 48

</div>

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless/beam_search.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless/beam_search.py
@@ -90,7 +90,7 @@ def fast_beam_search(
         logits = logits.squeeze(1).squeeze(1)
         log_probs = logits.log_softmax(dim=-1)
         decoding_streams.advance(log_probs)
-    decoding_streams.detach()
+    decoding_streams.terminate_and_flush_to_atreams()
     lattice = decoding_streams.format_output(encoder_out_lens.tolist())
 
     best_path = one_best_decoding(lattice)
--- a/egs/librispeech/ASR/pruned_transducer_stateless/beam_search.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless/beam_search.py
@@ -17,8 +17,85 @@
 from dataclasses import dataclass
 from typing import Dict, List, Optional
 
+import k2
 import torch
 from model import Transducer
+
+from icefall.decode import one_best_decoding
+from icefall.utils import get_texts
+
+
+def fast_beam_search(
+    decoding_graph: k2.Fsa,
+    model: Transducer,
+    encoder_out: torch.Tensor,
+    encoder_out_lens: torch.Tensor,
+    beam: float,
+    max_states: int,
+    max_contexts: int,
+) -> List[int]:
+    """It limits the maximum number of symbols per frame to 1.
+
+    Args:
+      model:
+        An instance of `Transducer`.
+      encoder_out:
+        A tensor of shape (N, T, C) from the encoder. Support only N==1 for now.
+      decoding_graph:
+        Decoding graph used for decoding, may be a TrivialGraph or a HLG.
+      beam:
+        Beam value, similar to the beam used in Kaldi..
+      max_states:
+        Max states per stream per frame.
+      max_contexts:
+        Max contexts pre stream per frame.
+    Returns:
+      Return the decoded result.
+    """
+    assert encoder_out.ndim == 3
+
+    context_size = model.decoder.context_size
+    vocab_size = model.decoder.vocab_size
+
+    B, T, C = encoder_out.shape
+
+    config = k2.RnntDecodingConfig(
+        vocab_size=vocab_size,
+        decoder_history_len=context_size,
+        beam=beam,
+        max_contexts=max_contexts,
+        max_states=max_states,
+    )
+    indivisual_streams = []
+    for i in range(B):
+        indivisual_streams.append(k2.RnntDecodingStream(decoding_graph))
+    decoding_streams = k2.RnntDecodingStreams(indivisual_streams, config)
+
+    for t in range(T):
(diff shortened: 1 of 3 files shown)


danpovey:
Great progress!!
I don't understand why the batch size would make a difference?  Is there anything in the algorithm that would behave differently when things are batched?
I would not necessarily regard WER differences of 0.02 or 0.03 as better or worse.  These  are well within the range of experimental noise.

csukuangfj:
> I don't understand why the batch size would make a difference?

Would the padding in the conv1d from the conformer make a difference here?

egs/librispeech/ASR/pruned_transducer_stateless/beam_search.py
```diff
@@ -17,9 +17,86 @@
...
+    encoder_out: torch.Tensor,
+    encoder_out_lens: torch.Tensor,
+    beam: float,
+    max_states: int,
+    max_contexts: int,
+) -> List[int]:
+    """It limits the maximum number of symbols per frame to 1.
+
+    Args:
+      model:
```
csukuangfj, line 40: The order of the arg doc should match the order of definition.

egs/librispeech/ASR/pruned_transducer_stateless/beam_search.py
```diff
@@ -17,9 +17,86 @@
...
+
+def fast_beam_search(
+    decoding_graph: k2.Fsa,
+    model: Transducer,
+    encoder_out: torch.Tensor,
+    encoder_out_lens: torch.Tensor,
+    beam: float,
+    max_states: int,
+    max_contexts: int,
+) -> List[int]:
```
csukuangfj, line 36:
```suggestion
) -> List[List[int]]:
```

egs/librispeech/ASR/pruned_transducer_stateless/beam_search.py
```diff
@@ -17,9 +17,86 @@
...
+    """It limits the maximum number of symbols per frame to 1.
+
+    Args:
+      model:
+        An instance of `Transducer`.
+      encoder_out:
+        A tensor of shape (N, T, C) from the encoder. Support only N==1 for now.
```
csukuangfj, line 43:
```suggestion
        A tensor of shape (N, T, C) from the encoder.
```

csukuangfj (changes requested):

egs/librispeech/ASR/pruned_transducer_stateless/beam_search.py
```diff
@@ -17,9 +17,86 @@
...
+    B, T, C = encoder_out.shape
+
+    config = k2.RnntDecodingConfig(
+        vocab_size=vocab_size,
+        decoder_history_len=context_size,
+        beam=beam,
+        max_contexts=max_contexts,
+        max_states=max_states,
+    )
+    indivisual_streams = []
```
csukuangfj, line 69: indivisual -> individual

PR #252, opened 2022-03-15
Diagnostics updates

danpovey:
This isn't really a pull request, it's an issue...
In this branch I have made a bunch of updates to the diagnostics code that was previously refactored and merged.
It would be nice if someone could do the merging.


luomingshuang:
I will do it.

PR #253, opened 2022-03-15
Cache pip packages for GitHub actions

csukuangfj:
@pzelasko You may find this PR and the following references useful:
- https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows
- https://github.com/actions/cache
- https://github.com/actions/setup-python#caching-packages-dependencies


pzelasko:
Very cool, thanks! I will try to do sth similar in Lhotse to run the notebook tutorials against mini librispeech to make sure they never go out of date again.

PR #254, opened 2022-03-15
Update diagnostics.py

luomingshuang:
As a response for #252 , I have run librispeech/transducer_stateless/train.py with print-diagnostics=True, and it runs normally.

### DIFF
--- a/icefall/diagnostics.py
+++ b/icefall/diagnostics.py
@@ -1,5 +1,6 @@
 # Copyright      2022  Xiaomi Corp.        (authors: Daniel Povey
-#                                                    Zengwei Yao)
+#                                                    Zengwei Yao
+#                                                    Mingshuang Luo)
 #
 # See ../LICENSE for clarification regarding multiple authors
 #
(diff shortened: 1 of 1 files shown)


icefall/diagnostics.py
```diff
@@ -28,51 +29,67 @@ class TensorDiagnosticOptions(object):
...
+    """
+    Returns the specified transformation of the Tensor (either x or x.abs()
+    or (x > 0), summed over all but the index `dim`.
 
     Args:
-      x:
-        Tensor, tensor to be analyzed
-      dim:
-        Dimension with 0 <= dim < x.ndim
+       x: Tensor, tensor to be analyzed
```
csukuangfj, line 59: Please use the previous style for docstring.

See https://google.github.io/styleguide/pyguide.html#383-functions-and-methods for more details.

icefall/diagnostics.py
```diff
@@ -240,7 +299,7 @@ class ModelDiagnostic(object):
         Options object.
     """
 
-    def __init__(self, opts: TensorDiagnosticOptions):
+    def __init__(self, opts: TensorDiagnosticOptions = TensorDiagnosticOptions()):
```
csukuangfj, line 302: If `opts` is not a readonly member, please change it to

```python3
def __init__(self, opts: Optional[TensorDiagnosticOptions] = None):
  if opts is None:
    self.opts = TensorDiagnosticOptions()
  else:
    self.opts = opts
```

csukuangfj (changes requested):

PR #262, opened 2022-03-21
Implement greedy search in batch mode for transducer decoding.

csukuangfj:
Now if you use
```bash
--decoding-method greedy_search
--max-sym-per-frame 1
```
in decoding, it will decode in batch mode.

For the following deocoding command
```
    ./pruned_transducer_stateless/decode.py \
      --epoch 99 \
      --avg 1 \
      --exp-dir ./pruned_transducer_stateless/exp \
      --max-duration 2000 \
      --decoding-method greedy_search \
      --max-sym-per-frame 1
```

The output is given below:
Screen Shot 2022-03-21 at 10 19 55 PM

**note**: Don't be confused by `epoch-99.pt`, it is just a symlink to the pretrained model from https://github.com/k2-fsa/icefall/pull/248

You can see that the total time in decoding test-clean and test-other is less than 1 minute. The RTF is roughly computed as below:
```
>>> 60 / ((5.3 + 5.4)*60*60)
0.001557632398753894
```
(test-clean has 5.4 hours of data and test-other has 5.3 hours)

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless/beam_search.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless/beam_search.py
@@ -106,7 +106,7 @@ def fast_beam_search(
 def greedy_search(
     model: Transducer, encoder_out: torch.Tensor, max_sym_per_frame: int
 ) -> List[int]:
-    """
+    """Greedy search for a single utterance.
     Args:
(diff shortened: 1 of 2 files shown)


danpovey:
nice!

PR #277, opened 2022-03-30
Add LG decoding

pkufool:
**Conclusion first :** I think our decoding method works well with the general  FSA graphs (using `log-add` and perhaps doing log weight pushing on the graph).

I did some experiments with an LG graph on our RNN-T decoding method, the results show that using `log-add` to merge hypotheses make a big difference on WERs in LG decoding.  I also found that doing weight pushing (i.e. weight pushing in determinize) for the LG graph helps to search for better results.  I summary the results as follows: (**Note:** For trivial graph, I used beam=4.0, max-contexts=4, max-states=8, nbest-path=200, nbest-scale=0.8, for LG graph, I used beam=8.0, max-contexts=8, max-states=64,ngram-lm-scale=0.01,nbest-path=200, nbest-scale=0.8)

|                              LibriSpeech                                             |merge operation | test-clean|test-other| 
|--------------------------------------------------------|-----------------|-----------|----------|
| greedy search (max-sym-per-frame=1)                            |/                            |       2.72     |       6.87 |                  
| FSA-based beam search (trivial graph)                             | max                     |       2.71     |  6.79 | 
| FSA-based beam search (trivial graph)                             | log-add              |     2.68    |  6.77 | 
|FSA-based beam search (LG graph)                                   | max                   |        7.25        |11.08 | 
|FSA-based beam search (LG graph)                                   | log-add            | 3.53 | 6.96 | 
|FSA-based beam search (LG graph with tropical weight pushing)| max                   | 5.01 | 8.43 | 
|FSA-based beam search (LG graph with tropical weight pushing)| log-add             | 3.53 |  6.95| 
|FSA-based beam search (LG graph with log weight pushing)| max                   |5.21 | 8.7 | 
|FSA-based beam search (LG graph with log weight pushing)| log-add             |3.53 | 6.94| 

**Note:** All the results are decoded from the same model, do not care too much about the absolute value of WERs, it may not be our best model.

You can see the results decoded with an LG graph are worse than trivial graph, I think it is because the OOV words in LibriSpeech dataset, I also tried the above experiments on Aishell dataset, see the results in the following table, it shows that   FSA-based LG decoding achieves the lowest CER. (**Note:** For trivial graph, I used beam=4.0, max-contexts=4, max-states=8, nbest-path=100, nbest-scale=0.8, for LG graph, I used beam=8.0, max-contexts=8, max-states=64,ngram-lm-scale=0.01,nbest-path=100, nbest-scale=0.8)

|                Aishell     |merge operation | test| 
| ------------------|-----------------|-----------|
| greedy search (max-sym-per-frame=1)|/ | 6.05 |
| FSA-based beam search (trivial graph)| max | 6.01|
| FSA-based beam search (trivial graph)| log-add | 6.0| 
|FSA-based beam search (LG graph)| max | 13.11 | 
|FSA-based beam search (LG graph)| log-add| 5.98 |
|FSA-based beam search (LG graph with tropical weight pushing)| max | 8.6 |
|FSA-based beam search (LG graph with tropical weight pushing)| log-add| 5.96 |
|FSA-based beam search (LG graph with log weight pushing)| max | 8.86 |
|FSA-based beam search (LG graph with log weight pushing)| log-add| 5.95 |

I also tried length-normalization, it makes no difference on WERs/CERs no matter using `max` or `log-add`.

@danpovey  Regarding the weight-pushing thing, does it make sense to help decreasing WERs/CERs,  I tried the weight pushing because I noticed that there are many deletions at the tail when using `max`,  and I saw many true paths in the lattice but they can not survive shortest path search.


pkufool:
This PR requires the latest k2, as it need `k2.linear_fsa_with_self_loops`.

danpovey:
Nice!
The effects regarding max vs. log-add may possibly relate to multiple paths through the language model, as well as
effects from multiple alignments.

pkufool:
Merging, as LG decoding is needed by phone/pinyin based models.

PR #278, opened 2022-03-30
Use Emformer as RNN-T encoder.

csukuangfj:
We `import` Emformer from `torchaudio` directly at present and use it to replace the Conformer model for the current RNN-T setup.

### DIFF
--- a/egs/aishell/ASR/transducer_emformer/asr_datamodule.py
+++ b/egs/aishell/ASR/transducer_emformer/asr_datamodule.py
@@ -1,0 +1,1 @@
+../transducer_stateless/asr_datamodule.py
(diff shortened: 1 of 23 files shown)


csukuangfj:
Here is a recording about the streaming decoding using RNN-T greedy search with Emformer.


https://user-images.githubusercontent.com/5284924/161257041-c54f4aa4-312a-484a-9ece-3dd0e3d2ca08.mov


The number of model parameters is: 65390556

The WERs  using `--epoch 29 --avg 6 --decoding_method modified_beam_search` is 4.58/11.54 (test-clean/test-other),
which is very close to the one listed in https://github.com/pytorch/audio/tree/main/examples/asr/emformer_rnnt, i.e., 4.56 for test-clean.

pzelasko:
Great news!

csukuangfj:
I am merging it to the `streaming` branch first.

 @yaozengwei will work on the streaming decoding part with beam search (--max-sym-per-frame=1)

PR #281, opened 2022-03-30
fix typo of torch.eig

glynpu:
Is "torch.eigs" a typo of "torch.eig"? 

BTW, according to https://pytorch.org/docs/1.11/generated/torch.eig.html?highlight=torch%20eig#torch.eig, it's better to use "torch.linalg.eigvals".

### DIFF
--- a/icefall/diagnostics.py
+++ b/icefall/diagnostics.py
@@ -140,7 +140,7 @@ def get_diagnostics_for_dim(
             stats = eigs.abs().sqrt()
         except:  # noqa
             print("Error getting eigenvalues, trying another method.")
-            eigs, _ = torch.eigs(stats)
+            eigs = torch.linalg.eigvals(stats)
             stats = eigs.abs().sqrt()
         # sqrt so it reflects data magnitude, like stddev- not variance
     elif sizes_same:


icefall/diagnostics.py
```diff
@@ -140,7 +140,7 @@ def get_diagnostics_for_dim(
             stats = eigs.abs().sqrt()
         except:  # noqa
             print("Error getting eigenvalues, trying another method.")
-            eigs, _ = torch.eigs(stats)
+            eigs = torch.linalg.eigvals(stats)
```
danpovey, line 143: torch.linalg.eigvals does not exist in torch 1.7.1.  I think this needs to be "torch.eig"

danpovey: I had this error:
```
  File "/ceph-dan/icefall/icefall/diagnostics.py", line 139, in get_diagnostics_for_dim
    eigs, _ = torch.symeig(stats)
RuntimeError: "symeig_cuda" not implemented for 'Half'

During handling of the above exception, another exception occurred:

  File "./pruned2_knowledge/train.py", line 997, in <module>
    main()
  File "./pruned2_knowledge/train.py", line 990, in main
    run(rank=0, world_size=1, args=args)
  File "./pruned2_knowledge/train.py", line 916, in run
    diagnostic.print_diagnostics()
  File "/ceph-dan/icefall/icefall/diagnostics.py", line 325, in print_diagnostics
    self.diagnostics[k].print_diagnostics()
  File "/ceph-dan/icefall/icefall/diagnostics.py", line 297, in print_diagnostics
    print_diagnostics_for_dim(self.name, dim, tensors, self.opts)
  File "/ceph-dan/icefall/icefall/diagnostics.py", line 218, in print_diagnostics_for_dim
    s = get_diagnostics_for_dim(
  File "/ceph-dan/icefall/icefall/diagnostics.py", line 143, in get_diagnostics_for_dim
    eigs = torch.linalg.eigvals(stats)
AttributeError: module 'torch.linalg' has no attribute 'eigvals'

```

glynpu: Ah, I was using 1.10.0.

danpovey: Also, above,  eigs, _ = torch.symeig(stats) needs to be  eigs, _ = torch.symeig(stats.to(torch.float32)), and this one here needs to be eigs, _ = torch.eig(stats.to(torch.float32)), for compatibility with the --use-fp16 option.

danpovey: ... actually, forget the .to(torch.float32)).  It's easier to just declare that the --use-fp16=True option is incompatible with --print-diagnostics=True, and check this.  We get inf's after summing the gradients in fp16.

PR #283, opened 2022-03-31
[WIP] Half precision training

pkufool:
I am trying half precision training on our pruned rnnt recipe according to the documents here (https://pytorch.org/docs/stable/notes/amp_examples.html).

Now, the training runs normally,  I can run it with `max-duration = 550`, and it takes about 1 hour 10 minutes to finish one epoch with 4 GPUs (for full precision training with `max-duration = 300`, this takes around 2 hours). 

It seems that the loss values are a little larger than the full precision training (at the beginning of epoch 2, 0.13 vs. 0.18). Here is the tensorboard: https://tensorboard.dev/experiment/pZgHoziMSXaoZxM6zhaoWA/#scalars

Corresponding fixes in k2:  https://github.com/k2-fsa/k2/pull/943

**[edit]:** Oops, the training has been unstable at the middle of epoch 2 ...

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless/model.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless/model.py
@@ -128,17 +128,18 @@ class Transducer(nn.Module):
         boundary[:, 2] = y_lens
         boundary[:, 3] = x_lens
 
-        simple_loss, (px_grad, py_grad) = k2.rnnt_loss_smoothed(
-            lm=decoder_out,
-            am=encoder_out,
-            symbols=y_padded,
-            termination_symbol=blank_id,
-            lm_only_scale=lm_scale,
-            am_only_scale=am_scale,
-            boundary=boundary,
-            reduction="sum",
-            return_grad=True,
-        )
+        with torch.autocast(device_type=x.device.type, enabled=False):
+            simple_loss, (px_grad, py_grad) = k2.rnnt_loss_smoothed(
+                lm=decoder_out.float(),
+                am=encoder_out.float(),
+                symbols=y_padded,
+                termination_symbol=blank_id,
+                lm_only_scale=lm_scale,
+                am_only_scale=am_scale,
+                boundary=boundary,
+                reduction="sum",
+                return_grad=True,
+            )
 
         # ranges : [B, T, prune_range]
         ranges = k2.get_rnnt_prune_ranges(
@@ -157,13 +158,14 @@ class Transducer(nn.Module):
         # logits : [B, T, prune_range, C]
         logits = self.joiner(am_pruned, lm_pruned)
 
-        pruned_loss = k2.rnnt_loss_pruned(
-            logits=logits,
-            symbols=y_padded,
-            ranges=ranges,
-            termination_symbol=blank_id,
-            boundary=boundary,
-            reduction="sum",
-        )
+        with torch.autocast(device_type=x.device.type, enabled=False):
+            pruned_loss = k2.rnnt_loss_pruned(
+                logits=logits.float(),
+                symbols=y_padded,
+                ranges=ranges,
+                termination_symbol=blank_id,
+                boundary=boundary,
+                reduction="sum",
+            )
 
         return (simple_loss, pruned_loss)
--- a/egs/librispeech/ASR/pruned_transducer_stateless/model.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless/model.py
@@ -128,7 +128,7 @@ class Transducer(nn.Module):
         boundary[:, 2] = y_lens
         boundary[:, 3] = x_lens
 
-        with torch.autocast(device_type=x.device.type, enabled=False):
+        with torch.cuda.amp.autocast(enabled=False):
             simple_loss, (px_grad, py_grad) = k2.rnnt_loss_smoothed(
                 lm=decoder_out.float(),
                 am=encoder_out.float(),
@@ -158,7 +158,7 @@ class Transducer(nn.Module):
         # logits : [B, T, prune_range, C]
         logits = self.joiner(am_pruned, lm_pruned)
 
-        with torch.autocast(device_type=x.device.type, enabled=False):
+        with torch.cuda.amp.autocast(enabled=False):
             pruned_loss = k2.rnnt_loss_pruned(
                 logits=logits.float(),
                 symbols=y_padded,
--- a/egs/librispeech/ASR/pruned_transducer_stateless/train.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless/train.py
@@ -647,9 +647,7 @@ def train_one_epoch(
         params.batch_idx_train += 1
         batch_size = len(batch["supervisions"]["text"])
 
-        with torch.autocast(
-            device_type=model.device.type, enabled=params.use_fp16
-        ):
+        with torch.cuda.amp.autocast(enabled=params.use_fp16):
             loss, loss_info = compute_loss(
                 params=params,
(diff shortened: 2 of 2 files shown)


pkufool:
I conducted several experiments and finally found out where was the problem of the unstable training. When I excluded all the loss computation parts from mix precision training, it is running normally now. See the tensorboard here: https://tensorboard.dev/experiment/uq5wVaBZQOuqM2mcDcVytA/#scalars.

**Note:** For this stable training, I **did not** make any changes to our rnnt loss.  Thanks to our fast and memory-efficient implementation, I can also use `max-duration=550` and it only costs 1 hour 12 minutes to finish one epoch on full librispeech using 4 GPUs.

Some simple conclusions from my experiments:

- The unstable training does not related to pruning, I tried our unpruned version, loss also exploded. 
- It also has nothing to do with `GradScaler`.
- torchaudio rnnt_loss also produced `nan` loss from epoch 17 (Our explosion occurs at the middle of epoch 2).

The WERs until now:

epoch | avg | sort_by_clean | avg | sort_by_other
-- | -- | -- | -- | --
15 | 6 | 3.41 \|\| 8.65 | 4 | 3.42 \|\| 8.57
16 | 4 | 3.36 \|\| 8.34 | 4 | 3.36 \|\| 8.34
17 | 4 | 3.3 \|\| 8.21 | 5 | 3.35 \|\| 8.19
18 | 4 | 3.22 \|\| 7.98 | 5 | 3.24 \|\| 7.94
19 | 6 | 3.19 \|\| 7.82 | 5 | 3.22 \|\| 7.8

pkufool:
@danpovey  I thought we should only exclude the recursion part in our `rnnt_loss`, but it seems that we have to exclude all the loss computation section. What do you think is the problem?  Does it relate to the normalization?

danpovey:
@pkufool  I am not surprised that the loss computation requires full precision.  IIRC I may even have coded the core recursion to use double always, out of concern about roundoff (not sure).  I think the issue may be that in the "simple joiner", to compute normalization factors we have to do an exp-log-sum for each (t,s) pair, and this is implemented by shifting the log values to avoid numerical overflow, doing exp(), and then doing matrix multiplication.  This requires quite a few bits in the exponent of the floating point value, because there may otherwise be cases where we get zeros here, or at least where the computed normalizer is far away from the true normalizer.  The easiest fix here is to simply keep this part of the computation in float32 (i.e. k2.rnnt_loss_simple or whatever).

pkufool:
Ok, In my current experiments, I keep all the loss computation parts in float32. And it doesn't affect too much about the speed and memory consumption (i.e. need more 2 ~ 3 minutes for each epoch).

danpovey:
I think that's fine.  I'd do the joiner in float16 and all the loss-computation (k2.whatever) in float.
It's great that it's so much faster.  My PR here https://github.com/k2-fsa/icefall/pull/288 is I think in the
region of 35% faster than the current setup per epoch, so hopefully these 2 speedups will combine for super-fast training.
(Plus my setup needs way fewer epochs, will update results there).

desh2608:
It seems `torch.autocast()` is only available for PyTorch versions 1.10 and above. For older versions, `torch.cuda.amp.autocast()` should be used.

pkufool:
@danpovey  Baseline (i.e. current master code), uses 4 GPUs, s_range = 5, it takes about 2 hours 6 ~ 8 minutes for one epoch.

image

This PR. uses 4 GPUs, s_range=5, it takes around 1 hour 12 ~ 13 minutes for one epoch.
image

pkufool:
> It seems `torch.autocast()` is only available for PyTorch versions 1.10 and above. For older versions, `torch.cuda.amp.autocast()` should be used.

Oh, I didn't notice that. Thanks!

pkufool:
The WERs currently got, a little worse than the full precision training (i.e. at epoch 36, 2.82||7.06 vs. 2.77||6.92).
**BTW**  I got `inf` loss from epoch 42.

epoch | avg | sort_by_clean | avg | sort_by_other
-- | -- | -- | -- | --
8 | 4 | 4.33 \|\| 10.82 | 4 | 4.33 \|\| 10.82
9 | 4 | 4.17 \|\| 10.25 | 4 | 4.17 \|\| 10.25
10 | 4 | 3.9 \|\| 9.8 | 4 | 3.9 \|\| 9.8
11 | 4 | 3.78 \|\| 9.37 | 4 | 3.78 \|\| 9.37
12 | 5 | 3.65 \|\| 9.24 | 4 | 3.67 \|\| 9.13
13 | 4 | 3.48 \|\| 8.96 | 5 | 3.59 \|\| 8.92
14 | 5 | 3.42 \|\| 8.72 | 5 | 3.42 \|\| 8.72
15 | 6 | 3.41 \|\| 8.65 | 4 | 3.42 \|\| 8.57
16 | 4 | 3.36 \|\| 8.34 | 4 | 3.36 \|\| 8.34
17 | 4 | 3.3 \|\| 8.21 | 5 | 3.35 \|\| 8.19
18 | 4 | 3.22 \|\| 7.98 | 5 | 3.24 \|\| 7.94
19 | 6 | 3.19 \|\| 7.82 | 5 | 3.22 \|\| 7.8
20 | 7 | 3.15 \|\| 7.69 | 5 | 3.18 \|\| 7.6
21 | 6 | 3.09 \|\| 7.61 | 7 | 3.11 \|\| 7.6
22 | 6 | 3.05 \|\| 7.5 | 6 | 3.05 \|\| 7.5
23 | 8 | 3.04 \|\| 7.44 | 8 | 3.04 \|\| 7.44
24 | 6 | 2.99 \|\| 7.53 | 8 | 3.03 \|\| 7.36
25 | 7 | 2.96 \|\| 7.4 | 10 | 2.99 \|\| 7.36
26 | 8 | 2.91 \|\| 7.31 | 8 | 2.91 \|\| 7.31
27 | 6 | 2.9 \|\| 7.35 | 9 | 2.92 \|\| 7.22
28 | 9 | 2.9 \|\| 7.23 | 9 | 2.9 \|\| 7.23
29 | 11 | 2.89 \|\| 7.23 | 10 | 2.9 \|\| 7.21
30 | 10 | 2.88 \|\| 7.19 | 9 | 2.89 \|\| 7.16
31 | 13 | 2.87 \|\| 7.18 | 11 | 2.89 \|\| 7.15
32 | 10 | 2.84 \|\| 7.16 | 11 | 2.85 \|\| 7.15
33 | 11 | 2.81 \|\| 7.13 | 11 | 2.81 \|\| 7.13
34 | 13 | 2.82 \|\| 7.11 | 11 | 2.85 \|\| 7.09
35 | 10 | 2.81 \|\| 7.08 | 10 | 2.81 \|\| 7.08
36 | 14 | 2.81 \|\| 7.06 | 16 | 2.84 \|\| 7.05
37 | 14 | 2.82 \|\| 7.06 | 9 | 2.87 \|\| 7.04
38 | 13 | 2.83 \|\| 7.03 | 11 | 2.85 \|\| 7.0
39 | 13 | 2.84 \|\| 6.98 | 13 | 2.84 \|\| 6.98
40 | 16 | 2.83 \|\| 7.0 | 11 | 2.87 \|\| 6.97
41 | 20 | 2.84 \|\| 6.95 | 20 | 2.84 \|\| 6.95

danpovey:
I think slightly-worse WERs, on the same iter, are expected due to fewer minibatches overall.  I.e. it may take a few more epochs to reach the same point. 
BTW we are trying increasing the epsilons in LayerNorm and Adam, to something larger than the min representable float16 value.

pkufool:
Half precision training has been merged in another PR.

PR #287, opened 2022-04-01
Modify icefall/__init__.py.

yaozengwei:
Modify icefall/__init__.py to import common functions and classes defined in icefall/utils.py. Hence we can directly import them from `icefall` instead of `icefall.utils`.


csukuangfj:
Please fix the style issues.

PR #288, opened 2022-04-02
Reworked model

danpovey:
I am putting this up so you can see the work I have been doing on the faster-to-train, better, reworked version of the conformer.
I don't have WERs for this specific version of the model yet, I'll put them up soon.


danpovey:
The full-libri is run with *almost* this PR (I will give the exact numbers for this exact PR when they are ready).
```
Full-libri:
Tested while running "epoch 13", with --avg-last-n=7, last checkpoint=80000
 greedy_search  3.04    best for test-clean
 greedy_search  7.3 best for test-other
 Tested while running "epoch 13", with --avg-last-n=4, last checkpoint=80000
 greedy_search  3.0 best for test-clean
 greedy_search  7.36    best for test-other
 Tested while running "epoch 14", with --avg-last-n=5, last checkpoint=84000
  greedy_search 2.94    best for test-clean
  greedy_search 7.2 best for test-other
 --epoch 17 --avg 5
 greedy_search  2.84    best for test-clean
 greedy_search  6.74    best for test-other

Libri-100:
WER at --epoch 19 --avg 8
 greedy_search  7.01    best for test-clean
 greedy_search  18.34   best for test-other
WER at --epoch 29 --avg 8
 greedy_search  6.77    best for test-clean
 greedy_search  17.67   best for test-other
```

danpovey:
At --epoch 19 --avg 8, with `2.83/6.63`, we're close to the RESULTS.md numbers of `2.67/6.67` from --epoch 63 --avg 19.  So nearly 3-fold speedup in terms of epochs!  (probably closer to 4-fold if we take into account faster training time per epoch).  Hopefully when combined with the mixed-precision stuff, we'll be able to claim something like 7 to 8-fold reduction in training time!

danpovey:
perl -e 'print ((80*6.0) / (66*8.0))'
0.909090909090909

danpovey:
I am trying to work out how much speedup there is per epoch from these changes.. it's not as much as I thought (although it's possible that I was just unlucky with the machine I was training on).
Fangjun's experiment with 8 GPUs is here 
https://tensorboard.dev/experiment/4kygy8qCQVWmY2lreHz29Q/#scalars&_smoothingWeight=0
which shows between 1h6m and 1h12m, I'm going to take 1h8m as the baseline.  My experiment seems to be
taking about 1h18m per epoch, with 6 GPUs.  Assuming the effect of num-GPUs on speed is roughly linear,
the speed difference would be:
 ((78*6.0) / (68*8.0)) = 0.8602
so this setup would take 14% less time per epoch than the baseline.  But an experiment with 8 GPUs, measuring the time taken per step, suggests to me that there may be a 10% penalty from going from 6 to 8 GPUs, apart from the expected linear speedup, so in that case the speedup would be less, maybe around 6%.  [But it's possible that I was just unlucky with the machine, or my disk was overloaded.]
It's odd, because I made some changes that, individually, seemed to provide a worthwhile speedup.

pkufool:
I tried mix-precision training with 8 GPUs, it takes 38 ~ 39 minutes to finish an epoch.

image

danpovey:
BTW, the reason that I'm  not proposing to merge this right away is that I'm working on a simpler version of this recipe where we get rid of the complex learning-rate schedule with warmup, and simplify the model initialization.

danpovey:
The reason I am not proposing to merge this right now is, I am trying a further cleanup/modification to this recipe where we use a much simpler learning rate schedule and initialization, and an optimizer that tries to keep the non-scalar parameters at a fixed rms value (0.1).  That way the relationship of the learning-rate schedule and the actual rate of parameter movement will become more intuitive.

danpovey:
Here WER results after the latest reworking.  It is run with something like this:
```
python3 ./pruned_transducer_stateless2/train.py --exp-dir=pruned_transducer_stateless2/exp_1a --world-size 1 --num-epochs 30  --start-epoch 0 --full-libri 0 --max-duration 300 
```
... with only --world-size and --num-epochs being possibly changed; the learning-rate setup is designed so that you don't have to change it when you change the --world-size.

With libri-100,
```
            num-jobs       test_clean||test_other
                                    epoch=19,avg=8             epoch=29,avg=8         epoch=39,avg=10
             1                       7.12||18.42
             2                      7.05||18.77                6.82||18.14                 6.81||17.66
             4                      7.31||19.55                 7.08||18.59                6.86||18.29
```
More details of libri-100 expts: 
[note: due to a bug that should be fixed now, the learning rate plot cannot be seen.]
1 job: /ceph-dan/icefall/egs/librispeech/ASR/rework2p_refactor https://tensorboard.dev/experiment/AhnhooUBRPqTnaggoqo7lg/
2 jobs:  /ceph-dan/icefall/egs/librispeech/ASR/rework2p_refactor_ws2  https://tensorboard.dev/experiment/dvOC9wsrSdWrAIdsebJILg/
4 jobs:  /ceph-dan/icefall/egs/librispeech/ASR/rework2p_refactor_ws4  https://tensorboard.dev/experiment/a3T0TyC0R5aLj5bmFbRErA/


With full librispeech
```
         num-jobs          test_clean||test_other
                                epoch=19,avg=8
               8                   2.75||6.53
```
This was run here: /ceph-dan/icefall/egs/librispeech/ASR/rework2p_full https://tensorboard.dev/experiment/UKI6z9BvT6iaUkXPxex1OA/

csukuangfj:
Note: I just uploaded the pretrained model, training logs, decoding logs, and decoding results for this PR to
https://huggingface.co/csukuangfj/icefall-asr-librispeech-pruned-transducer-stateless2-2022-04-29

desh2608:
@danpovey is there some documentation about the major changes between the original conformer and the "reworked" conformer?

danpovey:
Not really, I'm afraid.  I got rid of the batchnorm and replaced it with a form of LayerNorm that has a large trainable epsilon, that's probably the most significant change.  Also I gave all parameters a trainable scalar scale (it's trained in log space).  At some point I may push this into the optimizer though.

desh2608:
Thanks!

wangers:
@danpovey In reworked conformer, it use post-norm, and basicnorm is always company with an ActivationBalancer, which means to constrain median of output to be close to zero. However, in  Conv2dSubsampling it is basicnorm->ActivationBalancer, while in conformer blocks it is ActivationBalancer->basicnorm. 
Conv2dSubsampling :
```
        x = self.out_norm(x)
        x = self.out_balancer(x)
```
conformer blocks:
`        src = self.norm_final(self.balancer(src))`
Is there some explanation to how to desin a basicnorm-activationbalancer such as tdnn-relu-basicnorm?

danpovey:
The order of the ActivationBalancer versus BasicNorm makes little difference because the BasicNorm in this case makes ake very little difference to the scale of the output of that component.  In fact, I had forgotten that there was still a normalization component in there, I don't think it is necessary.  I will remove it in a future version of the recipe.

wangers:
Thanks. However, only the last of blocks have basicnorm, remove it means there is no norm layers anymore. or just left a basicnorm as post-norm.

danpovey:
I think there was always an extra normalization layer in the convolution module, even before my refactoring, although it may have been in the middle somewhere.
Likely what happened is I wasn't able to remove that normalization without the model diverging, at that time, but since then we have made a lot of other changes that help stability, so I think this won't be an issue in future.

PR #289, opened 2022-04-03
Support specifying iteration number of checkpoints for decoding.

csukuangfj:
Assume you have the following checkpoints:
- checkpoint-100.pt, checkpoint-500.pt, checkpoint-1000.pt
- epoch-0.pt, epoch-1.pt, epoch-2.pt, epoch-3.pt

If you want to use `checkpoint-1000.pt and checkpoint-500.pt` for decoding, you can use
```
--iter 1000 --avg 2
```

If you want to use `epoch-3.pt, epoch-2.pt, and epoch-1.pt` for decoding, you can use
```
--epoch 3 --avg 3
```


danpovey:
LGTM, thanks!

PR #294, opened 2022-04-06
update icefall/__init__.py to import more common functions.

yaozengwei:


icefall/__init__.py
```diff
@@ -1,3 +1,34 @@
+from .checkpoint import (
```
danpovey, line 1: I would also add to this file:
```
from . import  checkpoint
```
and the same for other local files, so people can use the qualified names.
IMO, just doing this (instead of importing all the names) would be more future proof, because then we 
could add different versions of the same function in different files later if we need.

yaozengwei: OK, I see.

PR #295, opened 2022-04-06
Export stateless pruned transducer to jit script module

pkufool:


pkufool:
See how to use the exported jit model here https://github.com/k2-fsa/k2/pull/947.

PR #301, opened 2022-04-08
Modify init

yaozengwei:
Add some imports like `from . import checkpoint`.


icefall/__init__.py
```diff
@@ -1,3 +1,10 @@
+# isort:skip_file
+from . import checkpoint
```
csukuangfj, line 2: To be consistent, you can use
```python3
from . import (
  checkpoint,
  decode,
  dist,
  env,
  utils
)
```

yaozengwei: ok

danpovey:
In future, whenever we modify icefall code, I think it would be better to do "import icefall" and then refer to icefall things as icefall.something.something or possibly icefall.something if they are common things.  I think this is a clearer approach.  But it may cause too much merging hassle if we modify existing example scripts.

PR #302, opened 2022-04-10
Add streaming feature extractor.

csukuangfj:
This PR uses the online feature extractor from `kaldifeat` (https://github.com/csukuangfj/kaldifeat/blob/master/kaldifeat/python/kaldifeat/online_feature.py), 
which is a re-implementation of Kaldi's OnlineGenericBaseFeature (https://github.com/kaldi-asr/kaldi/blob/master/src/feat/online-feature.h), without depending on Kaldi.

It supports only greedy search for a single utterance at present. Will extend it to support batches.


egs/librispeech/ASR/transducer_emformer/streaming_decode.py
```diff
@@ -147,126 +147,196 @@ def get_parser():
...
-        # since it re-computes the features for the right context.
-        chunk = audio_samples[i : i + chunk_size]  # noqa
-        i += segment_length
-        if chunk.size(0) < chunk_size:
-            chunk = torch.nn.functional.pad(
-                chunk, pad=(0, chunk_size - chunk.size(0))
-            )
-        features = fbank(chunk)
-        feature_lens = torch.tensor([features.size(0)], device=params.device)
+        # Simulate streaming.
```
csukuangfj, line 326: The code here shows how the online feature extractor works.

We can send an arbitrary number of samples to the extractor. There is a buffer in the extractor recording the samples that have not been used in computing features.

egs/librispeech/ASR/transducer_emformer/streaming_feature_extractor.py
```diff
@@ -0,0 +1,106 @@
...
+    opts = FbankOptions()
+    opts.device = "cpu"
+    opts.frame_opts.dither = 0
+    opts.frame_opts.snip_edges = False
+    opts.frame_opts.samp_freq = 16000
+    opts.mel_opts.num_bins = 80
+    return OnlineFbank(opts)
+
+
+class Stream(object):
```
danpovey, line 43: maybe calling this FeatureExtractionStream would be clearer.

egs/librispeech/ASR/transducer_emformer/emformer.py
```diff
@@ -27,6 +27,66 @@
 LOG_EPSILON = math.log(1e-10)
 
 
+def unstack_states(
```
csukuangfj, line 30: `unstack_states` and `stack_states` take the advantage that Emformer also outputs the states for the next chunk.

`unstack_states` is useful to split the states for a batch into a state list corresponding to separate utterances in the batch, which can be later stacked together using `stack_states`.

egs/librispeech/ASR/transducer_emformer/streaming_decode.py
```diff
@@ -147,126 +147,363 @@ def get_parser():
...
+    stream_list = StreamList(
+        batch_size=batch_size,
+        context_size=params.context_size,
+        blank_id=params.blank_id,
+    )
 
-            decoder_out = model.decoder(decoder_input, need_pad=False)
-        logging.info(f"Partial result:\n{sp.decode(hyp[context_size:])}")
+    while not streaming_audio_samples.done:
+        samples = streaming_audio_samples.get_next()
```
csukuangfj, line 486: This simulates that audio samples of different streams come in a streaming fashion.

egs/librispeech/ASR/transducer_emformer/streaming_feature_extractor.py
csukuangfj: Thanks, fixed.

PR #305, opened 2022-04-11
Support mix precision training on the reworked model

pkufool:
Add mix precision training support for https://github.com/k2-fsa/icefall/pull/288.

**Note:** I only verified the code before the new learning rate scheduler was added. 
For the mix precision training for the latest code, I only ran 1 epoch on libri-100 to make sure it works.


danpovey:
Will test

danpovey:
One epoch on libri-100 with 1 job and --max-duration=500 seems to take just 28 minutes!

PR #307, opened 2022-04-11
Update results for tedlium3 pruned RNN-T

luomingshuang:
This PR is for doing a update for tedlium3 pruned RNN-T. By the way, the colab examples have been updated.


csukuangfj:
Also, please add modified beam search and fast beam search to the notebook when you have time.

PR #338, opened 2022-04-28
Validate generated manifest files.

csukuangfj:
See also https://github.com/k2-fsa/icefall/issues/320#issuecomment-1108728769

> ... as a side note, maybe it makes sense to add a separate script in each recipe that validates the major assumptions about the data (single supervision per cut, supervisions time bounds == cut time bounds, maybe sth text normalization related like compatibility of BPE model with the supervision texts, etc.)

@pzelasko Please have a review.


pzelasko (approved):
Looks good!

PR #348, opened 2022-05-06
Keep model_avg on cpu

yaozengwei:
Now `model_avg` is kept on cpu instead of cuda device, to avoid extra cuda memory usage.


egs/librispeech/ASR/pruned_transducer_stateless4/train.py
```diff
@@ -867,10 +867,6 @@ def run(rank, world_size, args):
         model = DDP(model, device_ids=[rank])
     model.device = device
 
-    if rank == 0:
-        model_avg.to(device)
-        model_avg.device = device
-
```
danpovey, line 873: Have you tested with this change?
I wonder whether you need to do `.to(torch.device('cpu'))` when updating this.

yaozengwei: Yes, I have tested it. `model_avg` is on cpu default, since it is a deepcopy of `model` before the line `model.to(device)`.

egs/librispeech/ASR/pruned_transducer_stateless4/train.py
```diff
@@ -868,7 +868,7 @@ def run(rank, world_size, args):
     model.device = device
 
     if rank == 0:
-        model_avg.to(device)
+        model_avg.to(torch.device("cpu"))
         model_avg.device = device
```
csukuangfj, line 873: These lines can be removed if you want to keep `model_avg` on CPU.

egs/librispeech/ASR/pruned_transducer_stateless4/train.py
danpovey: OK.

egs/librispeech/ASR/pruned_transducer_stateless4/train.py
yaozengwei: OK.

csukuangfj:
> This is a bit confusing, if we are setting a .device that is different from the device it is actually on.

We can replace
```
model.device =  device
```

with
```python3
@property
def device(self) -> torch.device:
  return next(self.parameters()).device
```
where the above block should be placed in `class Transducer` in `model.py`.

yaozengwei:
> > This is a bit confusing, if we are setting a .device that is different from the device it is actually on.
> 
> We can replace
> 
> ```
> model.device =  device
> ```
> 
> with
> 
> ```python
> @property
> def device(self) -> torch.device:
>   return next(self.parameters()).device
> ```
> 
> where the above block should be placed in `class Transducer` in `model.py`.

Ok. I will do this.

yaozengwei:
I think it can be merged.

PR #351, opened 2022-05-06
First attempt to add WEB client to the streaming emformer.

csukuangfj:
It supports reading a wav file from the client and sends it to the server for recognition at present.

Note: Even though we send all the samples at once, the server supports accepting samples coming in a streaming fashion.

## Example screen recording:

https://user-images.githubusercontent.com/5284924/167114390-de28e2c7-4f97-4641-b5ff-bfe71f0554e8.mov

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -17,6 +17,7 @@ exclude =
   .git,
   **/data/**,
   icefall/shared/make_kn_lm.py,
+  egs/librispeech/ASR/transducer_emformer/train.py,
   icefall/__init__.py
 
 ignore =
--- a/egs/librispeech/ASR/transducer_emformer/client/index.html
+++ b/egs/librispeech/ASR/transducer_emformer/client/index.html
@@ -1,0 +1,36 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <!-- Required meta tags -->
+  <meta charset="utf-8">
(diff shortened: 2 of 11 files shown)


pkufool:
I also use websocket, it is in python. I think they are similar, but mine is very simple, I implement it in case we will need a demo for speechhome meeting. @csukuangfj  you can go ahead, I have not done much survey about the server framework.

csukuangfj:
The following is a demo for streaming recognition from real-time recordings.


https://user-images.githubusercontent.com/5284924/167253183-7115a96e-1095-4145-969b-69d72079e043.mov

csukuangfj:
> wow, nice!! Does this actually reflect when it is doing the recognition?

Do you mean the tail padding issue? I think this PR is using a version of beam search before we fix the padding issue.

danpovey:
I meant the video plays audio and shows recognition, I wonder if this is somehow "real" and reflects how fast you could recognize real live audio.

pzelasko:
It looks like it has sth around 200-300ms latency between the time I hear a word and it appears in the recognition, I'd that's easily representative of actual live performance.

csukuangfj:
> I meant the video plays audio and shows recognition, I wonder if this is somehow "real" and reflects how fast you could recognize real live audio.

Yes, I replayed two test waves from the test-clean dataset from my phone.

PR #344, opened 2022-05-05
Model average

yaozengwei:
This PR implements Dan's idea about model average. Detailed description could be found on the issue https://github.com/k2-fsa/icefall/issues/337. 
The comment history about the implementation could be found on https://github.com/yaozengwei/icefall/pull/1.

PR #353, opened 2022-05-07
Support decoding with averaged model when using --iter

yaozengwei:
This PR implements decoding with averaged model when using `--iter`. 
The following result is a continuation of https://github.com/k2-fsa/icefall/pull/344#issuecomment-1118592180.
When trained on full librispeech with 6 gpu for 30 epochs, average_period=100, I got following results with greedy search decoding:
- decode with checkpoint-176000, avg=4, 2.83, 6.83 (without averaged model) -> 2.72, 6.66 (with averaged model);
- decode with checkpoint-176000, avg=8, 2.75, 6.75 (without averaged model) -> 2.74, 6.67 (with averaged model).


egs/librispeech/ASR/pruned_transducer_stateless4/decode.py
```diff
@@ -540,23 +540,52 @@ def main():
...
+                )
+            elif len(filenames) < params.avg + 1:
+                raise ValueError(
+                    f"Not enough checkpoints ({len(filenames)}) found for"
+                    f" --iter {params.iter}, --avg {params.avg}"
+                )
+            filename_start = filenames[-1]
+            filename_end = filenames[0]
+            logging.info(
+                f"Calculating the averaged model over iteration checkpoints"
```
csukuangfj, line 560:
```suggestion
                "Calculating the averaged model over iteration checkpoints"
```

icefall/checkpoint.py
```diff
@@ -1,4 +1,4 @@
-# Copyright      2021  Xiaomi Corp.        (authors: Fangjun Kuang,
```
csukuangfj, line 1: 2021-2022

PR #354, opened 2022-05-07
Stringify torch.__version__ before serializing it.

csukuangfj:
Otherwise, it causes problems when loading a checkpoint saved by torch 1.10.0 from torch 1.7.1


danpovey:
LGTM!

PR #367, opened 2022-05-15
Emformer augmented with positional encoding and convolution module

yaozengwei:
The training and decoding codes are based on https://github.com/k2-fsa/icefall/tree/master/egs/librispeech/ASR/pruned_transducer_stateless4, with scaling mechanism.

Now it supports positional encoding for emformer attention and adds convolution module.

The implementation is based on the torchaudio 
- emformer: https://github.com/pytorch/audio/blob/main/torchaudio/models/emformer.py
- conv module: https://github.com/pytorch/audio/blob/main/torchaudio/prototype/models/conv_emformer.py


yaozengwei:
This PR is too big. I will create a PR merge the master branch into the streaming branch first, then create another PR for the emformer codes.

PR #373, opened 2022-05-19
Improve diagnostics code memory-wise and accumulate more stats.

danpovey:


danpovey:
If no-one sees any obvious problems I suggest to merge this and fix any problems that might arise, later.
I have tested the main part of this on a local branch.  
I realized that the diagnostics for the gradients (especially the eigenvalues) were not robust due to too little data,
and I had to rework how the diagnostics code works to accumulate more stats without blowing up memory.

PR #377, opened 2022-05-19
Add non-streaming ASR server implementation

csukuangfj:
This PR tries to implement the framework proposed in https://github.com/k2-fsa/k2/pull/975

The current implementation is all in Python.

The following shows a screen recording for the running results: 1 server + 4 clients.

https://user-images.githubusercontent.com/5284924/169350034-05f5cad8-d6f5-413f-ab52-0855141a26b2.mov

## Usage
```
# start the server
cd icefall/bin
./offline_server.py
```

```
# start the client(s)
./offline_client.py


# to start multiple clinets, open a new terminal, and run ./offline_client.py again
```

From the output of `py-spy top --pid <pid-of-the-server>`, we can see that a large part of the time is spent in neural network computation. I am going to write a c++ extension to do the nerual network computation, which can be run in a separate thread with the global interpretor lock being released.

The decoding part will still be in Python, we can move it to C++ if needed.

Shall we put the server stuff in a separate repo, e.g., `k2-fsa/server` or can I put it in `csukuangfj/server`?

### DIFF
--- a/bin/offline_client.py
+++ b/bin/offline_client.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python3
+# Copyright      2022  Xiaomi Corp.        (authors: Fangjun Kuang)
+#
+# See ../../../../LICENSE for clarification regarding multiple authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


pzelasko:
Very cool. If it's an "official" server that is going to be supported with the rest of the project, then I vote for `k2-fsa/server`

csukuangfj:
Closing it. 

See https://github.com/k2-fsa/sherpa, which is dedicated to the ASR server stuff.

PR #242, opened 2022-03-07
Streaming conformer pruned transducer

glynpu:
A streaming conformer training/decoding/stimulate_streaming decoding pipeline modified from pruned_transducer_stateless; 
The model trained with full librispeech by 50 epochs.

decoding with ctc greedy search:

right_context|1|8|16|32|64|full
--|--|--|--|--|--|--
latency|0.07s|0.35s|0.67s|1.31s|2.59s|*
test_clean|5.60|4.00|3.76|3.75|3.65|3.28|
+20 tailing dummy frames|5.52|3.98|3.75|3.75|3.65|3.28
simulate streaming with chunk_by_chunk decoding|5.52|3.98|3.75|3.75|3.65|3.28
test_other|14.07|10.69|9.80|9.48|9.01|8.05|
+20 tailing dummy frames|14.00|10.69|9.80|9.48|9.0|8.04
simulate streaming with chunk_by_chunk decoding|14.00|10.69|9.80|9.48|9.0|8.04

Other information could be found in the streaming_pruned_transducer_stateless/README.md in this pr.

Issue #337, opened 2022-04-28
Model averaging


danpovey:
OK, we have some results locally (thanks, @yaozengwei!)  showing that model averaging over finely spaced checkpoints is a bit better than averaging over epochs.  It's about 0.05% on test-clean and 0.15% on test-other, at around 3%/7% WER, but probably still worth doing as at this point we are picking up pennies in WER.

... OK, here's the idea.  The idea is that we always store a separate version of the model, say model_avg, in which for each floating-point parameter, it contains the average *from the start of training* of all the parameters.  We update this every average_period epochs, for, say, average_period = 10 or 100 (could be every epoch but this is for speed).  Each time we average, we do:
```
    model_avg = model_avg * ((cur_batch_idx_train-average_period) / cur_batch_idx_train) + model * (average_period/cur_batch_idx_train)
```
[this is not the syntax we'd use, we'd have to write a function to do this weighted-average.]
I propose that we include this averaged version inside the `checkpoints-*.pt` and  `epoch-*.pt` as a separate key in the dict.
Then the way we would implement decoding epoch-29.pt with --avg 5 would be something like the following.
```
   Let epoch_30 = epoch-29.pt ['averaged']    # Unfortunately we have to add 1.. epoch-29.pt was really 
                                                                            #  trained for 30 epochs!!  I wish there was a way to fix this without disruption.
         epoch_25 = epoch-24.pt ['averaged']
  epochs_25_to_30 = (epoch_30 * 30 - epoch_25 * 25) / 5
```

yaozengwei:
I will write the function.

pzelasko:
I think PyTorch has sth similar here https://pytorch.org/docs/stable/optim.html#stochastic-weight-averaging

PR #380, opened 2022-05-22
Using streaming conformer as transducer encoder

pkufool:
I slightly refactored https://github.com/k2-fsa/icefall/pull/242, and create this new pull request.  I think it is no need to duplicate a new folder, as making conformer to streaming just needs very minor fixes.

~~I also add the symbol delay penalty to this PR, see https://github.com/k2-fsa/k2/issues/955 for how we penalize and measure symbol delay.~~

Will post the experiment results here soon.


egs/librispeech/ASR/pruned_transducer_stateless2/conformer.py
```diff
@@ -117,17 +151,148 @@ def forward(
...
-            x, pos_emb, src_key_padding_mask=mask, warmup=warmup
+        src_key_padding_mask = make_pad_mask(lengths)
+        mask = None
+
+        if self.dynamic_chunk_training:
+            assert (
+                self.causal
+            ), "Causal convolution is required for streaming conformer."
+            max_len = x.size(0)
+            chunk_size = torch.randint(1, max_len, (1,)).item()
```
pkufool, line 163: The method to choose `chunk_size` here is from wenet's paper, we may tune it to find a better one in the future.

egs/librispeech/ASR/pruned_transducer_stateless2/conformer.py
```diff
@@ -117,17 +151,148 @@ def forward(
...
+    def streaming_forward(
+        self,
+        x: torch.Tensor,
+        x_lens: torch.Tensor,
+        warmup: float = 1.0,
+        states: Optional[Tensor] = None,
+        chunk_size: int = 16,
+        left_context: int = 64,
+        simulate_streaming: bool = False,
+    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
```
pkufool, line 195: The streaming forword interface is emformer-like, it handles decoding states (i.e. caches) outside, which will make it easer to implement async decoding.

danpovey:
Cool!
One thing we should perhaps discuss is whether this belongs as a change to pruned_transducer_stateless2, vs.  pruned_transducer_stateless4 which I believe is the latest recipe?  @csukuangfj what are the changes from 2 to 4?

yaozengwei:
> Cool! One thing we should perhaps discuss is whether this belongs as a change to pruned_transducer_stateless2, vs. pruned_transducer_stateless4 which I believe is the latest recipe? @csukuangfj what are the changes from 2 to 4?

`pruned_transducer_stateless4` (see https://github.com/k2-fsa/icefall/pull/344) supports saving averaged model and the epoch number counts from 1.

csukuangfj:
>  @csukuangfj what are the changes from 2 to 4?

"4" has the changes from zengwei that supports model averaging that you proposed in https://github.com/k2-fsa/icefall/issues/337

Otherwise, 4 is identical to 2. I would suggest to always use 4 as zengwei has showed it is helpful.

From my results of deeper and narrower conformer, it is also helpful.

danpovey:
@csukuangfj BTW, are we recommending the deeper and narrower conformer right now?  How much better was it, again?

danpovey:
@pkufool it may be fairly easy to use `git diff` to move your changes from 2 to 4.  E.g.
cd pruned_transducer_stateless2 
git diff master..HEAD . > ~/diff
cd ../pruned_transducer_stateless4
patch -p4 < ~/diff
(the 4 may need to be changed, e.g. to 3 or 5, if this fails).

csukuangfj:
> @csukuangfj BTW, are we recommending the deeper and narrower conformer right now? How much better was it, again?

I am updating the results and uploading the pre-trained models to huggingface. Please wait for a moment.

pkufool:
> Cool! One thing we should perhaps discuss is whether this belongs as a change to pruned_transducer_stateless2, vs. pruned_transducer_stateless4 which I believe is the latest recipe? @csukuangfj what are the changes from 2 to 4?

I think they share the same conformer.py, only a little changes on train.py & decode.py. 

> @pkufool it may be fairly easy to use git diff to move your changes from 2 to 4. E.g.
cd pruned_transducer_stateless2
git diff master..HEAD . > ~/diff
cd ../pruned_transducer_stateless4
patch -p4 < ~/diff
(the 4 may need to be changed, e.g. to 3 or 5, if this fails).

Thanks, I will try it.

pkufool:
## Summarize my experiment results for streaming conformer below:
## About the dynamic chunk training
I tried several settings(Note: The chunk sizes and contexts are all in frames after subsampling):
* The first row is the results reproduced with https://github.com/k2-fsa/icefall/pull/242 .
* The second row  was trained with limited left context, because we can only see limited left context for parallel streaming decoding (Note: the left context is in chunks, and the chunk size is dynamicly sampled from [1, 25]). 
* The third and forth rows were trained with fixed chunk size, I tried 16 and 8.
* The fifth row was trained with fixed left context, I only tried left context = 32.

Comparing with the first and second row,  we can see if we need to decode with limited left context, we have to trained with limited left context. Comparing with the second row and last three rows, we can see the models trained with fixed chunk size and fixed left context do not outperform the model trained with dynamic chunk size.
image

### About using different left-context for different layers
I also tried train the conformer encoder using different left context for different layers. (The models are all 12 layers conformer).
* The baseline system was trained with num-left-chunks=4 for all layers and decoded with chunk size =16, left-context=64.
* For the second row, the lower 8 layers used num-left-chunks=2, the upper 4 layers used num-left-chunks=4. In decoding time, they also used different left context(32 for lower 8 layers, 64 for upper 4 layers).
* The third row is the same as second row, the only differnece is I splited the layers into 6 + 6.

We can see that the WERs are close to each other for all three settings, smaller left context can save more memories when decoding, though. I did not do much experiments on this idea(i.e. how to split layers, how many left context needed by different layers), so the conclusion might not soild. 
image

### About streaming decoding and right-context
I also implemented the real paralell streaming decoding for the conformer model, you can see the code in `streaming_decode.py` in each directory. The table below shows the decoding results of several models using different chunk sizes and contexts.  You can see in the table, adding right context usually helps to decrease the WERs, but for some settings, like decoding with left-context=32 and chunk-size=8 in pruned_transducer_stateless2, pruned_transducer_stateless3, adding right context makes it worse. I still have not figured out the reasons.

image

**Note:** You can also try these streaming conformer models in our server framework Sherpa (https://github.com/k2-fsa/sherpa/tree/master/sherpa/bin/streaming_conformer_rnnt).

egs/librispeech/ASR/pruned_transducer_stateless/decode.py
```diff
@@ -250,6 +262,69 @@ def get_parser():
...
+        "--short-chunk-size",
+        type=int,
+        default=25,
+        help="""Chunk length of dynamic training, the chunk size would be either
+        max sequence length of current batch or uniformly sampled from (1, short_chunk_size).
+        Note: not needed for decoding, adding it here to construct transducer model,
+              as we reuse the code in train.py.
+        """,
+    )
+
```
csukuangfj, line 286: You can move these options to train.py and import them into here.
See
https://github.com/k2-fsa/icefall/blob/dc89b61b808a22b64e7e38cb09771ea7a6bb64d2/egs/librispeech/ASR/pruned_transducer_stateless5/train.py#L331

egs/librispeech/ASR/pruned_transducer_stateless/decode.py
```diff
@@ -301,9 +376,19 @@ def decode_one_batch(
...
     feature_lens = supervisions["num_frames"].to(device)
 
-    encoder_out, encoder_out_lens = model.encoder(
-        x=feature, x_lens=feature_lens
-    )
+    if params.simulate_streaming:
+        encoder_out, encoder_out_lens, _ = model.encoder.streaming_forward(
+            x=feature,
+            x_lens=feature_lens,
+            states=[],
```
csukuangfj, line 383: I would suggest supporting `None` and not using an empty list `[]` here.

egs/librispeech/ASR/pruned_transducer_stateless/decode.py
```diff
@@ -250,6 +262,69 @@ def get_parser():
...
+    parser.add_argument(
+        "--simulate-streaming",
+        type=str2bool,
+        default=False,
+        help="""Whether to simulate streaming in decoding, this is a good way to
+        test a streaming model.
+        """,
+    )
+
+    parser.add_argument(
```
csukuangfj, line 306: I recommend sharing the parameters from train.py whenever possible.

egs/librispeech/ASR/pruned_transducer_stateless/decode.py
```diff
@@ -301,9 +376,19 @@ def decode_one_batch(
...
-        x=feature, x_lens=feature_lens
-    )
+    if params.simulate_streaming:
+        encoder_out, encoder_out_lens, _ = model.encoder.streaming_forward(
+            x=feature,
+            x_lens=feature_lens,
+            states=[],
+            chunk_size=params.decode_chunk_size,
+            left_context=params.left_context,
+            simulate_streaming=True,
```
csukuangfj, line 386: When using `simulate_streaming=False,` does it have any difference with the `else` part?

egs/librispeech/ASR/pruned_transducer_stateless2/train.py
```diff
@@ -40,6 +40,18 @@
...
 
+# train a streaming model
+./pruned_transducer_stateless2/train.py \
+  --world-size 4 \
+  --num-epochs 30 \
+  --start-epoch 0 \
+  --exp-dir pruned_transducer_stateless/exp \
+  --full-libri 1 \
+  --dynamic-chunk-training 1 \
+  --causal_convolution 1 \
```
luomingshuang, line 51: causal_convolution-->causal-convolution

pkufool: Fixed, Thanks!

egs/librispeech/ASR/pruned_transducer_stateless/streaming_decode.py
```diff
@@ -0,0 +1,679 @@
...
+            finished_streams.append(i)
+
+    return finished_streams
+
+
+def decode_dataset(
+    cuts: CutSet,
+    params: AttributeDict,
+    model: nn.Module,
+    sp: spm.SentencePieceProcessor,
```
luomingshuang, line 403: It seems that this parameter `sp` is not used in this function.

egs/librispeech/ASR/pruned_transducer_stateless/streaming_decode.py
```diff
@@ -0,0 +1,679 @@
...
+    lattice = decoding_streams.format_output(processed_lens.tolist())
+    best_path = one_best_decoding(lattice)
+    hyp_tokens = get_texts(best_path)
+    return hyp_tokens
+
+
+def decode_one_chunk(
+    params: AttributeDict,
+    model: nn.Module,
+    sp: spm.SentencePieceProcessor,
```
luomingshuang, line 288: It seems that the parameter `sp` is not used in this function.

pkufool:
Merging now. Most of the code added is to support streaming decoding, it won't break any previous recipes. Will fix the issues later in other PRs if there is any.

PR #384, opened 2022-05-24
Add tail padding to the Emformer model.

csukuangfj:
Will try to port the Emformer model to the master branch in a separate PR.


egs/librispeech/ASR/transducer_emformer/decode.py
```diff
@@ -247,13 +250,20 @@ def decode_one_batch(
     supervisions = batch["supervisions"]
     feature_lens = supervisions["num_frames"].to(device)
 
+    feature_lens += params.left_context_length
+    feature = torch.nn.functional.pad(
+        feature,
+        pad=(0, 0, 0, params.left_context_length),
+        value=LOG_EPS,
+    )
```
csukuangfj, line 258: These lines add tail paddings.

PR #386, opened 2022-05-25
Ask for help: Emformer augmented with positional encoding and convolution module

yaozengwei:
This PR is to ask for help. 

I try to support `relative positional encoding` for emformer attention and also add `convolution module`. I hope someone can help me review the implementation of the `relative positional encoding`. 

Now the model training is too slow, which needs about 3 hours per epoch on 6 GPU. The training code is based on `pruned_transducer_stateless4`, with scaling mechanism and averaged model saving.

The attention module in Emformer model (https://arxiv.org/pdf/2010.10759.pdf) contains three parts: 
(a) [current chunk] (in query) -> [memory vectors of past chunks, left context, current chunk, right context] (in key);
(b) [right context] (in query) -> [memory vectors of past chunks, left context, current chunk, right context] (in key);
(c) [summary value of current chunk] (in query) -> [left context, current chunk, right context] (in key), this is to calculate the memory vector for current chunk. 

Currently I add the position encoding on part (a). See `EmformerAttention` class in `conv_emformer_transducer_stateless/emformer.py` for more details. 

The emformer implementation is based on the torchaudio codes:
- emformer attention: https://github.com/pytorch/audio/blob/main/torchaudio/models/emformer.py (without positional encoding yet)
- convolution module: https://github.com/pytorch/audio/blob/main/torchaudio/prototype/models/conv_emformer.py

### DIFF
--- a/egs/librispeech/ASR/conv_emformer_transducer_stateless/asr_datamodule.py
+++ b/egs/librispeech/ASR/conv_emformer_transducer_stateless/asr_datamodule.py
@@ -1,0 +1,1 @@
+../pruned_transducer_stateless2/asr_datamodule.py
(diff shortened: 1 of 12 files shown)


csukuangfj:
Ok, I will.

PR #392, opened 2022-06-02
Update decode.py

fanlu:
fix bug ```TypeError: greedy_search_batch() missing 1 required positional argument: 'encoder_out_lens'```

### DIFF
--- a/egs/wenetspeech/ASR/pruned_transducer_stateless2/decode.py
+++ b/egs/wenetspeech/ASR/pruned_transducer_stateless2/decode.py
@@ -274,6 +274,7 @@ def decode_one_batch(
         hyp_tokens = greedy_search_batch(
             model=model,
             encoder_out=encoder_out,
+            encoder_out_lens=encoder_out_lens,
         )
         for i in range(encoder_out.size(0)):
             hyps.append([lexicon.token_table[idx] for idx in hyp_tokens[i]])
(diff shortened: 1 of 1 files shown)


egs/wenetspeech/ASR/pruned_transducer_stateless2/decode.py
```diff
@@ -274,6 +274,7 @@ def decode_one_batch(
         hyp_tokens = greedy_search_batch(
             model=model,
             encoder_out=encoder_out,
+            encoder_out_lens=encoder_out_lens,
```
csukuangfj, line 277: Can you also fix fast_beam_search? I think its name has been changed to fast_beam_serach_one_best.

Also, for the following `modified_beam_search`, it also needs `encoder_out_lens`.

luomingshuang:
It seems that merging the wenetspeech recipe was before changing the beam_search.py.

csukuangfj:
Had we had CI tests for each model of each dataset, we could detect such problems easily

luomingshuang:
I&nbsp;am&nbsp;planning&nbsp;to&nbsp;add&nbsp;these.


---Original---
From: ***@***.***&gt;
Date: Thu, Jun 2, 2022 22:07 PM
To: ***@***.***&gt;;
Cc: &quot;Mingshuang ***@***.******@***.***&gt;;
Subject: Re: [k2-fsa/icefall] Update decode.py (PR #392)




  
Had we had CI tests for each model of each dataset, we could detect such problems easily
  
maybe give new param a default value None, and assert param is not None.
 
&mdash;

PR #400, opened 2022-06-06
[Ready to merge] Do a change for wenetspeech/pruned_transducer_stateless2/decode.py

luomingshuang:
It is to do a change for `wenetspeech/pruned_transducer_stateless2/decode.py` .


csukuangfj:
Thanks!

PR #401, opened 2022-06-06
Fix Emformer for torchscript using torch 1.6.0

csukuangfj:
torchscript in torch 1.6.0 and 1.7.1 complains that it cannot convert list-of-list to bool.

This PR fixes that.

### DIFF
--- a/egs/librispeech/ASR/pruned_stateless_emformer_rnnt2/emformer.py
+++ b/egs/librispeech/ASR/pruned_stateless_emformer_rnnt2/emformer.py
@@ -296,7 +296,7 @@ class Emformer(EncoderInterface):
           Return the initial state of each layer. NOTE: the returned
           tensors are on the given device. `len(ans) == num_emformer_layers`.
         """
-        if self._init_state:
+        if len(self._init_state) > 0:
             # Note(fangjun): It is OK to share the init state as it is
             # not going to be modified by the model
             return self._init_state


egs/librispeech/ASR/pruned_stateless_emformer_rnnt2/emformer.py
```diff
@@ -296,7 +296,7 @@ def get_init_state(self, device: torch.device) -> List[List[torch.Tensor]]:
           Return the initial state of each layer. NOTE: the returned
           tensors are on the given device. `len(ans) == num_emformer_layers`.
         """
-        if self._init_state:
+        if len(self._init_state) > 0:
```
pkufool, line 299: I found that `len(self._init_state)` is always 2 and `if self._init_state` always returns True.

image

Should use `len(self._init_state.value)`.

My torch version is `1.10.0+cu102`.

csukuangfj: Thanks.

> I found that len(self._init_state) is always 2 and if self._init_state always returns True.

Yes, you are right. The code should be changed.


From https://github.com/pytorch/pytorch/blob/master/torch/jit/_script.py#L67

```python3
if _enabled:
    Attribute = collections.namedtuple("Attribute", ["value", "type"])
else:

    def Attribute(value, type):  # type: ignore[no-redef]
        return value
```

We should check whether torch script is enabled or not before changing it.

csukuangfj: I just verified that the following works:
```diff
--- a/egs/librispeech/ASR/pruned_stateless_emformer_rnnt2/emformer.py
+++ b/egs/librispeech/ASR/pruned_stateless_emformer_rnnt2/emformer.py
@@ -202,6 +202,7 @@ class Emformer(EncoderInterface):
         )
         self.log_eps = math.log(1e-10)

+        self.has_init_state = False
         self._init_state = torch.jit.Attribute([], List[List[torch.Tensor]])

     def forward(
@@ -296,7 +297,7 @@ class Emformer(EncoderInterface):
           Return the initial state of each layer. NOTE: the returned
           tensors are on the given device. `len(ans) == num_emformer_layers`.
         """
-        if self._init_state:
+        if self.has_init_state:
             # Note(fangjun): It is OK to share the init state as it is
             # not going to be modified by the model
             return self._init_state
@@ -308,6 +309,7 @@ class Emformer(EncoderInterface):
             s = layer._init_state(batch_size=batch_size, device=device)
             ans.append(s)

+        self.has_init_state = True
         self._init_state = ans

         return ans
```

PR #404, opened 2022-06-07
Replace LilcomChunkyWriter with ChunkedLilcomHdf5Writer.

csukuangfj:
It seems that there are issues with LilcomChunkyWriter in the multi-threaded scenario.

See
- https://github.com/k2-fsa/icefall/discussions/391#discussioncomment-2870627
- https://github.com/k2-fsa/icefall/discussions/391#discussioncomment-2885434
- https://github.com/lhotse-speech/lhotse/issues/607#issue-1157450209
- https://github.com/lhotse-speech/lhotse/issues/607#issuecomment-1148373507

One advantage of using ChunkedLilcomHdf5Writer is that it does not need to compete for locks in
the case of multiple threads.

### DIFF
--- a/egs/aishell/ASR/local/compute_fbank_aidatatang_200zh.py
+++ b/egs/aishell/ASR/local/compute_fbank_aidatatang_200zh.py
@@ -29,7 +29,7 @@ import os
 from pathlib import Path
 
 import torch
-from lhotse import CutSet, Fbank, FbankConfig, LilcomChunkyWriter
+from lhotse import ChunkedLilcomHdf5Writer, CutSet, Fbank, FbankConfig
 from lhotse.recipes.utils import read_manifests_if_cached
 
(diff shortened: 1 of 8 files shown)


pzelasko:
Unfortunately, I wasn’t able to reproduce the issues you had in the other discussion. If you can create a minimal case that helps with repro I could try to fix it in LilcomChunky. HDF5 unfortunately will cause issues to ‚big data’ users…

danpovey:
Definitely having 2 worker threads or more per data-loader process is important.
Unfortunately thread-safety bugs can be hard to reproduce.

csukuangfj:
I am trying to extract a subset of the data, upload it to huggingface and try to reproduce in colab notebook.

csukuangfj:
It is 100% reproducible locally.

Issue #405, opened 2022-06-08
How to finetuning model to recognize english abbreviations? eg: GDP，CPI etc


shushanxingzhe:
@luomingshuang Very thanks for your great work in wenetspeech recipe. The model have archived a good CER, but it is hard to recognize english abbreviations, eg: GDP，CPI etc. Is there any advice to improving it?

luomingshuang:
Thanks for your observation. Yes, in my public recognized results, the English abbreviations still can't be recognized correctly. It's a good question. How to improve the performance of recognizing English abbreviations in Chinese texts. I think there are some reasons leading this case. 
For example, the English abbreviations is `GDPR`, and my recognized results is `极低P而`. Some chinese chars which have the same tone as the English abbreviations mislead the model. 
How to solve this problem is a good topic. Maybe we can leverage the language model or context to relieve it. And maybe others can have more suggestions for this problem.

luomingshuang:
I think one place which can be improved is to do some changes when generate the tokens.txt. In my current, I split the 
proper English abbreviations into individual letters. (GDP-->G D P, generating three tokens: G, D, P.) So one way that we can do is to regard the GDP as a single token, not split it. Maybe I will have some experiments based on this way.

shushanxingzhe:
Very thanks for your reply. I look forward to your release of the new model as I don't have enough hardware to experiment.

PR #413, opened 2022-06-09
[WIP] Add English abbreviations into tokens and finetune the model based on pretrained model for WenetSpeech

luomingshuang:
This PR can be a response for this issue #405 . I re-generate the tokens.txt which regards the English abbreviation as a single token. (the number of tokens changes: 5537->5679.) At the same time, I do some changes for the process of encoding the text into ids. For training the L subset, I set a choice for loading a pretrained model which is trained with 5537 tokens. (https://huggingface.co/luomingshuang/icefall_asr_wenetspeech_pruned_transducer_stateless2/blob/main/exp/pretrained_epoch_10_avg_2.pt). This is to save training time. Of course, if load the pretrained model as the initial state, I think the initial lr should be decreased. (3.0e-3->1e-3)


luomingshuang:
The results (WER(%)) for pruned_transducer_stateless3 fine-tuned with wenetspeech (I will update the results on real time.):
|decoding-method | epoch | avg | dev | test-net | test-meeting |
|--|--|--|--|--|--|
|greedy_search | 0 | 1 | 10.09 | 10.41 | 19.38 |

luomingshuang:
It seems that this method can't get a good performance. Close it.

PR #390, opened 2022-05-31
Add streaming Emformer stateless RNN-T.

csukuangfj:
This PR uses the Emformer model from torchaudio, which requires torchaudio >= 0.11.0.

## Training command
```bash
./pruned_stateless_emformer_rnnt2/train.py \
  --world-size 8 \
  --num-epochs 40 \
  --start-epoch 1 \
  --exp-dir pruned_stateless_emformer_rnnt2/exp-full \
  --full-libri 1 \
  --use-fp16 0 \
  --max-duration 200 \
  --prune-range 5 \
  --lm-scale 0.25 \
  --master-port 12358 \
  --num-encoder-layers 18 \
  --left-context-length 128 \
  --segment-length 8 \
  --right-context-length 4
```

## Decoding command
```bash
for m in greedy_search fast_beam_search modified_beam_search; do
  for epoch in 39; do
    for avg in 6; do
      ./pruned_stateless_emformer_rnnt2/decode.py \
        --epoch $epoch \
        --avg $avg \
        --use-averaged-model 1 \
        --exp-dir pruned_stateless_emformer_rnnt2/exp-full \
        --max-duration 50 \
        --decoding-method $m \
        --num-encoder-layers 18 \
        --left-context-length 128 \
        --segment-length 8 \
        --right-context-length 4
    done
  done
done
```

| method | test-clean | test-other| comment|
|---|---|---|--|
|greedy search | 4.28 | 11.42 | epoch 39, avg 6 |
|modified beam search| 4.22 | 11.16| epoch 39, avg 6|
|fast beam search |  4.29 | 11.26 | epoch 39, avg 6|

The baseline is from
https://github.com/pytorch/audio/blob/main/examples/asr/emformer_rnnt/README.md
- test-clean: 4.56
- test-other: 10.66

Note that the baseline is trained for 120 epochs, with 32 GPUs

Also, the baseline uses vocab size 4098.

Will switch to #389

The pretrained model can be used in https://github.com/k2-fsa/sherpa/pull/6 for streaming ASR recognition.

I am uploading the training logs, decoding results, decoding logs, and pretrained model to hugging face.

PR #427, opened 2022-06-16
Disable drop_last in testing time

pkufool:


csukuangfj:
thanks

csukuangfj:
#390 added that argument by mistake.

PR #429, opened 2022-06-16
Upload extracted codebook indexes

yaozengwei:
This PR is about pruned_transducer_stateless6. It aims to allow people directly download the extracted codebook indexes for model distillation.


yaozengwei:
The detailed steps are in `distillation_with_hubert.sh`.
To directly download the extracted codebook indexes for model distillation, you can set stage=2, stop_stage=4, use_extracted_codebook=True.
To start from scratch, you can set stage=0, stop_stage=4, use_extracted_codebook=False.

pzelasko:
I see you fixed the same thing as I in #430; can you also update the HUBERT download URL? Then we could close my PR to avoid conflicts.

yaozengwei:
> I see you fixed the same thing as I in #430; can you also update the HUBERT download URL? Then we could close my PR to avoid conflicts.

Sure. I have also updated the HUBERT download URL. See https://github.com/k2-fsa/icefall/blob/10ea04ae807b0f2eeb5d5039269e6e2f91a137cc/egs/librispeech/ASR/distillation_with_hubert.sh#L99

pzelasko:
Thanks!

egs/librispeech/ASR/distillation_with_hubert.sh
```diff
@@ -33,10 +35,35 @@ stage=$1
...
+
+# use_extracted_codebook can be "True" or "False"
+#   "True" -> stage 0 and stage 1 would be skipped,
+#     and directly download the extracted codebook indexes for distillation
+#   "False" -> start from scratch
+use_extracted_codebook=True
+
+# teacher_model_id can be one of
+#   "hubert_xtralarge_ll60k_finetune_ls960" -> fine-tuned model, it is the one we currently use.
+#   "hubert_xtralarge_ll60k.pt" -> pretrained model without fintuing
```
glynpu, line 56: Maybe should remove the ".pt" suffix of "hubert_xtralarge_ll60k.pt"?

PR #430, opened 2022-06-16
Update file names in Hubert VQ recipe

pzelasko:
I'm also updating the Hubert download URL, I was getting 403 error if ".pt" is not there.


pzelasko:
Duplicated work with #429

danpovey:
Thanks anyway!

PR #433, opened 2022-06-17
Support torch 1.6.0

csukuangfj:


egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
```diff
@@ -52,7 +52,15 @@ def forward(
...
                 channel_dim += x.ndim
-            sum_dims = [d for d in range(x.ndim) if d != channel_dim]
+
+            #  sum_dims = [d for d in range(x.ndim) if d != channel_dim]
+            # The above line is not torch scriptable for torch 1.6.0
+            # torch.jit.frontend.NotSupportedError: comprehension ifs not supported yet:  # noqa
+            sum_dims = 0
+            for d in range(x.ndim):
+                if d != channel_dim:
+                    sum_dims += d
```
pkufool, line 62: I think this is not equivalent to the original one, is it right?

Should be like this:
```
sum_dims = []
# If torch version less than 1.7.0, `if` in List will cause
# torch.jit.frontend.NotSupportedError: comprehension ifs not
# supported yet
if torch.jit.is_scripting() and torch.__version__ < '1.7.0':
    for d in range(x.ndim):
        if d != channel_dim:
            sum_dims.append(d)
else:
    sum_dims = [d for d in range(x.ndim) if d != channel_dim]
```

Correct me if I am wrong.

csukuangfj: thanks

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
```diff
@@ -234,6 +242,9 @@ def __init__(
     ):
         super(ScaledConv1d, self).__init__(*args, **kwargs)
         initial_scale = torch.tensor(initial_scale).log()
+
+        self.bias_scale: Optional[nn.Parameter]  # for torchscript
```
pkufool, line 246: I did not include this line, it did not raise any error.

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
pkufool: I verified  that this was fixed in torch 1.7.0. But it is ok to remove the if statement, just make it
``` 
sum_dims = []
for d in range(x.ndim):
        if d != channel_dim:
            sum_dims.append(d)
```

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
csukuangfj: It raised an error last night. I just tried again by removing it and it's strange there are no errors. You can remove it.

PR #434, opened 2022-06-18
Fix an error introduced by supporting torchscript for torch 1.6.0

csukuangfj:


danpovey:
How long was this bug present?  Because it could have degraded results (assuming it did not cause a crash)

csukuangfj:
Less than half a day.

csukuangfj:
It was introduced in https://github.com/k2-fsa/icefall/pull/433 by mistake.

danpovey:
OK no problem.  How did you discover it, did it crash?

csukuangfj:
I was trying to support torch script for torch 1.6.0. Actually, the buggy code is not used by torch script at all, but torch 1.6.0 tries to make it torch scritable anyway.

It was found by @pkufool in https://github.com/k2-fsa/icefall/pull/433#discussion_r900626174

csukuangfj:
I would suggest adding tests to cover every module that we add to icefall in the future so that we can refactor it without any worries about breaking it.

PR #435, opened 2022-06-18
Fix warmup

yaozengwei:
Set warmup=0.0 if params.start_epoch == 1 else 1.0 for during scan_pessimistic_batches_for_oom.


egs/librispeech/ASR/pruned_transducer_stateless2/train.py
```diff
@@ -883,6 +883,7 @@ def remove_short_and_long_utt(c: Cut):
             optimizer=optimizer,
             sp=sp,
             params=params,
+            warmup=0.0 if params.start_epoch == 0 else 1.0,
```
luomingshuang, line 886: So what's the advantage of using warmup=0.0 if params.start_epoch=0? It can help the model run quickly in scan_pessimistic_batches_for_oom?

csukuangfj: It is for when you resume training you don't want to re-do the warmup.

yaozengwei: > It is for when you resume training you don't want to re-do the warmup.

Yes.

luomingshuang: I notice the fault of start-epoch is `1`. So maybe you should change `warmup=0.0 if params.start_epoch == 0 else 1.0,` to `warmup=0.0 if params.start_epoch == 1 else 1.0,` .

yaozengwei: > I notice the fault of start-epoch is `1`. So maybe you should change `warmup=0.0 if params.start_epoch == 0 else 1.0,` to `warmup=0.0 if params.start_epoch == 1 else 1.0,` .

For pruned 2 and 3, epoch number counts from 0. For pruned 4, 5, 6, epoch number counts from 1.

luomingshuang: OK.

PR #440, opened 2022-06-20
Simplified memory bank for Emformer

yaozengwei:
This PR implements simplified memory bank (vectors) for Emformer encoder, in recipe `conv_emformer_transducer_stateless2`. 
The idea is from https://arxiv.org/pdf/2110.05241.pdf.

In the existing emformer implememtation (https://github.com/k2-fsa/icefall/tree/master/egs/librispeech/ASR/conv_emformer_transducer_stateless),  the memory vector of each chunk is obtained from attention computation:
```
average value of current chunk (as query) -> left context, current chunk, right context (as key, value)
```
The memory bank is from the preceding layer,  to get rid of the auto-regression property for block processing in training. Details can be found on https://arxiv.org/pdf/2010.10759.pdf. 

Now, the memory vector is simplified as the average value of current chunk. It also solves the problem of representation mismatch between successive layers, as discussed in https://arxiv.org/pdf/2110.05241.pdf.

When trained on train-clean-100, decode with averaged model epoch-30-avg-10, I get the following WERs on test-clean and test-other subsets:
1. Attention output as memory (in existing method)
- greedy search, 9.16 & 24.67
- fast beam search, 9.01 & 24.02
- modified beam search, 8.88 & 24.01

2. Averaged value as memory (in modified method)
- greedy search, 8.79 & 23.93
- fast beam search, 8.58 & 23.31
- modified beam search, 8.46 & 23.1


.flake8
```diff
@@ -10,6 +10,7 @@ per-file-ignores =
     egs/*/ASR/*/optim.py: E501,
     egs/*/ASR/*/scaling.py: E501,
     egs/librispeech/ASR/conv_emformer_transducer_stateless/*.py: E501, E203
+    egs/librispeech/ASR/conv_emformer_transducer_stateless2/*.py: E501, E203
```
luomingshuang, line 13: Maybe you can change it to `egs/*/ASR/conv_emformer_transducer_stateless2/*.py`.

luomingshuang: Actually, I don't suggest you use this `egs/librispeech/ASR/conv_emformer_transducer_stateless2/*.py: E501, E203` for the all py files in conv_emformer_transducer_stateless2. The better method is focus on some common files or some common code lines, which will not be modified in other datasets' conv_emformer_transducer_stateless2.

yaozengwei: > Actually, I don't suggest you use this `egs/*/ASR/conv_emformer_transducer_stateless2/*.py: E501, E203` for the all py files in conv_emformer_transducer_stateless2. The better method is focus on some common files or some common code lines, which will not be modified in other datasets' conv_emformer_transducer_stateless2.

Thanks.

yaozengwei:
For the higher latency setup: chunk-length=64 frames (0.64s), right-context=16 frames (0.16s), left-context=64 frames, memory size=32, I got the following results with the model at epoch-30-avg-10:
* greedy-search: 3.3 & 8.71 
* modified-beam-search: 3.26 & 8.56
* fast-beam-search: 3.27 & 8.58

yaozengwei:
> For the higher latency setup: chunk-length=64 frames (0.64s), right-context=16 frames (0.16s), left-context=64 frames, memory size=32, I got the following results with the model at epoch-30-avg-10:
> 
> * greedy-search: 3.3 & 8.71
> * modified-beam-search: 3.26 & 8.56
> * fast-beam-search: 3.27 & 8.58

For the lower latency setup: chunk-length=32 frames (0.32s), right-context=8 frames (0.08s), left-context=32 frames, memory size=32, I got the following results with the model at epoch-30-avg-10:
* greedy-search: 3.5 & 9.09
* modified-beam-search: 3.5 & 8.91
* fast-beam-search: 3.43 & 8.86

egs/librispeech/ASR/conv_emformer_transducer_stateless/emformer.py
```diff
@@ -1252,6 +1252,10 @@ def __init__(
     ):
         super().__init__()
 
+        assert int(math.log(chunk_length, 2)) == math.log(
```
csukuangfj, line 1255:
```python3
assert (chunk_length - 1) & chunk_length == 0
```

egs/librispeech/ASR/conv_emformer_transducer_stateless/emformer.py
```diff
@@ -1580,10 +1584,8 @@ def infer(
         chunk_mask = make_pad_mask(output_lengths).to(x.device)
         memory_mask = (
             (
-                torch.div(
-                    num_processed_frames,
-                    self.chunk_length,
-                    rounding_mode="floor",
+                (
+                    num_processed_frames << int(math.log(self.chunk_length, 2))
```
csukuangfj: Please cache the result of  `int(math.log(self.chunk_length, 2))`
in the constructor.

egs/librispeech/ASR/conv_emformer_transducer_stateless/emformer.py
```diff
@@ -1252,8 +1252,8 @@ def __init__(
     ):
         super().__init__()
 
-        assert int(math.log(chunk_length, 2)) == math.log(
-            chunk_length, 2
+        assert (
+            chunk_length - 1 & chunk_length == 0
```
csukuangfj, line 1256:
```suggestion
            (chunk_length - 1) & chunk_length == 0
```
is safer.

yaozengwei: Thanks

PR #456, opened 2022-06-30
Fix TIMIT lexicon generation bug

wangtiance:
The path to the manifest file was wrong. This would cause prepare.sh to err and stop at stage 5.


csukuangfj:
Thanks!

PR #462, opened 2022-07-04
CTC attention model with reworked Conformer encoder and reworked Transformer decoder

WayneWiser:
This PR implements a 'reworked' version of CTC attention model. As demenstrated in other PRs, reworked Conformer model has superior performance compared to the original Conformer. So in this modified version of CTC attention model, it has the reworked Conformer as the encoder with CTC loss and the reworked Transformer as the decoder with label smooth loss to improve the ASR performance. The 'reworked' idea is rather broad, please refer to github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/pruned_transducer_stateless2/ as well. It also integrates with the idea of the 'averaging models' in pruned_transducer_stateless4.

On the internal training and testing sets, it performs very well, the ASR accuracy is even slightly better than the pruned_transducer_stateless4 model, and much better than the original Conformer CTC attention model. The results on the public data sets, like Librispeech, might be uploaded soon.


csukuangfj:
Wow, thanks!

csukuangfj:
Please don't use force push.

egs/librispeech/ASR/conformer_ctc2/attention.py
```diff
@@ -0,0 +1,238 @@
...
+        \text{MultiHead}(Q, K, V) = \text{Concat}(head_1,\dots,head_h)W^O
+
+    where :math:`head_i = \text{Attention}(QW_i^Q, KW_i^K, VW_i^V)`.
+
+    Args:
+        embed_dim: Total dimension of the model.
+        num_heads: Number of parallel attention heads. Note that ``embed_dim`` will be split
+            across ``num_heads`` (i.e. each head will have dimension ``embed_dim // num_heads``).
+        dropout: Dropout probability on ``attn_output_weights``. Default: ``0.0`` (no dropout).
+        bias: If specified, adds bias to input / output projection layers. Default: ``True``.
```
csukuangfj, line 42:
```suggestion
```

There is no `bias` in the following constructor.

egs/librispeech/ASR/conformer_ctc2/attention.py
```diff
@@ -0,0 +1,238 @@
...
+        factory_kwargs = {"device": device, "dtype": dtype}
+        super(MultiheadAttention, self).__init__()
+        self.embed_dim = embed_dim
+        self.kdim = kdim if kdim is not None else embed_dim
+        self.vdim = vdim if vdim is not None else embed_dim
+        self._qkv_same_embed_dim = (
+            self.kdim == embed_dim and self.vdim == embed_dim
+        )
+        self.bias = True
+        bias = self.bias
```
csukuangfj, line 82: Can you pass `bias` as a parameter?

egs/librispeech/ASR/conformer_ctc2/decode.py
```diff
@@ -0,0 +1,875 @@
...
+            - (7) nbest-oracle. Its WER is the lower bound of any n-best
+              rescoring method can achieve. Useful for debugging n-best
+              rescoring method.
+        """,
+    )
+
+    parser.add_argument(
+        "--use-averaged-model",
+        type=str2bool,
+        default=False,
```
csukuangfj, line 128: Please change it to `True` by default.

csukuangfj:
Thanks! Please update `RESULTS.md`  to include
- WERs
- training and decoding commands
- link to tensorboard logs
- link to pretrained models.

egs/librispeech/ASR/conformer_ctc2/attention.py
WayneWiser: OK, I could set bias configurable but for bias=False, the code has not been tested.

WayneWiser:
For this PR, the experimental results are listed as follows. 

The baseline model is the original conformer CTC attention model trained with icefall/egs/librispeech/ASR/conformer_ctc. The model is downloaded from  [https://huggingface.co/csukuangfj/icefall-asr-librispeech-conformer-ctc-jit-bpe-500-2021-11-09](url). This model has 12 layers of Conformer encoder layers and 6 Transformer decoder layers. Number of model parameters is 109,226,120. It has been trained with 90 epochs with full Librispeech dataset.

For this reworked CTC attention model, it has 12 layers of reworked Conformer encoder layers and 6 reworked Transformer decoder layers.  Number of model parameters is 103,071,035. With full Librispeech data set, it was trained for **only** 30 epochs because the reworked model would converge much faster.

The training commands are:

```bash
WORLD_SIZE=8
      export CUDA_VISIBLE_DEVICES="0,1,2,3,4,5,6,7"
      ./conformer_ctc2/train.py \
      --manifest-dir data/fbank \
      --exp-dir conformer_ctc2/exp \
      --full-libri 1 \
      --spec-aug-time-warp-factor 80 \
      --max-duration 300 \
      --world-size ${WORLD_SIZE} \
      --start-epoch 1 \
      --num-epochs 30 \
      --att-rate 0.7 \
      --num-decoder-layers 6
```

Here are the WER comparisons. The decoding configuration for the reworked model is --epoch 30, --avg 8, --use-averaged-model True, which is the best after searching. 

| WER | reworked ctc attention|      |   | ctc attention|              |      | 
|-----|----------------------------|-----|----|---------|-----------|-----|
| test sets                    | test-clean|   test-other |  Avg | test-clean|   test-other| Avg| 
| ctc-greedy-search       | 2.98%	|     7.14%|	5.06%| 2.90%|	7.47%| 5.19%|
|ctc-decoding               | 2.98%	|     7.14%|	5.06%| 2.90%|  7.47%| 5.19%|
|1best                            | 2.93%	|     6.37%|	4.65%| 2.70%|	6.49%| 4.60%|
|nbest                            | 2.94%	|     6.39%|	4.67%| 2.70%|	6.48%| 4.59%|
|nbest-rescoring            | 2.68%     |      5.77%|        4.23%| 2.55%|	6.07%| 4.31%|
|whole-lattice-rescoring| 2.66%     |      5.76%|	4.21%| 2.56%|	6.04%| 4.30%|
|attention-decoder        | 2.59%     |      5.54%|	4.07%| 2.41%|  5.77%| 4.09%|
|nbest-oracle                  | 1.53%    |	      3.47%|	2.50%| 1.69%|	4.02%| 2.86%|




This PR also implemented the CTC greedy search decoding, it has the indentical WERs with the CTC-decoding method. For other decoding method, the average WER of the two test sets with the two model is similar. Except for the 1best and nbest method, the overall performance of reworked model is better than the baseline model.





As this PR used model averaging #337 during training and decoding, one can also compare the above results with icefall/egs/librispeech/ASR/pruned_transducer_stateless4 in the [https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/RESULTS.md](url).

The model will be uploaded to huggingface soon.

csukuangfj:
Thanks!  So test-other benefits more from the reworked conformer model. Also, it converges much faster and requires fewer epochs.

Merging.

PR #472, opened 2022-07-13
psd algorithm with ngram rescoring

glynpu:
@danpovey This pr is about "Algorithm 1 PSD algorithm" from [this paper](https://arxiv.org/pdf/2101.06856.pdf) discussed in yesterday's meeting.

**Questions**: 
How to compile a better decoding_graph? Or are there bugs of compiling HLG and Trivial_LG?
    Since Exp C is much better than Exp B by only changing the "H" part from k2.ctc_topo to k2.trivial_graph.
    
 

It's based on "greedy_search_batch".  
Results with different decoding_graph are:

exp-id|decoding_graph|test-clean|test-other
--|--|--|--
A|NO decoding_graph; original greedy_search_batch |2.04|4.72
B|HLG|3.43|5.76
C|Trivial_LG|2.42|5.03
D|trivial graph|2.04|4.72

Results analysis:
Exp A vs. Exp B:  Originally the HLG is compiled for conformer_ctc setup. Maybe this's the reason that it gets the highest WERs.
Exp C vs. (Exp B, A): k2.trivial_graph(used to compile Trivial_LG) seems more compatible with RNN-T setup than k2.ctc_topo(used to compile HLG). Though its WERs are still higher than baseline Exp A.
Exp D vs. (Exp C, B, A): its results are identical to that of Exp A. This indicates that "non-blank frames" selected code works as expected. But "better" paths are pruned by "HLG" or "Trivial_LG" somehow.

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless2/beam_search.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless2/beam_search.py
@@ -14,16 +14,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 import warnings
 from dataclasses import dataclass
 from typing import Dict, List, Optional
 
 import k2
 import torch
+from torch.nn.utils.rnn import PackedSequence, pad_packed_sequence
 from model import Transducer
 
-from icefall.decode import Nbest, one_best_decoding
-from icefall.utils import get_texts
+from icefall.decode import get_lattice, Nbest, one_best_decoding
+from icefall.utils import get_alignments, get_texts
 
 
 def fast_beam_search_one_best(
@@ -534,6 +536,8 @@ def greedy_search_batch(
     model: Transducer,
     encoder_out: torch.Tensor,
     encoder_out_lens: torch.Tensor,
+    decoding_graph: Optional[k2.Fsa] = None,
+    ngram_rescoring: bool = False,
 ) -> List[List[int]]:
     """Greedy search in batch mode. It hardcodes --max-sym-per-frame=1.
     Args:
@@ -583,6 +587,18 @@ def greedy_search_batch(
 
     encoder_out = model.joiner.encoder_proj(packed_encoder_out.data)
 
+    if ngram_rescoring:
+        vocab_size = model.decoder.vocab_size
+        total_t = encoder_out.shape[0]
+        # cached all joiner outputs during greedy search,
+        # from which non-blank frames are selected before n-gram rescoring.
+        all_logits = torch.zeros([total_t, vocab_size], device=device)
+
+        # A flag indicating a frame is a blank frame or not.
+        # 0 for blank frame and 1 for non-blank frame.
+        # Used to select non-blank frames for n-gram rescoring.
+        non_blank_flag = torch.zeros([total_t], device=device)
+
     offset = 0
     for batch_size in batch_size_list:
         start = offset
@@ -600,7 +616,36 @@ def greedy_search_batch(
         # logits'shape (batch_size, 1, 1, vocab_size)
 
         logits = logits.squeeze(1).squeeze(1)  # (batch_size, vocab_size)
-        assert logits.ndim == 2, logits.shape
+
+        if ngram_rescoring:
+            all_logits[start:end] = logits
+
+            assert logits.ndim == 2, logits.shape
+            logits_argmax = logits.argmax(dim=1)
+            logits_softmax = logits.softmax(dim=1)
+
+            # detailed in below fuction verify_non_blank_logits.
+            selection_verification = True
+
+            # 0 for blank frame and 1 for non-blank frame.
+            non_blank_flag[start:end] = torch.where(
+                logits_argmax == blank_id, 0, 1
+            )
+
+            if False:
+                # In paper: https://arxiv.org/pdf/2101.06856.pdf
+                # A gama_blank threshold value is used to determinze blank frame.
+                # Currently, results are worse than baseline greedy_search
+                #  and also very sensitive to gama_blank.
+                # (TODO): debug this later.
+                gama_blank = 0.50
+                non_blank_flag[start:end] = torch.where(
+                    logits_softmax[:, 0] >= gama_blank, 0, 1
+                )
+
+                # function verify_non_blank_logits only works with logits_argmax == blank_id.
+                selection_verification = False
+
         y = logits.argmax(dim=1).tolist()
         emitted = False
         for i, v in enumerate(y):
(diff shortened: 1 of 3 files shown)


egs/librispeech/ASR/pruned_transducer_stateless2/beam_search.py
```diff
@@ -600,7 +616,36 @@ def greedy_search_batch(
...
+                logits_argmax == blank_id, 0, 1
+            )
+
+            if False:
+                # In paper: https://arxiv.org/pdf/2101.06856.pdf
+                # A gama_blank threshold value is used to determinze blank frame.
+                # Currently, results are worse than baseline greedy_search
+                #  and also very sensitive to gama_blank.
+                # (TODO): debug this later.
+                gama_blank = 0.50
```
yuekaizhang, line 641: Did you try large gama_blank value? 0.5 means p_t_u(blank) > 0.5 would be skipped. Maybe 0.95 or 0.99 better?

egs/librispeech/ASR/pruned_transducer_stateless2/beam_search.py
```diff
@@ -624,6 +669,105 @@ def greedy_search_batch(
...
+    logits_to_rescore_softmax = non_blank_logits.softmax(dim=2)
+    logits_to_rescore = logits_to_rescore_softmax.log()
+
+    # In paper: https://arxiv.org/pdf/2101.06856.pdf
+    # blank deweight is applied before non_blank frames selected.
+    # However, in current setup, that results in a higher WER.
+    # So just put this blank deweight before ngram rescoring.
+    # (TODO): debug this blank deweight issue.
+
+    blank_deweight = 100
```
yuekaizhang, line 739: We previously tuned deweight value from 1-3, and could observe the deletion errors improvement. 100 may be too large?

egs/librispeech/ASR/pruned_transducer_stateless2/beam_search.py
glynpu: Thanks for the information.
Actually I tried several weights and 0.5 is the best among them.  Also the result is quite sensitive to game_blank.
Maybe  because we use different WFSTDecoding methods, or current 3-gram is not appropriate for this task.
gama_blank|test-clean|detail error
--|--|--
0.4|20.75|Errors: 365 insertions, 5187 deletions, 5357 substitutions, over 52576 reference words
0.5|2.57|Errors: 331 insertions, 81 deletions, 994 substitutions, over 52576 reference words
0.6|21.66|Errors: 7788 insertions, 146 deletions, 3456 substitutions, over 52576 reference words

egs/librispeech/ASR/pruned_transducer_stateless2/beam_search.py
glynpu: with current decoding method and 3-gram, blank_deweight only has tiny contribution, I tried [0, 10, 50, 100] and results are the same.

Maybe the bottleneck is the 3-gram or decoding method.

glynpu:
There are some OOV words which are successfully recognized by baseline greedy search but failed by 3-gram based WFSTDecoding. 
Since these words neither in words.txt nor 3-gram.pruned.1e-7.arpa, it's impossible to recognize them with WFSTDecoding.


Some example results related OOV words.
OOV word|greedy search | 3-gram based WFSTDecoding
--|--|--
CHELFORD |SAID LORD CHELFORD ADDRESSING ME|SAID LORD (CHELFORD->C SHELFORD) ADDRESSING ME
UNCASING |COME COME RETURNED HAWKEYE UNCASING HIS HONEST|COME COME RETURNED HAWKEYE (UNCASING->ENCASING) HIS HONEST
ELECTROPLATING|IT COULD NOT BE USED FOR ELECTROPLATING OR DEPOSITION|IT COULD NOT BE USED FOR (ELECTROPLATING->ELECTRO COLLATING) OR DEPOSITION
GAMEWELL|THESE ESCAPADES ARE NOT FOR OLD GAMEWELL LAD HIS DAY HAS COME TO TWILIGHT|THESE ESCAPADES ARE NOT FOR OLD (GAMEWELL->GAMWELL) LAD HIS DAY HAS COME TO TWILIGHT
MAINHALL |MAINHALL VOUCHED FOR HER CONSTANCY|(MAINHALL->MAIN HALL) VOUCHED FOR HER CONSTANCY

glynpu:
>  dictionary-based decoding on Librispeech with Librispeech's provided dictionary is simply not capable of state of the art WERs

I think so.

 A better result is obtained (2.04||4.72 --> 2.02||4.66) with two mofidications:

1.  including OOV words into word.txt
2.  adding self loops to L.pt.


decoding with gamma_blank=0.5 when L or (L + self loop) used.
decoding-graph|words.txt|test-clean|test-other
--|--|--|--
base greedy search|librispeech lang_phone |2.04|4.72
L|librispeech lang_phone|2.81|6.01
L + self loop|librispeech lang_phone|2.34|4.89
L + self loop|+train words|2.11|4.75
L + self loop|+train and test words|2.02|4.66


Also, It's NOT so sensitive to gamma_blank any more with L + self loop, since a blank is allowed inside a word.
 
gamma_blank|test-clean|test-other
--|--|--
0.3|35.83|36.98
0.4|19.92|21.91
0.5|2.02|4.66
0.6|2.02|4.66
0.7|2.01|4.68
0.8|2.01|4.65
0.9|2.01|4.65
1.0|2.01|4.74



Conclusion: if all OOV words are included into L,  WER could decrease from (2.04||4.72) to (2.02||4.66) even without using any n-gram. (Note: previous results are obtained with only a L, not LG)

A better results maybe obtained based on this if we manage solve OOV problem in n_grams.

egs/librispeech/ASR/pruned_transducer_stateless2/beam_search.py
glynpu: >  I tried [0, 10, 50, 100] and results are the same.


It seems a problem of decoding_graph. gamma_blank works as expected after adding self loop to L.

danpovey:
Ah I see, they are epsilon self loops.  Yes, I think the results would have been wrong before adding these, if this made a difference.  In fact, if the decoding or rescoring process requires epsilon self-loops, I'm a little unclear how it was working at all.  I don't understand the details of what you are doing, but please make sure that you understand exactly why adding self-loops helped.

glynpu:
>  I don't understand the details of what you are doing

The long-term goal is PSD based n-gram rescoring method, which mainly contains two steps:
1. Using greedy_search_bath as the first-pass decoding and selected non-blank frames with a threshold `gamma_blank`;
2.  Only selected non-blank frames are used during n-gram rescoring, NOT all frames. This will  improve rescoring efficiency.  According to my observations now, only 30%~50% frames are needed during rescoring pass.

 What I have done now is kind of a "**ablation study**", since my first try with "HLG" and "Trivial_graph_LG" failed. So I want to figure out what will happen after integrating "Trivial_graph", "L", "G" seperately.

Now the better results is obtained by rescoring with "L" if all OOV words are included in words.txt.


>  why adding self-loops helped.

The main idea is to allow blanks  in a word during decoding.  

Because in our setup, blank_id == 0. k2.add_epsilon_self_loops  works like `add_blank_self_loops` to each state.

Here is an example:
Suppose words.txt only contains these:
```
<eps> 0                                                                                       │
<UNK> 1                                                                                       │
AUDIOBOOK 2                                                                                   │
CHELFORD 3
```

This is the comparison before and after k2.add_epsilon_self_loops. 

image

glynpu:
> Now the better results is obtained by rescoring with "L" if all OOV words are included in words.txt.

Question: Why adding L helps, even it contains no information to be rescored like ngram?

Answer: In theory, greedy_search + bpe model could recognize any word, but sometimes it will generates non-existent word.
While dictionary-based decoding could only generate words from words.txt when building L. But this is not necessary a bad thing, since all words generated by L are meaningful word.

If all OOV words are added into words.txt, "rescoring" with L will stop bpe model generate a non-existent word.

Here are some results comparison between greedy search  and "rescoring" with L.
decoding-graph|words.txt|test-clean|test-other
--|--|--|--
base greedy search|librispeech lang_phone |2.04|4.72
L + self loop|+train and test words|2.02|4.66

greedy search| "rescore" with (L + self loop)
--|--
WHO BEING THE FRIEND OF (DROPIDAS->TROPIDUS) MY GREAT GRANDFATHER|WHO BEING THE FRIEND OF DROPIDAS MY GREAT GRANDFATHER
THE ABILITY OF (BALAAM'S->BALAM'S) ASS TO SPEAK|THE ABILITY OF BALAAM'S ASS TO SPEAK

danpovey:
OK.  When you added all OOVs, I assume this is a "cheating" experiment, since they were derived from the test set?
The thing which confused me is, how did it work at all previously, with the L without self-loops?  Like, why wasn't the WER close to 100%?  See if y

glynpu:
>  I assume this is a "cheating" experiment, since they were derived from the test set?

Yes, it's "cheating" if including words from test test. So I teat these OOVs as two kinds.
1. from train set. (This is not cheating, I think)
2. from test set. (This is indeed cheating. This tells us the influence of OOVs to dictonary-based  decoding. Previously, we have tried several n-gram based rescoring, only obtain limit gains or even no gains. Maybe OOV is a potential reason).

> how did it work at all previously, with the L without self-loops? 

With no-self-loop L, I tried two following methods to select non-blank frames. 
```
# method 1
non_blank_flag[start:end] = torch.where(logits_argmax == blank_id, 0, 1)
# method2
non_blank_flag[start:end] = torch.where(logits_softmax[:, 0] >= gamma_blank, 0, 1)
```

Without adding self-loops to L, only method 1 words , while method 2 is very sensitive to gamma blank.

With method 1, all blank frames are filtered out, there is no need to handle "blanks" inside a word.

With method2,  it only works with a specific gamma_blank threshold, whose results may be similar to that of method 1, i.e. no "blank frames" left inside a word.
However, if with a larger gamma_blank, more blank frames would be left inside a word. The no-self-loop-L doesn't know how to handle these blank frames.  
That's why method 2 still works at some point but very sensitive to gamma_blank threshold.

PR #479, opened 2022-07-17
Use ScaledLSTM as streaming encoder

yaozengwei:
This PR implements `ScaledLSTM` as a light streaming encoder with the scaling mechanism in reworked model. We do not need to use time masking during training as in streaming `Conformer`, since the LSTM computation is essentially streaming. 

The `ScaledLSTM` class is written in `pruned_transducer_stateless2/scaling.py`, which is modified from https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/rnn.py, using the highly optimized operator `_VF.lstm` in pytorch. 
@danpovey  Could you help me check for correctness of the value of `fan_in` in `_reset_parameters` function if you are free? 

In `lstm_transducer_stateless/lstm.py`, each encoder layer consists of a `lstm` module and a `feed_forward` module, which adopts similar residual connection as in `Conformer`.


egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
```diff
@@ -326,12 +326,12 @@ def __init__(
...
     def _reset_parameters(self, initial_speed: float):
         std = 0.1 / initial_speed
-        a = (3**0.5) * std
+        a = (3 ** 0.5) * std
         nn.init.uniform_(self.weight, -a, a)
         if self.bias is not None:
             nn.init.constant_(self.bias, 0.0)
         fan_in = self.weight.shape[1] * self.weight[0][0].numel()
-        scale = fan_in**-0.5  # 1/sqrt(fan_in)
+        scale = fan_in ** -0.5  # 1/sqrt(fan_in)
```
danpovey, line 334: The reset_parameters in torch rnn.py looks like:
```
    def reset_parameters(self) -> None:
        stdv = 1.0 / math.sqrt(self.hidden_size)
        for weight in self.parameters():
            init.uniform_(weight, -stdv, stdv)
```
so I suggest to set: 
```
     scale = self.hidden_size ** -0.5
```

yaozengwei: OK. Thanks.

egs/librispeech/ASR/pruned_transducer_stateless3/scaling_converter.py
```diff
@@ -164,6 +170,38 @@ def scaled_embedding_to_embedding(
     return embedding
 
 
+def scaled_lstm_to_lstm(scaled_lstm: ScaledLSTM):
```
csukuangfj, line 173: Please add type hint for the return value.

yaozengwei: Ok. Thanks.

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
```diff
@@ -377,6 +378,155 @@ def forward(self, input: Tensor) -> Tensor:
...
+    # This class implements LSTM with scaling mechanism, using `torch._VF.lstm`
+    # Please refer to https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/rnn.py
+    def __init__(
+        self,
+        *args,
+        initial_scale: float = 1.0,
+        initial_speed: float = 1.0,
+        **kwargs
+    ):
+        # Hardcode bidirectional=False
```
csukuangfj, line 392: I suggest using
```python3
if 'bidirectional' in kwargs:
  assert kwargs['bidirectional'] is False
```
`bidirectional` is `False` by default, so you don't need to pass it below.

egs/librispeech/ASR/lstm_transducer_stateless/test_scaling_converter.py
```diff
@@ -0,0 +1,257 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+To run this file, do:
+
+    cd icefall/egs/librispeech/ASR
+    python ./pruned_transducer_stateless3/test_scaling_converter.py
```
csukuangfj, line 22:
```suggestion
    python ./lstm_transducer_stateless/test_scaling_converter.py
```

yaozengwei: Ok. Thanks.

yaozengwei:
Currently, it only supports using `jit.trace` to export the SclaedLSTM model.

Pytorch does not support `jit.script` for nn.LSTM class, see https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/rnn.py#L909 for detail.

Thers is an issue while exporting the model using `onnx`. `torch.onnx` does not support the case of using `projections` in LSTM class. See https://github.com/pytorch/pytorch/blob/master/torch/onnx/symbolic_opset9.py#L3666

csukuangfj:
Support only torch.jit.trace is fine with me.

egs/librispeech/ASR/lstm_transducer_stateless/export.py
```diff
@@ -0,0 +1,388 @@
+#!/usr/bin/env python3
+#
+# Copyright 2021 Xiaomi Corporation (Author: Fangjun Kuang)
```
csukuangfj, line 3: Please change the author.

egs/librispeech/ASR/lstm_transducer_stateless/streaming_decode.py
```diff
@@ -0,0 +1,968 @@
...
+    """
+    Args:
+      model:
+        The Transducer model.
+      streams:
+        A list of Stream objects.
+      params:
+        It is returned by :func:`get_params`.
+      decoding_graph:
+        The decoding graph. Can be either a `k2.trivial_graph` or HLG, Used
```
csukuangfj, line 514:
```suggestion
        The decoding graph. Can be either a `k2.trivial_graph` or LG, Used
```

egs/librispeech/ASR/lstm_transducer_stateless/train.py
```diff
@@ -0,0 +1,1082 @@
...
+            "model_warm_step": 3000,  # arg given to model, not for lrate
+            "env_info": get_env_info(),
+        }
+    )
+
+    return params
+
+
+def get_encoder_model(params: AttributeDict) -> nn.Module:
+    # TODO: We can add an option to switch between Conformer and Transformer
```
csukuangfj, line 390:
```suggestion
```

egs/librispeech/ASR/lstm_transducer_stateless/train.py
```diff
@@ -0,0 +1,1082 @@
...
+            prune_range=params.prune_range,
+            am_scale=params.am_scale,
+            lm_scale=params.lm_scale,
+            warmup=warmup,
+        )
+        # after the main warmup step, we keep pruned_loss_scale small
+        # for the same amount of time (model_warm_step), to avoid
+        # overwhelming the simple_loss and causing it to diverge,
+        # in case it had not fully learned the alignment yet.
+        pruned_loss_scale = (
```
csukuangfj, line 616: Please use the changes from the master to filter nan/inf losses.

csukuangfj (commented):
Thanks!

Great work! Just left some minor comments.

egs/librispeech/ASR/lstm_transducer_stateless/train.py
yaozengwei: Ok. Thanks for the detailed review. I will modify them.

RuABraun:
What is the recommended recipe if one wants to do online ASR? Not clear to me whether this is recommended over the streaming conformer stuff or not? @danpovey

RuABraun:
Trained a model on 1k hours and got twice the WER compared to a conformer model (greedy search) unfortunately.

csukuangfj:
> Trained a model on 1k hours and got twice the WER compared to a conformer model (greedy search) unfortunately.

Do you mean the streaming conformer?

---

https://github.com/k2-fsa/icefall/pull/558 shows that when you use a larger dataset, e.g., 30k hours of data, ScaledLSTM
has competitive results with the streaming conformer in icefall.

RuABraun:
Interesting. I just realized I wasn't using musan augmentation which I presume is suboptimal.

I'm comparing to the normal conformer, the impression I got from the [librispeech results file](https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/RESULTS.md) was that the lstm model will keep up with pruned_transducer_stateless2.

It does make sense to me that a 12 layer LSTM model needs a lot of data to work well.

PR #484, opened 2022-07-20
Update conformer.py for aishell4

luomingshuang:


luomingshuang:
The conformer.py has changed comparing to the used version for pruned rnnt5 aishell4.

luomingshuang:
I have test it by decoding with the pretrained model and compiling torchscript model.

csukuangfj:
> I have test it by decoding with the pretrained model and compiling torchscript model successfully.

Could you replace it with a symlink to the one in the librispeech recipe?

PR #488, opened 2022-07-22
fixing TypeError during loading onnx model.

EmreOzkose:
When exporting a Pytorch model to onnx version of itself, if there is a constant in the model, loading onnx model raises error.  A simple example is given below:

```
import torch
import numpy as np
import torch.nn as nn

import onnx
import onnxruntime as ort


class ExampleModule(nn.Module):
    def __init__(self) -> None:
        super().__init__()

    def forward(self, x: torch.Tensor):
        x = x - 1.0
        return x


example_module = ExampleModule()
example_in = torch.Tensor([[1, 2, 3, 4], [5, 6, 7, 8]])

torch.onnx.export(example_module,
                  args=(example_in),
                  f="example_module.onnx",
                  input_names=["x"],
                  output_names=["output1"],
                  opset_version = 11,
                  dynamic_axes={
                      'x': {0 : 'batch_size'},
                      'output1': {0 : 'batch_size'},
                  },
                  verbose=True)

model = onnx.load("example_module.onnx")
onnx.checker.check_model(model)

ort_session = ort.InferenceSession("example_module.onnx")
```

`ort.InferenceSession` raises error:
```
---------------------------------------------------------------------------
Fail                                      Traceback (most recent call last)
Input In [6], in <cell line: 8>()
      5 model = onnx.load("example_module.onnx")
      6 onnx.checker.check_model(model)
----> 8 ort_session = ort.InferenceSession("example_module.onnx")
     10 outputs = ort_session.run(
     11     None,
     12     {
     13         "x": example_in.numpy().astype(np.float32),
     14     },
     15 )
     16 print(example_in)

File ~/anaconda3/envs/k2_sherpa/lib/python3.8/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py:335, in InferenceSession.__init__(self, path_or_bytes, sess_options, providers, provider_options, **kwargs)
    332 disabled_optimizers = kwargs['disabled_optimizers'] if 'disabled_optimizers' in kwargs else None
    334 try:
--> 335     self._create_inference_session(providers, provider_options, disabled_optimizers)
    336 except ValueError:
    337     if self._enable_fallback:

File ~/anaconda3/envs/k2_sherpa/lib/python3.8/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py:370, in InferenceSession._create_inference_session(self, providers, provider_options, disabled_optimizers)
    368 session_options = self._sess_options if self._sess_options else C.get_default_session_options()
    369 if self._model_path:
--> 370     sess = C.InferenceSession(session_options, self._model_path, True, self._read_config_from_model)
    371 else:
    372     sess = C.InferenceSession(session_options, self._model_bytes, False, self._read_config_from_model)

Fail: [ONNXRuntimeError] : 1 : FAIL : Load model from example_module.onnx failed:Type Error: Type parameter (T) of Optype (Sub) bound to different types (tensor(float) and tensor(double) in node (Sub_1).
```

Hence, I traced exported graph and converted type of problematic tensors from Double to Float. I just multiple these tensors with `torch.ones((1),  dtype=torch.float32)`.

### DIFF
--- a/egs/librispeech/ASR/RESULTS.md
+++ b/egs/librispeech/ASR/RESULTS.md
@@ -618,6 +618,80 @@ done
 
 Pre-trained models, training and decoding logs, and decoding results are available at <https://huggingface.co/pkufool/icefall_librispeech_streaming_pruned_transducer_stateless4_20220625>
 
+#### [pruned_transducer_stateless5](./pruned_transducer_stateless5)
+
+See <https://github.com/k2-fsa/icefall/pull/454> for more details.
+
+##### Training on full librispeech
+The WERs are (the number in the table formatted as test-clean & test-other):
+
+We only trained 25 epochs for saving time, if you want to get better results you can train more epochs.
+
+| decoding method      | left context | chunk size = 2 | chunk size = 4 | chunk size = 8 | chunk size = 16|
+|----------------------|--------------|----------------|----------------|----------------|----------------|
+| greedy search        | 32           | 3.93 & 9.88    | 3.64 & 9.43    | 3.51 & 8.92    | 3.26 & 8.37    |
+| greedy search        | 64           | 4.84 & 9.81    | 3.59 & 9.27    | 3.44 & 8.83    | 3.23 & 8.33    |
+| fast beam search     | 32           | 3.86 & 9.77    | 3.67 & 9.3     | 3.5 & 8.83     | 3.27 & 8.33    |
+| fast beam search     | 64           | 3.79 & 9.68    | 3.57 & 9.21    | 3.41 & 8.72    | 3.25 & 8.27    |
+| modified beam search | 32           | 3.84 & 9.71    | 3.66 & 9.38    | 3.47 & 8.86    | 3.26 & 8.42    |
+| modified beam search | 64           | 3.81 & 9.59    | 3.58 & 9.2     | 3.44 & 8.74    | 3.23 & 8.35    |
+
+
+**NOTE:** The WERs in table above were decoded with simulate streaming method (i.e. using masking strategy), see commands below. We also have [real streaming decoding](./pruned_transducer_stateless5/streaming_decode.py) script which should produce almost the same results. We tried adding right context in the real streaming decoding, but it seemed not to benefit the performance for all the models, the reasons might be the training and decoding mismatching.
+
+The training command is:
+
+```bash
+./pruned_transducer_stateless5/train.py \
+  --exp-dir pruned_transducer_stateless5/exp \
+  --num-encoder-layers 18 \
+  --dim-feedforward 2048 \
+  --nhead 8 \
+  --encoder-dim 512 \
+  --decoder-dim 512 \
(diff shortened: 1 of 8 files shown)


egs/librispeech/ASR/pruned_transducer_stateless2/conformer.py
```diff
@@ -1098,7 +1098,7 @@ def multi_head_attention_forward(
             head_dim * num_heads == embed_dim
         ), "embed_dim must be divisible by num_heads"
 
-        scaling = float(head_dim) ** -0.5
+        scaling = torch.ones((1), dtype=torch.float32) * float(head_dim) ** -0.5
```
pkufool, line 1101: Do we need the `device` argument here? I suspect it would raise an error when training.

EmreOzkose: I tried to export with `device=cuda`, and it raises device mismatch errors, I am adding device argument.

csukuangfj:
Instead of using
```python3
example_module = ExampleModule()
example_in = torch.Tensor([[1, 2, 3, 4], [5, 6, 7, 8]])

torch.onnx.export(example_module,
```

Have you tried
```python3
example_module = ExampleModule()
example_in = torch.Tensor([[1, 2, 3, 4], [5, 6, 7, 8]])

torch.onnx.export(torch.jit.script(example_module),
```

That is, use `torch.jit.script(example_module)`, not `example_module`.

If you pass `example_module`, it is equivalent to pass `torch.jit.trace(example_module)`.

egs/librispeech/ASR/pruned_transducer_stateless2/conformer.py
```diff
@@ -1098,7 +1098,10 @@ def multi_head_attention_forward(
             head_dim * num_heads == embed_dim
         ), "embed_dim must be divisible by num_heads"
 
-        scaling = float(head_dim) ** -0.5
+        scaling = (
```
csukuangfj, line 1101: Can we register it as a buffer in the constructor?

EmreOzkose: Actually I just closed laptop on friday, and opened again on monday (today), then I am not facing with this problem anymore. Same environment, same device, same codes. torch.onnx.export() function exports a graph which doesn't contain double type tensor. There are same issues: [1](https://github.com/microsoft/CodeXGLUE/issues/65), [2](https://github.com/pytorch/pytorch/issues/58367), [3](https://github.com/microsoft/onnxruntime/issues/1764). I also create a new environment to re-produce, but I couldn't reproduce, everything works well. Still, I can register it as a buffer in the constructor and make a commit, someone may face with this in the future?

csukuangfj: In that case, I suggest leaving the current code as it is.

We can fix it if someone can reproduce it.

PR #489, opened 2022-07-22
Add modified_beam_search for streaming decode

pkufool:
- [ ] Add to other recipes

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless/streaming_beam_search.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless/streaming_beam_search.py
@@ -1,0 +1,278 @@
+# Copyright    2022  Xiaomi Corp.        (authors: Wei Kang)
+#
+# See ../../../../LICENSE for clarification regarding multiple authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import warnings
+from typing import List
+
+import k2
+import torch
+import torch.nn as nn
+from beam_search import Hypothesis, HypothesisList, get_hyps_shape
+from decode_stream import DecodeStream
+
+from icefall.decode import one_best_decoding
+from icefall.utils import get_texts
+
+
+def greedy_search(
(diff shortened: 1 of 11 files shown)


egs/librispeech/ASR/pruned_transducer_stateless/streaming_beam_search.py
```diff
@@ -0,0 +1,278 @@
...
+    context_size = model.decoder.context_size
+    device = model.device
+    T = encoder_out.size(1)
+
+    decoder_input = torch.tensor(
+        [stream.hyp[-context_size:] for stream in streams],
+        device=device,
+        dtype=torch.int64,
+    )
+    # decoder_out is of shape (N, decoder_out_dim)
```
csukuangfj, line 58:
```suggestion
    # decoder_out is of shape (N, 1, decoder_out_dim)
```

egs/librispeech/ASR/pruned_transducer_stateless/streaming_beam_search.py
```diff
@@ -0,0 +1,278 @@
...
+                decoder_input,
+                need_pad=False,
+            )
+
+
+def modified_beam_search(
+    model: nn.Module,
+    encoder_out: torch.Tensor,
+    streams: List[DecodeStream],
+    beam: int = 4,
```
csukuangfj, line 96: I suggest using `num_active_paths` in the **new** code.

People can get confused with `beam` and `beam_size`.

egs/librispeech/ASR/pruned_transducer_stateless/streaming_decode.py
```diff
@@ -31,7 +31,6 @@
 import argparse
 import logging
 import math
-import warnings
```
csukuangfj: In line 20, it should be
```
./pruned_transducer_stateless/streaming_decode.py \
```

egs/librispeech/ASR/pruned_transducer_stateless/streaming_beam_search.py
pkufool: I know you renamed it to `num_active_paths` in sherpa, I thought it is better to make it be consistent with the one in `decode.py`. So do we need to rename the one in `decode.py`?

csukuangfj: I suggest only doing this for **new** code.

The current file associates two different meanings with `beam`, which can confuse users, I think.
https://github.com/k2-fsa/icefall/blob/72d76a4ff8aa43b88f967a37c0346590e4ecbfca/egs/librispeech/ASR/pruned_transducer_stateless/streaming_beam_search.py#L96

https://github.com/k2-fsa/icefall/blob/72d76a4ff8aa43b88f967a37c0346590e4ecbfca/egs/librispeech/ASR/pruned_transducer_stateless/streaming_beam_search.py#L203

PR #493, opened 2022-07-25
Support quantization

csukuangfj:
```
-rw-r--r-- 1 kuangfangjun root 326M Jul 25 18:03 pruned_transducer_stateless3/exp-0.9/cpu_jit.pt
-rw-r--r-- 1 kuangfangjun root 162M Jul 25 17:59 pruned_transducer_stateless3/exp-0.9/quantized_cpu_jit.pt
```

I tested `quantized_cpu_jit.pt` with sherpa and it can decode waves correctly.

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless2/quantize.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless2/quantize.py
@@ -1,0 +1,133 @@
+# Copyright    2022  Xiaomi Corp.        (authors: Fangjun Kuang)
+#
+# See ../../../../LICENSE for clarification regarding multiple authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import copy
+import re
+
+import torch
+from scaling import ScaledLinear
+
+
+def _get_weight(self: torch.nn.Linear):
+    return self.weight
+
+
+def _get_bias(self: torch.nn.Linear):
+    return self.bias
+
+
+def scaled_linear_to_linear(scaled_linear: ScaledLinear) -> torch.nn.Linear:
+    """Convert a ScaledLinear layer to a Linear layer.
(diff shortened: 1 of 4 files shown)


csukuangfj:
Note:
https://github.com/k2-fsa/icefall/blob/2777c0b0b37eb4cc514fe2da55884796cf93f54f/egs/librispeech/ASR/pruned_transducer_stateless2/quantize.py#L90-L96

`in_proj` and `out_proj` ScaledLinear layers in `RelPositionMultiheadAttention` are not quantized:
https://github.com/k2-fsa/icefall/blob/b1d0956855d5a113a1f3b25eadf7bd60f8a61935/egs/librispeech/ASR/pruned_transducer_stateless2/conformer.py#L892-L895

The reason is that we pass their `weight` and `bias` directly to `multi_head_attention_forward`:
https://github.com/k2-fsa/icefall/blob/b1d0956855d5a113a1f3b25eadf7bd60f8a61935/egs/librispeech/ASR/pruned_transducer_stateless2/conformer.py#L972-L989

Had we also quantized them, we had to dequantize them before passing them to `multi_head_attention_forward`, which would affect running speed.

If we care more about RAM usage and care less about speed, we can also quantize them.

---

Note:

`linear_pos` in `RelPositionMultiheadAttention` is quantized.  We try to quantize as many ScaledLinear layers as possible.

https://github.com/k2-fsa/icefall/blob/b1d0956855d5a113a1f3b25eadf7bd60f8a61935/egs/librispeech/ASR/pruned_transducer_stateless2/conformer.py#L898

egs/librispeech/ASR/pruned_transducer_stateless2/quantize.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    if not inplace:
+        model = copy.deepcopy(model)
+
+    d = {}
+    excluded_patterns = r"self_attn\.(in|out)_proj"
+    p = re.compile(excluded_patterns)
+    for name, m in model.named_modules():
+        if isinstance(m, ScaledLinear):
+            if p.search(name) is not None:
```
csukuangfj, line 94: See https://github.com/k2-fsa/icefall/pull/493#issuecomment-1193933580
for why we do it in this way.

csukuangfj:
Here is the decoding results comparison between a quantized torchscript model and a not quantized torchscript model.
(The cell contains WERs for `test_clean/test_other`)

| model type | greedy search | modified_beam_search | fast_beam_search |
|---|----|----|----|
|quantized|2.26/5.45| 2.16/5.27 | 2.27/5.32|
|not quantized| 2.07/4.82|2.00/4.63 | 2.14/4.77|

It shows quantization has a negative impact on WERs. test-other is more sensitive to quantization.

PR #514, opened 2022-08-03
Use explicit relative imports

shaynemei:
- switch from implicit relative imports to explicit relative imports as recommended by Guido: https://peps.python.org/pep-0328/#guido-s-decision
- sort import order using `isort`

### DIFF
--- a/egs/aidatatang_200zh/ASR/local/prepare_char.py
+++ b/egs/aidatatang_200zh/ASR/local/prepare_char.py
@@ -39,6 +39,7 @@ from typing import Dict, List
 
 import k2
 import torch
+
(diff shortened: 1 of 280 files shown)


csukuangfj:
Could you please fix the CI errors?

danpovey:
Do the scripts actually run correctly when using relative imports?
In my experience, if you use relative imports like that and you run the scripts in the normal way, like
python3 pruned_transducer_stateless2/decode.py
you get errors like "cannot use relative import because decode.py is not part of a package"
(this is from memory, may not be exact).
That's why I don't tend to use them.

Issue #240, opened 2022-03-06
RuntimeError for class values on conformer_ctc.


mesut92:
Hi k2
I am trying to run "librispeech/ASR/conformer_ctc/". I build docker image from your Dockerfile. I use Librispeech dataset.  However it gives an error, and i could not fix. 

I use thşs command:
`conformer_ctc/train.py --max-duration 140`

```
`Traceback (most recent call last):
  File "conformer_ctc/train.py", line 799, in <module>
    main()
  File "conformer_ctc/train.py", line 792, in main
    run(rank=0, world_size=1, args=args)
  File "conformer_ctc/train.py", line 685, in run
    scan_pessimistic_batches_for_oom(
  File "conformer_ctc/train.py", line 758, in scan_pessimistic_batches_for_oom
    loss, _ = compute_loss(
  File "conformer_ctc/train.py", line 416, in compute_loss
    att_loss = mmodel.decoder_forward(
  File "/media/mesut/Tosun/ses/010322k2/icefall/egs/librispeech/ASR/conformer_ctc/transformer.py", line 313, in decoder_forward
    decoder_loss = self.decoder_criterion(pred_pad, ys_out_pad)
  File "/opt/conda/lib/python3.8/site-packages/torch/nn/modules/module.py", line 727, in _call_impl
    result = self.forward(*input, **kwargs)
  File "/media/mesut/Tosun/ses/010322k2/icefall/egs/librispeech/ASR/conformer_ctc/label_smoothing.py", line 81, in forward
    true_dist = torch.nn.functional.one_hot(
RuntimeError: Class values must be non-negative.
`
```

Thanks, 
Mesut

csukuangfj:
Could you change
https://github.com/k2-fsa/icefall/blob/1603744469d167d848e074f2ea98c587153205fa/egs/librispeech/ASR/conformer_ctc/train.py#L413-L416

to

```python3
 unsorted_token_ids = graph_compiler.texts_to_ids( 
     supervisions["text"] 
 ) 
 import pdb
 pdb.set_trace()
 att_loss = mmodel.decoder_forward(
```

When it enters pdb, you can print out the value of `unsorted_token_ids` to see whether it is `None`.

mesut92:
my output:

(Pdb) print(unsorted_token_ids)
[...]

PR #525, opened 2022-08-06
Filter non-finite losses

csukuangfj:
See
https://github.com/k2-fsa/icefall/pull/396#issuecomment-1207277213

> The pruned loss is inf. @csukuangfj perhaps we should modify the code that sums up the losses, in icefall/icefall/, to filter out inf's and nan's before incorporating them into the sum?

---

Would be nice if someone can help to propagate the changes to other directories.

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless5/train.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless5/train.py
@@ -655,7 +655,23 @@ def compute_loss(
             am_scale=params.am_scale,
             lm_scale=params.lm_scale,
             warmup=warmup,
+            reduction="none",
         )
+        simple_loss_is_finite = torch.isfinite(simple_loss)
+        pruned_loss_is_finite = torch.isfinite(pruned_loss)
+        is_finite = simple_loss_is_finite & pruned_loss_is_finite
+        if not torch.all(is_finite):
+            logging.info(
+                "Not all losses are finite!\n"
+                f"simple_loss: {simple_loss}\n"
+                f"pruned_loss: {pruned_loss}"
+            )
+            display_and_save_batch(batch, params=params, sp=sp)
+            simple_loss = simple_loss[simple_loss_is_finite]
+            pruned_loss = pruned_loss[pruned_loss_is_finite]
+
+        simple_loss = simple_loss.sum()
+        pruned_loss = pruned_loss.sum()
         # after the main warmup step, we keep pruned_loss_scale small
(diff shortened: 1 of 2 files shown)


egs/librispeech/ASR/pruned_transducer_stateless5/train.py
```diff
@@ -675,6 +691,7 @@ def compute_loss(
     info = MetricsTracker()
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
+        # info["frames"] is an approximate number
```
danpovey, line 694: perhaps should say why this is so? (presumably due to filtering inf/nan losses).
also, I suggest that if all of the either simple or pruned losses are infinite and the
batch has more than, say, 10 items, we could crash?  it's a waste of compute to 
continue to train a model that has diverged.  (The idea here is that 1 or 2 losses can
be infinite for valid reasons, but it's unlikely for a whole batch of 10; I'm picking a number
such that most setups will sometimes processa at least that many batches.  And it doesn't
require adding extra numbers to keep track off like the number of failures).

csukuangfj: Yes, will add it later today.

csukuangfj: Fixed. Please have a review again.

Sorry for the late reply.

csukuangfj:
@marcoyang1998 

Could you please help to propagate changes from this PR to other `pruned_transducer_statelessX` folders?

egs/librispeech/ASR/pruned_transducer_stateless5/train.py
```diff
@@ -655,7 +655,35 @@ def compute_loss(
...
+        pruned_loss_is_finite = torch.isfinite(pruned_loss)
+        is_finite = simple_loss_is_finite & pruned_loss_is_finite
+        if not torch.all(is_finite):
+            logging.info(
+                "Not all losses are finite!\n"
+                f"simple_loss: {simple_loss}\n"
+                f"pruned_loss: {pruned_loss}"
+            )
+            display_and_save_batch(batch, params=params, sp=sp)
+            simple_loss = simple_loss[simple_loss_is_finite]
```
desh2608, line 670: Just a heads up: I was facing the same logical indexing issue here that we had [seen earlier](https://github.com/k2-fsa/icefall/issues/240#issuecomment-1065402350), so I had to replace this with the `torch.where()`. I suppose this might just be a CUDA 11.1 issue after all.

PR #527, opened 2022-08-09
WIP: Support exporting pretrained models to mace/ncnn and others via pnnx

csukuangfj:
The aim of this PR is to support exporting pretrained models in a way such that they can be used by other inference frameworks such as
- [ ] [mace](https://github.com/xiaomi/mace)
- [ ] [ncnn](https://github.com/tencent/ncnn)

and possibly others like: [tnn](https://github.com/tencent/tnn), [mnn](https://github.com/alibaba/MNN), etc.

We can also use the quantization support from the above frameworks, e.g., using full int8 support from [ncnn](https://github.com/tencent/ncnn)

As the first step, I am using [pnnx](https://github.com/Tencent/ncnn/tree/master/tools/pnnx) as the intermediate format.

---

It requires
https://github.com/csukuangfj/ncnn/pull/1

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless2/conformer.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless2/conformer.py
@@ -1592,6 +1592,10 @@ class Conv2dSubsampling(nn.Module):
             channel_dim=-1, min_positive=0.45, max_positive=0.55
         )
 
+        # ncnn support only batch size == 1
+        self.for_ncnn = False
+        self.conv_out_dim = self.out.weight.shape[1]
+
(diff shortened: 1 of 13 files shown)


csukuangfj:
It requires
https://github.com/csukuangfj/ncnn/pull/1

yaozengwei:
In the export.py, if we just save the `model.state_dict`, we don't need to execute `convert_scaled_to_non_scaled(model, inplace=True)`.

csukuangfj:
> In the export.py, if we just save the `model.state_dict`, we don't need to execute `convert_scaled_to_non_scaled(model, inplace=True)`.

Yes, you are right.

csukuangfj:
Closing via #571

PR #522, opened 2022-08-05
Sort results to make it more convenient to compare decoding results

pkufool:
After sorting we can compare the decoding results sequence by sequence, that would be easier to find patterns.

image

PR #533, opened 2022-08-17
Fix not enough values to unpack error .

TeaPoly:
The bug was introduced by this PR https://github.com/k2-fsa/icefall/pull/522


pkufool:
Thanks!

PR #546, opened 2022-08-24
[Draft] Support streaming encoder onnx export

yuekaizhang:
I warpped the encoder.streaming_forward method with a new class StreaminEncoder. However, I am not sure if it is necessary to export it into onnx.


csukuangfj:
Thanks! I think it would be nice to have this class dedicated to streaming. It simplifies the calling code from the C++ side.

yuekaizhang:
I am going to use this streaming_encoder.onnx for triton streaming service. Would be back this PR after that.

PR #551, opened 2022-08-25
modified_beam_search with decoder output cached.

kobenaxie:
At each decoding step, if one hyp emits a `blank`, then the output embedding of predictor network can be cached for next step decoding.


csukuangfj:
Thanks! 

Do you have any numbers to show how the cache affects the decoding speed?

PR #552, opened 2022-08-27
minor changes for correct path names && import module text2segments.py

rickychanhoyin:
minor changes for correct path names && import module text2segments.py (in alimeeting recipe)


csukuangfj:
Thanks!

PR #560, opened 2022-08-31
Clip rnn gradients in a chunk-wise manner

yaozengwei:
This PR aim to clip the rnn gradients in a chunk-wise manner, to solve the gradient explosion problem in the backward pass. 
When computing each chunk, we clip the gradients of hidden states and cell states which are passed between chunks.

The gradient clipping strategy applied on hidden states and cell states is as follows:
1. If the gradient norm is lager than a specific threshould, we directly zero the gradients.
2. Scaling down the gradient by a factor of 0.9.
3. Limit the gradient norm to a maximum.


egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
```diff
@@ -165,7 +165,7 @@ def __init__(
         grad_scale_factor: float = 0.9,
```
danpovey, line 165: If we are applying this on batches of 10 frames, then probably this default for grad_scale_factor is too large, should be e.g. 0.5.

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
```diff
@@ -1016,19 +1016,19 @@ def _test_scaled_lstm_forward():
     assert torch.allclose(c, c_chunk)
 
 
-def _test_grad_cliper():
-    grad_cliper = GradientCliper(
+def _test_grad_clipper():
+    grad_clipper = GradientClipper(
         grad_norm_threshold=10.0, grad_scale_factor=0.9, grad_max_norm=1.0
```
danpovey, line 1021: I actually think the threshold thing may not end up being helpful... I don't know if I ever tested it independently from the max_norm. I think it will be very rare that it goes above the grad_norm_threshold.

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
```diff
@@ -111,6 +111,77 @@ def backward(
...
 
 
+class GradientClipFunction(torch.autograd.Function):
+    @staticmethod
+    def forward(
+        ctx,
+        x: Tensor,
+        grad_norm_threshold: float,  # e.g., 10.0
+        grad_scale_factor: float,  # e.g., 0.9
+        grad_max_norm: float,  # e.g., 5.0
```
danpovey, line 121: should have a `channel_dim: int = -1` parameter here.

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
```diff
@@ -111,6 +111,77 @@ def backward(
...
+    def backward(ctx, x_grad: Tensor) -> Tuple[Tensor, None]:
+        if x_grad.norm() > ctx.grad_norm_threshold:
+            # Zero gradients if its norm is larger than the threshold
+            x_grad = torch.zeros_like(x_grad)
+        else:
+            # Scale down gradient
+            x_grad = x_grad * ctx.grad_scale_factor
+
+            # Limit gradient norm to a maxmimum value
+            clip_coef = ctx.grad_max_norm / (x_grad.norm() + 1e-6)
```
danpovey, line 139: here, instead of x_grad.norm(), need `x_grad.norm(dim=ctx.channel_dim)`.
Also, cannot apply grad_norm_threshold using an if statement like this, would need to apply it as some kind
of multiplicative mask.  But it might be easier to just not apply it for now.

yaozengwei:
Now I am running the experiements with following options:
* `--rnn-clip-grad 1 --rnn-chunk-size 20 --rnn-grad-scale-factor 1.0 --rnn-grad-max-norm 0.5`
* `--rnn-clip-grad 1 --rnn-chunk-size 20 --rnn-grad-scale-factor 1.0 --rnn-grad-max-norm 1.0`
* `--rnn-clip-grad 1 --rnn-chunk-size 20 --rnn-grad-scale-factor 1.0 --rnn-grad-max-norm 2.0`

PR #564, opened 2022-09-05
Gradient filter for training lstm model

yaozengwei:
This PR implements a gradient filter module, which is used to filter out the sequence elements in batch that have undergone gradient explosion during backward pass. 
Specifically, for each sequence element in batch, its gradient will be filtered out if the gradient norm is larger than`threshold * median`, where `median` is the median value of gradient norms of all elements in batch, and `threshold` is a configurable parameter. 

The gradient filter module is applied on each lstm module's input, i.e., `lstm(gradient_filter(src))`. Hence if the lstm layer explodes, the created huge gradients will be filtered out and not propagate to its input.  

Compared to the chunk-wise gradient clipping https://github.com/k2-fsa/icefall/pull/560, this strategy saves training time since it does not need to perform chunk-wise computation.


egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
```diff
@@ -111,6 +111,74 @@ def backward(
...
+                batch_dim += x.ndim
+            ctx.batch_dim = batch_dim
+            ctx.threshold = threshold
+        return x, *params
+
+    @staticmethod
+    def backward(
+        ctx,
+        x_grad: Tensor,
+        *param_grads: nn.Parameter,
```
danpovey, line 134: these will be Tensor, not nn.Parameter.

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
```diff
@@ -111,6 +111,74 @@ def backward(
...
 
 
+class GradientFilterFunction(torch.autograd.Function):
+    @staticmethod
+    def forward(
+        ctx,
+        x: Tensor,
+        batch_dim: int,  # e.g., 1
+        threshold: float,  # e.g., 10.0
+        *params: nn.Parameter,  # module parameters
```
danpovey, line 121: better to say Tensor here, because the scaled parameters are actually Tensor not Parameter, and anyway Parameter is a subclass of Tensor

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
```diff
@@ -111,6 +111,74 @@ def backward(
...
+        *param_grads: nn.Parameter,
+    ):
+        dim = ctx.batch_dim
+        norm_dims = [d for d in range(x_grad.ndim) if d != dim]
+        norm_of_batch = (x_grad ** 2).mean(dim=norm_dims, keepdim=True).sqrt()
+        median_norm = norm_of_batch.median()
+        mask = norm_of_batch <= median_norm * ctx.threshold
+        x_grad = x_grad * mask
+
+        if not torch.all(mask):
```
danpovey, line 143: it would probably be more efficient, if we are using GPU to not have a conditional here, and to use multiplication, so we don't have to sync the GPU.
Could do:
```
   all = torch.all(mask).to(x_grad.dtype)  # 1 if no grad was zeroed, 0 if any was zeroed
   param_grads = [ all * x for x in param_grads ]
```
... OK to use this code if on CPU though.

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
```diff
@@ -111,6 +111,74 @@ def backward(
...
+            # If any of elements of x_grad is zeroed,
+            # the param_grads would be fully zeroed.
+            for g in param_grads:
```
danpovey, line 146: in this case it is probably fine, but in general we are probably not supposed to use zero_() or other in-place operations, because the grad tensor could potentially be shared with some other tensor that will be used in future.  more safe to do: g = torch.zeros_like(g), for instance.

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
```diff
@@ -111,6 +111,74 @@ def backward(
...
+        value of gradient norms of all elememts in batch.
+    """
+
+    def __init__(self, batch_dim: int = 1, threshold: float = 10.0):
+        super(GradientFilter, self).__init__()
+        self.batch_dim = batch_dim
+        self.threshold = threshold
+
+    def forward(self, x: Tensor, *params: nn.Parameter) -> Tensor:
+        if torch.jit.is_scripting() or is_jit_tracing():
```
danpovey, line 171: fix return type.  can just be Tuple[Tensor] probably, but not sure if typing expets that to be just one tensor.

egs/librispeech/ASR/lstm_transducer_stateless2/export.py
```diff
@@ -0,0 +1,388 @@
...
+
+# This script converts several saved checkpoints
+# to a single one using model averaging.
+"""
+
+Usage:
+
+(1) Export to torchscript model using torch.jit.trace()
+
+./lstm_transducer_stateless/export.py \
```
csukuangfj, line 27:
```suggestion
./lstm_transducer_stateless2/export.py \
```

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
```diff
@@ -513,10 +586,14 @@ def forward(
...
         self.check_forward_args(input, hx, None)
+
+        flat_weights = self._get_flat_weights()
+        input, *flat_weights = self.grad_filter(input, *flat_weights)
+
         result = _VF.lstm(
             input,
             hx,
-            self._get_flat_weights(),
+            flat_weights,
```
danpovey, line 596: should probably be including self.bias with the flat_weights.

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
danpovey: ... also the type would probably be Tuple[nn.Parameter], but I'm not 100% sure, you could perhaps just omit the typing information (or check).

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
csukuangfj: Also, `params` is of type `tuple`.

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
csukuangfj: I am afraid
```python3
g = torch.zeros_like(g)
```
will not change `param_grads` since it changes the temporary `g`.

csukuangfj (changes requested):

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
yaozengwei: OK i see.

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
yaozengwei: `_get_flat_weights()` returns all parameters including weights and bias.

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
yaozengwei: Referring to https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values, the type of `*param_grads` is `Tensor` instead of `Tuple[Tensor]`.

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
danpovey: yes, would need different code construct, like param_grads = [somethng for x in param_grads]

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
danpovey: ah, bias is a bool, I guess

egs/librispeech/ASR/pruned_transducer_stateless2/scaling.py
yaozengwei: Finished it in the last commit https://github.com/k2-fsa/icefall/pull/564/commits/e71b541a21985e3ad2646f9215a7e9bda2536a91.

yaozengwei:
The experimental results training on full librispeech manifest the advantage of using the gradient filter inside the lstm modules over the baseline:

| epoch-40-avg-15 | greedy search | modified beam search | fast beam search |
|-------------------|----------------|----------------------|------------------|
| baseline | 3.79 & 9.71 | 3.66 & 9.43 | 3.73 & 9.6 |
| with the gradient filter | 3.66 & 9.51 | 3.55 & 9.28 | 3.55 & 9.33 |

yaozengwei:
Now the gradient filter is applied on the input tensor and parameters inside each lstm module, as `input, *flat_weights = self.grad_filter(input, *flat_weights)`. 

The gradient filter module works as follows:

```
norm_of_batch = (x_grad ** 2).mean(dim=norm_dims, keepdim=True).sqrt()
median_norm = norm_of_batch.median()

cutoff = median_norm * ctx.threshold                                                                                                                                                                      
inv_mask = (cutoff + norm_of_batch) / (cutoff + eps)                                                                                                                                                      
mask = 1.0 / (inv_mask + eps)                                                                                                                                                                             
x_grad = x_grad * mask

avg_mask = 1.0 / (inv_mask.mean() + eps)
param_grads = [avg_mask * g for g in param_grads]
```

egs/librispeech/ASR/RESULTS.md
```diff
@@ -18,9 +106,10 @@ The WERs are:
 | modified_beam_search                | 2.75       | 7.08       | --iter 472000 --avg 18  |
 | fast_beam_search                    | 2.77       | 7.29       | --iter 472000 --avg 18  |
 
+
 The training command is:
 
-```bash
+```
```
csukuangfj, line 112: Please add `bash` here.

csukuangfj (approved):

yaozengwei:
This PR could be merged first. I'll make a new PR to support exporting to ncnn format via PNNX as in https://github.com/k2-fsa/icefall/pull/571 for this recipe later.

PR #565, opened 2022-09-05
Add gradient filter for tdnn_lstm_ctc 

yaozengwei:
This PR add the gradient filter for tdnn_lstm_ctc recipe. You could see https://github.com/k2-fsa/icefall/pull/564 for details.


danpovey:
@huangruizhe you can see whether this resolves your problem.

danpovey:
It will be hard to diagnose what's really going on here without looking at the diagnostics files (obtained by starting from intermediate epochs and adding the flag --print-diagnostics=True.. should take about 5 minutes).

yaozengwei:
This recipe does not support using flag `--print-diagnostics=True`.

danpovey:
Ruizhe can figure out how to add the code from other recipes, and make a PR.

PR #571, opened 2022-09-19
support exporting to ncnn format via PNNX

csukuangfj:
You can find the converted models at

https://huggingface.co/csukuangfj/sherpa-ncnn-2022-09-05

---

See https://github.com/k2-fsa/sherpa-ncnn for how to use them.
`sherpa-ncnn` uses [ncnn](https://github.com/tencent/ncnn) for neural network model computation and everything can be compiled into a single binary with static linking.
It has been tested on  Linux/macOS/Windows/Raspberry Pi.

(Note: `sherpa-ncnn` is self contained and does not depend other libraries, such as libtorch).


egs/librispeech/ASR/lstm_transducer_stateless/lstm.py
```diff
@@ -209,7 +218,13 @@ def forward(
         # lengths = ((x_lens - 3) // 2 - 1) // 2 # issue an warning
         #
         # Note: rounding_mode in torch.div() is available only in torch >= 1.8.0
-        lengths = (((x_lens - 3) >> 1) - 1) >> 1
+        if not self.is_pnnx:
+            lengths = (((x_lens - 3) >> 1) - 1) >> 1
```
danpovey, line 222: so // (integer division) not supported either in ncnn?

csukuangfj: No, ncnn's binary op only supports floating point operation.
Also, `>>` the right shift operator is not supported either in ncnn.
(Note: we can extend ncnn to support `>>` if there is a need.)

csukuangfj:
Will add doc and CI test tomorrow.

PR #593, opened 2022-09-29
[WIP] RNN-T + MBR training.

pkufool:
This PR depends on https://github.com/k2-fsa/k2/pull/1057 in k2.


pkufool:
The model structure is like the diagram below,  it has two joiners, one is the joiner for regular RNN-T, the other is `quasi-joiner` that produces the expected wer. To make the quasi-joiner work well, we use an `Enhanced embedding` instead of the `Encoder output`.  The `Embedding enhancer` is some kind of model that has self-attention from `masked_encoder_output` and cross-attention from `text_embedding` produced by a tranformer LM.

image

egs/librispeech/ASR/pruned_transducer_stateless_mbr/model.py
```diff
@@ -0,0 +1,963 @@
...
+          warmup:
+            A floating point value that gradually increases from 0 throughout
+            training; when it is >= 1.0 we are "fully warmed up".  It is used
+            to turn modules on sequentially.
+        """
+
+        N, T, C = embedding.shape
+        mask = torch.randn((N, T, C), device=embedding.device)
+        mask = mask > mask_proportion
+        masked_embedding = torch.masked_fill(embedding, ~mask, 0.0)
```
pkufool, line 947: I randomly mask the `encoder output` here.

Issue #603, opened 2022-10-06
Error while training pruned_transducer_stateless2


ngoel17:
While training, I am getting the following error. If I re-start the training from an earlier checkpoint, it proceeds to several next checkpoints before stopping like this again. The purpose of saving the batch I think is for debugging purposes, but I am not sure what to look for at this stage. Please advise. The recipe is adapted from librispeech for 8K data.

2022-10-05 23:42:56,919 INFO [train.py:814] (3/4) Epoch 0, batch 25600, loss[loss=0.2133, simple_loss=0.2315, pruned_loss=0.09752, over 4663.00 frames. utt_duration=3110 frames, utt_pad_proportion=0.009079, over 6.00 utterances.], tot_loss[loss=0.312, simple_loss=0.33, pruned_loss=0.1471, over 932672.09 frames. utt_duration=1056 frames, utt_pad_proportion=0.06028, over 3537.38 utterances.], batch size: 6, lr: 1.09e-03
2022-10-05 23:43:02,341 INFO [utils.py:968] (1/4) Saving batch to pruned_transducer_stateless2/exp_multi/batch-bdd640fb-0667-1ad1-1c80-317fa3b1799d.pt
2022-10-05 23:43:02,391 INFO [utils.py:974] (1/4) features shape: torch.Size([202, 131, 80])
2022-10-05 23:43:02,392 INFO [utils.py:978] (1/4) num tokens: 564
Traceback (most recent call last):
  File "/mnt/dsk1/icefall/egs/8k/./pruned_transducer_stateless2/train.py", line 1089, in <module>
    main()
  File "/mnt/dsk1/icefall/egs/8k/./pruned_transducer_stateless2/train.py", line 1080, in main
    mp.spawn(run, args=(world_size, args), nprocs=world_size, join=True)
  File "/home/ngoel/anaconda3/envs/k2/lib/python3.9/site-packages/torch/multiprocessing/spawn.py", line 240, in spawn
    return start_processes(fn, args, nprocs, join, daemon, start_method='spawn')
  File "/home/ngoel/anaconda3/envs/k2/lib/python3.9/site-packages/torch/multiprocessing/spawn.py", line 198, in start_processes
    while not context.join():
  File "/home/ngoel/anaconda3/envs/k2/lib/python3.9/site-packages/torch/multiprocessing/spawn.py", line 160, in join
    raise ProcessRaisedException(msg, error_index, failed_process.pid)
[...]

PR #604, opened 2022-10-07
Filter utterances with number_tokens > number_feature_frames.

csukuangfj:
Also, print warnings when an utterance is filtered out.

See also
- https://github.com/k2-fsa/icefall/issues/603#issuecomment-1270203039
- https://github.com/k2-fsa/icefall/issues/603#issuecomment-1270491321

@ngoel17
Could you try this PR?

### DIFF
--- a/.github/workflows/style_check.yml
+++ b/.github/workflows/style_check.yml
@@ -29,8 +29,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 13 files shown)


csukuangfj:
Note that only the librispeech dataset is updated. Would be great if someone can help propagate the changes to other datasets.

ngoel17:
Looks great. testing it out.

Issue #132, opened 2021-11-26
Exception when building HLG on large language model


drawfish:
It seems that overflow occurs when building HLG on large language model.
The apra language model info:
```
\data\
ngram 1=136920
ngram 2=30620323
ngram 3=58191282
ngram 4=63623739
```
After successfully detrminize(compose(L,G)) , an exception occurs when remove_epsilon from LG.
From the exception traceback we can see that something like overflow when cating impl_->byte_offset into int64_t ? 
```
2021-11-25 16:23:09,208 INFO [compile_hlg.py:173] Processing data/lang_phone
2021-11-25 16:23:09,479 INFO [lexicon.py:177] Loading pre-compiled data/lang_phone/Linv.pt
2021-11-25 16:23:09,900 INFO [compile_hlg.py:79] Building ctc_topo. max_token_id: 218
2021-11-25 16:23:09,911 INFO [compile_hlg.py:86] Loading L_disambig.fst.txt
2021-11-25 16:23:12,274 INFO [compile_hlg.py:91] Loading G.fst.txt
2021-11-25 16:34:12,174 INFO [compile_hlg.py:110] Intersecting L and G
2021-11-25 17:04:55,417 INFO [compile_hlg.py:112] LG shape: (552371407, None)
2021-11-25 17:04:55,417 INFO [compile_hlg.py:114] Connecting LG
2021-11-25 17:04:55,417 INFO [compile_hlg.py:116] LG shape after k2.connect: (552371407, None)
2021-11-25 17:04:55,418 INFO [compile_hlg.py:118] <class 'torch.Tensor'>
2021-11-25 17:04:55,418 INFO [compile_hlg.py:119] Determinizing LG
2021-11-25 17:51:51,788 INFO [compile_hlg.py:122] <class '_k2.ragged.RaggedTensor'>
2021-11-25 17:51:51,788 INFO [compile_hlg.py:124] Connecting LG after k2.determinize
2021-11-25 17:51:51,788 INFO [compile_hlg.py:127] Removing disambiguation symbols on LG
[F]/k2/k2/k2/csrc/tensor.cu:159:k2::Tensor::Tensor(k2::Dtype, const k2::Shape&, k2::RegionPtr, int32_t) Check failed: int64_t(impl_->byte_offset) + begin_elem * element_size >= 0 (-156
3030948 vs. 0) impl_->byte_offset: -1563030948, begin_elem: 0, element_size: 4


[ Stack-Trace: ]
[...]

Issue #23, opened 2021-08-23
Problem during preparing G


EmreOzkose:
Hi, I am trying to train a tdnn-lstm model on a Turkish data. I run an experiment successfully once (including decoding part) with smaller langauge model. Then I tried a new corpus for language modeling. During `preparing G` step, I got this error:

```
2021-08-23 11:35:45 (prepare.sh:118:main) Stage 6: Compile HLG
2021-08-23 11:35:46,389 INFO [compile_hlg.py:126] Processing data/lang_phone
2021-08-23 11:35:46,859 INFO [lexicon.py:99] Converting L.pt to Linv.pt
2021-08-23 11:35:47,640 INFO [compile_hlg.py:48] Building ctc_topo. max_token_id: 52
2021-08-23 11:35:47,826 INFO [compile_hlg.py:57] Loading G_3_gram.fst.txt
2021-08-23 11:38:15,955 INFO [compile_hlg.py:68] Intersecting L and G
2021-08-23 11:51:37,889 INFO [compile_hlg.py:70] LG shape: (301909252, None)
2021-08-23 11:51:37,889 INFO [compile_hlg.py:72] Connecting LG
2021-08-23 11:51:37,889 INFO [compile_hlg.py:74] LG shape after k2.connect: (301909252, None)
2021-08-23 11:51:37,889 INFO [compile_hlg.py:76] <class 'torch.Tensor'>
2021-08-23 11:51:37,889 INFO [compile_hlg.py:77] Determinizing LG
2021-08-23 12:09:11,585 INFO [compile_hlg.py:80] <class '_k2.RaggedInt'>
2021-08-23 12:09:11,585 INFO [compile_hlg.py:82] Connecting LG after k2.determinize
2021-08-23 12:09:11,585 INFO [compile_hlg.py:85] Removing disambiguation symbols on LG
[F] /usr/share/miniconda/envs/k2/conda-bld/k2_1628135473078/work/k2/csrc/tensor.cu:159:k2::Tensor::Tensor(k2::Dtype, const k2::Shape&, k2::RegionPtr, int32_t) Check failed: int64_t(impl_->byte_offset) + begin_elem * element_size >= 0 (-1246502780 vs. 0) 


[ Stack-Trace: ]
/path/to/miniconda3/envs/k2/lib/python3.8/site-packages/libk2_log.so(k2::internal::GetStackTrace()+0x4c) [0x7fb12e7c76bc]
[...]

PR #606, opened 2022-10-08
Support using OpenFst to compile HLG.

csukuangfj:
It is to avoid OOM in determinizing LG when G is very large.

See also
- #132 
- https://github.com/k2-fsa/k2/issues/1010
- https://github.com/k2-fsa/k2/issues/730
- https://github.com/k2-fsa/icefall/issues/23

@Jarvan-Wang  @wwx007121  Could you help test this PR? It should be able to handle large LMs and won't throw OOM as long as it does not throw OOM in kaldi as it is also using OpenFst to build LG.

---

I have tested the generated HLG with the pretrained model from
https://huggingface.co/Zengwei/icefall-asr-librispeech-lstm-transducer-stateless3-2022-09-28

It produces identical WER as the HLG generated by the master code for the first 30 utterances of test-clean and test-other.

One thing to note is that the file size of the resulting HLG is much smaller.
```bash
-rw-r--r-- 1 kuangfangjun root 806M Oct  8 10:51 data/lang_bpe_500/HLG.pt
-rw-r--r-- 1 kuangfangjun root 583M Oct  8 21:03 data/lang_bpe_500/HLG_fst.pt
```

HLG.pt is generated by the master code, while HLG_fst.pt is generated by this PR.

You have to install [kaldifst](https://github.com/k2-fsa/kaldifst) to use this PR.
```
pip install kaldifst
```
or
```
conda install -c kaldifst kaldifst
```

The documentation for `kaldifst` is available at
https://k2-fsa.github.io/kaldifst/


wangtiance:
Glad that I found this. Would it further help reducing the size of HLG if I do determinize and minimize after composing H with LG?

wwx007121:
> Glad that I found this. Would it further help reducing the size of HLG if I do determinize and minimize after composing H with LG?

This is caused by the size of the FST itself, and the smaller the FST will affect the recognition effect. So It is better to use Lattice instead of H to be composed with LG, if you want to decode with phone-level modeling/wfst

wangtiance:
@wwx007121 I thought after minimizing the FST is equivalent as before, so it shouldn't affect recognition result?

Issue #610, opened 2022-10-10
Speed of ActivationBalancer


danpovey:
Guys,
I discovered an issue with the speec of the ActivationBalancer.  After implementing NVTX profiling
by using `import nvtx` and then using `nvtx.annotate` as both decorators and context managers, I
found that the ActivationBalancer is taking up quite a bit of time.
I reduced this quite a bit by changing some code in scaling.py to read like this:
```
        balance_prob = 0.25
        if random.random() < balance_prob:
            return ActivationBalancerFunction.apply(
                x,
                self.channel_dim,
                self.min_positive,
                self.max_positive,
                self.max_factor / balance_prob,
                self.min_abs,
                self.max_abs,
            )
        else:
            return x
```
and after running an experiment, it seems not to affect convergence at all.  
The screenshot is from a setup with some other unrelated changes.  But I think this small change is pretty safe, perhaps
someone could make a PR to apply it in our recipes?  Notice I also divide max_factor by balance_prob.

Something else: the internal code of ActivationBalancer has a few elementwise expressions that don't need gradient, and which could be speeded up by writing a kernel.  Does anyone know of a way to have pytorch generate just-in-time kernels for such things?

<img width="901" alt="Screen Shot 2022-10-10 at 10 49 04 AM" src="https://user-images.githubusercontent.com/3298747/194793020-5d320bce-0943-4a96-a334-ed661392e32e.png">

yaozengwei:
I will make a PR to apply the above change of ActivationBalancer in our recipes.

pzelasko:
> Something else: the internal code of ActivationBalancer has a few elementwise expressions that don't need gradient, and which could be speeded up by writing a kernel. Does anyone know of a way to have pytorch generate just-in-time kernels for such things?

If you don't want to write a full blown JIT extension (see https://pytorch.org/tutorials/advanced/cpp_extension.html#jit-compiling-extensions), perhaps using CuPy is a fairly straightforward way of achieving this: https://docs.cupy.dev/en/stable/user_guide/interoperability.html#using-custom-kernels-in-pytorch

danpovey:
That's interesting.
Have you used CuPy?  I'm wondering how easy the installation process is.
(Things that have a dependency on the CUDA toolkit tend to be tricky to install.)

danpovey:
> I will make a PR to apply the above change of ActivationBalancer in our recipes.

[...]

PR #612, opened 2022-10-10
Modify ActivationBalancer for speed

yaozengwei:
This PR applies the ActivationBalancer module with a probability (e.g., 0.25) in the reworked model, to save training time. See https://github.com/k2-fsa/icefall/issues/610 for details.


yaozengwei:
The results training on train-clean-100 with 2 jobs shows that applying the ActivationBalancer module with a probability of 0.25 gets similar results but saves about 1.5 minutes per epoch. 

Decode with greedy search at epoch-20-avg-5: 
* baseline, 6.97 & 18.64, 25~26 minutes / epoch
* probability=0.25, 7.08 & 18.5, 24 minutes / epoch

danpovey:
OK cool, I suppose this can be merged.

PR #420, opened 2022-06-14
Add fast_beam_search_nbest.

csukuangfj:

PR #622, opened 2022-10-17
Add fast_beam_search_LG

pkufool:
In our previous experiments, we found using `log_add` can get a better result comparing with using `max` when decoding with a general LG graph, #420 gives a detaily analysis.   

Recently, I found that using `max` can also get good results after adding `allow_partial` (default False) option in https://github.com/k2-fsa/k2/pull/1059 

Here is some results of `pruned_transducer_stateless2` (model was downloaded from https://huggingface.co/csukuangfj/icefall-asr-librispeech-pruned-transducer-stateless2-2022-04-29)

| decoding method | before fixing `allow_partial` | after fixing `allow_partial` |
| ---------------------| -------------------- | ------------------ |
| fast_beam_search_LG |    6.53 & 10.21 |  2.84 & 6.08 |
| fast_beam_search_nbest_LG | 2.84 & 6.0 |  2.84 & 5.98 |

**Note:** `fast_beam_search_LG` is twice as faster as `fast_beam_search_nbest_LG`.

### DIFF
--- a/.github/scripts/run-librispeech-pruned-transducer-stateless2-2022-04-29.sh
+++ b/.github/scripts/run-librispeech-pruned-transducer-stateless2-2022-04-29.sh
@@ -83,4 +83,5 @@ if [[ x"${GITHUB_EVENT_NAME}" == x"schedule" || x"${GITHUB_EVENT_LABEL_NAME}" ==
   done
 
   rm pruned_transducer_stateless2/exp/*.pt
+  rm -r data/*
 fi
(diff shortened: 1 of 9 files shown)


pkufool:
The issue is previous version `rnnt_decoding` (before fixing `allow_partial`) did something wrong with final state, it treated all the graph states on the final acoustic frame as final state in the generated lattice, which encourages shorter symbol sequences. So, in our previous experiments, we saw a lot of deletion errors at the tail of sentences when decoding with `max`.

francisr:
I'm using the LG decode, and I regularly get lattices with no path going to the final state, did you observe that?

pkufool:
> I'm using the LG decode, and I regularly get lattices with no path going to the final state, did you observe that?

Can you try adding  `allow_partial=True` to the `format_output` in this line: https://github.com/k2-fsa/icefall/blob/c30b8d3a1c095671d89321323390f8858722bcd1/egs/librispeech/ASR/pruned_transducer_stateless2/beam_search.py#L465

francisr:
Yes that does work fine, however I was wondering if other people have had the same issue as me when allow_partial is false.

pkufool:
> Yes that does work fine, however I was wondering if other people have had the same issue as me when allow_partial is false.

Could be, the allow_partial option was merged recently. Will revisit it, thanks!

PR #649, opened 2022-11-02
Update README.md

JinZr:
Replace two inaccessible urls in README.md to align with updated TDNN-LSTM-CTC recipe for Librispeech and Aishell.


csukuangfj:
Thanks

PR #662, opened 2022-11-07
Apply delay penalty on pytorch ctc loss

yaozengwei:
This PR supports to apply delay penalty on pytorch-based ctc loss, to reduce the symbol delay.  The current model uses only ctc loss, without attention decoder head. 

We want to add `lambda * torch.arange(T)` to the ctc blank probs (after softmax), to encourage blanks to align to later and encourage non-blank symbols to align earlier. However, it would also encourage blank over non-blanks. This is not what we want. This effect will be stronger for longer utterances, where the penalty scores `lambda * torch.arange(T)` get too large.

To minimize the above "repeat-penalizing effect", we first split utterances into shorter sub-utterances, for purpose of adding the "arange" expression. We can declare that there is a sub-utterance boundary every time the blank prob is >= 0.99, for instance. Specifically, we 
(1) compute a mask that is 1 when blank_prob >= 0.99 (or t==0);
(2) compute cummax_out = torch.cummax(torch.arange(T) * mask)
(3) compute the sawtooth-like "blank-bonus" values: penalty = torch.arange(T) - cummax_out
(4) add `penalty * lambda` to ctc blank probs (after softmax). 

See `egs/librispeech/ASR/conformer_ctc3/model.py` for details.


yaozengwei:
Note: The delay penalty is applied when warmup >= 1.0. Otherwise the loss would not converge.

PR #668, opened 2022-11-08
Fix exception in find_checkpoints

abb128:
I manually renamed a checkpoint to something like `checkpoint-476000-1.pt` and this later caused a sudden exception and training to stop. This PR fixes the exception and instead logs and ignores invalid checkpoint filenames
image


csukuangfj:
@abb128 
Thanks for your contribution!

PR #669, opened 2022-11-09
Apply delay penalty on k2 ctc loss

yaozengwei:
This PR supports to apply delay penalty on k2 ctc loss, to reduce the symbol delay. The encoder is streaming conformer model. 
We provide two methods of delay penalty here: 
(1) Add penalty scores on encoder ctc outputs after log-softmax. Use param `--nnet-delay-penalty` in `train.py`. 
(2) Add penalty scores on ctc decoding lattice. Use param `--delay-penalty` in `train.py`. It requires https://github.com/k2-fsa/k2/pull/1086 in k2. 

For method (1), we want to add lambda * torch.arange(T) to the ctc blank probs (after log-softmax), to encourage blanks to align to later and encourage non-blank symbols to align earlier. However, it would also encourage blank over non-blanks. This is not what we want. This effect will be stronger for longer utterances, where the penalty scores lambda * torch.arange(T) get too large. To minimize the above "repeat-penalizing effect", we first split utterances into shorter sub-utterances, for purpose of adding the "arange" expression. We can declare that there is a sub-utterance boundary every time the blank prob is >= 0.99, for instance. Specifically, we
(1) compute a mask that is 1 when blank_prob >= 0.99 (or t==0);
(2) compute cummax_out = torch.cummax(torch.arange(T) * mask)
(3) compute the sawtooth-like "blank-bonus" values: penalty = torch.arange(T) - cummax_out
(4) add penalty * lambda to ctc blank probs (after log-softmax).

For method (2), we add penalty scores (frame offset relative to the middle frame, scaled by lambda) on the non-repeat and non-blank arcs in ctc decoding lattice, which are at the frames when the non-blank symbols are firstly emitted. It is the same as we did in transducer loss (see https://arxiv.org/pdf/2211.00490.pdf).

- Note: The delay penalty is applied when warmup >= 1.0. Otherwise the loss would not converge.


yaozengwei:
I train the models on full-librispeech for 25 epochs, and decode with `ctc-decoding` method. The following table presents the experimental results of two penalty methods: (1) Add "sawtooth" on blank log-prob; (2) Add frame offsets on Non-blank & non-repeat arcs. It shows that the second method achieves better WER-Delay trade-off than the first method. 

Method | WER (%) | Mean symbol delay (ms) | Comment
-- | -- | -- | --
Baseline | 3.9 & 10.19 | 0.325s & 0.321s | epoch-25-avg-5
"sawtooth" on blank log-prob, lambda=0.01 | 4.18 & 10.68 | 0.167s & 0.168s | epoch-25-avg-5
"sawtooth" on blank log-prob, lambda=0.02 | 4.65 & 11.66 | 0.076s & 0.08s | epoch-25-avg-5
"sawtooth" on blank log-prob, lambda=0.03 | 4.69 & 11.96 | 0.006s & 0.012s | epoch-25-avg-5
Non-blank & non-repeat arcs, lambda=0.01 | 4.06 & 10.3 | 0.164s & 0.168s | epoch-25-avg-6
Non-blank & non-repeat arcs, lambda=0.02 | 4.21 & 10.73 | -0.056s & -0.03s | epoch-25-avg-7
Non-blank & non-repeat arcs, lambda=0.03 | 4.65 & 11.51 | -0.298s & -0.269s | epoch-25-avg-8

</div>

PR #670, opened 2022-11-10
fix mask errors when padding audios

yuekaizhang:
Here https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/pruned_transducer_stateless2/conformer.py#L159, we assert `x.size(0) == lengths.max().item() `. It's correct during training time. However, when I tried to send multiple audios from different clients, to dynamic batch the requests, I padded all audios to the nearest 5 before sending them. Then it would cause an error if x.size(0) > mask.size(-1) (for attention scores masking).

This PR fixed the error and we need to use mask = make_pad_mask(lengths, x.size(0)) when dynamic batch with onnx or jit.

### DIFF
--- a/icefall/utils.py
+++ b/icefall/utils.py
@@ -1017,7 +1017,7 @@ def add_eos(ragged: k2.RaggedTensor, eos_id: int) -> k2.RaggedTensor:
     return concat(ragged, eos_id, direction="right")
 
 
-def make_pad_mask(lengths: torch.Tensor,  max_len: int = 0) -> torch.Tensor:
+def make_pad_mask(lengths: torch.Tensor, max_len: int = 0) -> torch.Tensor:
     """
     Args:
       lengths:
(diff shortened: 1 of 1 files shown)


csukuangfj:
> . It's correct during training time. However, when I tried to send multiple audios from different clients, to dynamic batch the requests, I padded all audios to the nearest 5 before sending them. Then it would casue an error if x.size(0) > mask.length(-1) (for attention scores masking).

Could you give more details about why `x.size(0) == lengths.max().item()` does not hold?

yuekaizhang:
> > . It's correct during training time. However, when I tried to send multiple audios from different clients, to dynamic batch the requests, I padded all audios to the nearest 5 before sending them. Then it would casue an error if x.size(0) > mask.length(-1) (for attention scores masking).
> 
> Could you give more details about why `x.size(0) == lengths.max().item()` does not hold?

When all audios in a batch are padded, it can‘t hold. Specifically，triton dynamic batching requires all requests have the same seq len.

csukuangfj:
Thanks!

PR #686, opened 2022-11-16
Zipformer output length

desh2608:
Refer: https://github.com/k2-fsa/icefall/issues/603#issuecomment-1316136756

The output length is given as `T = ((N - 7) // 2 + 1) // 2`. For training, we should ensure that cuts are filtered to remove the ones for which S (number of BPE tokens) is smaller than T. This can be done using, for instance:

```python
def _remove_short_and_long_utt(c: Cut):
    if c.duration < 0.1 or c.duration > 25.0:
        return False

    # In pruned RNN-T, we require that T >= S
    # where T is the number of feature frames after subsampling
    # and S is the number of tokens in the utterance

    # In ./zipformer.py, the conv module uses the following expression
    # for subsampling
    T = ((c.num_frames - 7) // 2 + 1) // 2
    tokens = sp.encode(c.supervisions[0].text, out_type=str)
    return T >= len(tokens)
    
cuts = cuts.filter(_remove_short_and_long_utt)
```

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless7/zipformer.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless7/zipformer.py
@@ -1828,6 +1828,7 @@ def _test_zipformer_main():
         torch.randn(batch_size, seq_len, feature_dim),
         torch.full((batch_size,), seq_len, dtype=torch.int64),
     )
+    assert ((seq_len - 7) // 2 + 1) // 2 == f[0].shape[1], (seq_len, f.shape[1])
     f[0].sum().backward()
     c.eval()
     f = c(
--- a/egs/librispeech/ASR/local/filter_cuts.py
+++ b/egs/librispeech/ASR/local/filter_cuts.py
@@ -101,6 +101,9 @@ def filter_cuts(cut_set: CutSet, sp: spm.SentencePieceProcessor):
         # Note: for ./lstm_transducer_stateless/lstm.py, the formula is
         #  T = ((num_frames - 3) // 2 - 1) // 2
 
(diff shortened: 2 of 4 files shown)


egs/librispeech/ASR/pruned_transducer_stateless7/zipformer.py
```diff
@@ -1828,6 +1828,7 @@ def _test_zipformer_main():
         torch.randn(batch_size, seq_len, feature_dim),
         torch.full((batch_size,), seq_len, dtype=torch.int64),
     )
+    assert ((seq_len - 7) // 2 + 1) // 2 == f[0].shape[1], (seq_len, f.shape[1])
```
csukuangfj, line 1831: Could you update
https://github.com/k2-fsa/icefall/blob/855c76655b49de61a5c6d054a7ff2158a639e6f7/egs/librispeech/ASR/pruned_transducer_stateless7/train.py#L1037
and
https://github.com/k2-fsa/icefall/blob/855c76655b49de61a5c6d054a7ff2158a639e6f7/egs/librispeech/ASR/pruned_transducer_stateless8/train.py#L1048

You can copy the code from
https://github.com/k2-fsa/icefall/blob/855c76655b49de61a5c6d054a7ff2158a639e6f7/egs/librispeech/ASR/pruned_transducer_stateless5/train.py#L1037

desh2608: Done.

(It seems several style changes were done by the pre-commit hook.)

csukuangfj:
Thanks!

PR #687, opened 2022-11-16
Small code refactoring

videodanchik:
This is a small refactoring before the tedlium3 `conformer_ctc2` recipe. Also, extend `1best` decoding with LM weights scales.

### DIFF
--- a/icefall/decode.py
+++ b/icefall/decode.py
@@ -459,6 +459,7 @@ class Nbest(object):
 def one_best_decoding(
     lattice: k2.Fsa,
     use_double_scores: bool = True,
+    lm_scale_list: Optional[List[float]] = None,
 ) -> k2.Fsa:
     """Get the best path from a lattice.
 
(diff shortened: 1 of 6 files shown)


icefall/decode.py
```diff
@@ -468,11 +469,28 @@ def one_best_decoding(
...
+        for lm_scale in lm_scale_list:
+            am_scores = saved_am_scores / lm_scale
+            lattice.scores = am_scores + lattice.lm_scores
+
+            best_path = k2.shortest_path(
+                lattice, use_double_scores=use_double_scores
+            )
+            key = f"lm_scale_{lm_scale}"
+            ans[key] = best_path
+        return ans
```
csukuangfj, line 491: Could you also update the type hint of the return value of this function?

icefall/decode.py
```diff
@@ -459,6 +459,7 @@ def build_levenshtein_graphs(self) -> k2.Fsa:
 def one_best_decoding(
     lattice: k2.Fsa,
     use_double_scores: bool = True,
+    lm_scale_list: Optional[List[float]] = None,
```
csukuangfj, line 462: Is there an example decoding script that is using it?

We found today that it helps by tuning this value for HLG decoding.

videodanchik: Yes I have an example it is in my current tedlium3 `conformer_ctc2` recipe it is almost ready (give me one or two days, and I will do PR). I also discovered during my early work with k2/icefall that tuning this parameter affects WER dramatically during HLG decoding. My current best model achieves SOTA on tedlium3 with HLG decoding and tuning language model weight
image

icefall/decode.py
videodanchik: sure

videodanchik:
@csukuangfj I fixed the type hint.

csukuangfj:
Thanks!

PR #690, opened 2022-11-16
Apply new Black style changes

desh2608:
This PR applies the new line-length configuration (see https://github.com/lhotse-speech/lhotse/issues/890) on all the files. 

We have added a `.git-blame-ignore-revs` file containing the hash of the main commit to ensure that git blame history is not affected. Please follow the instructions in [this link](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html) to enable this setting in your git config. This option is available in git 2.23 and later (and also enabled by default on Github).

PR #692, opened 2022-11-17
Style change 2.0

desh2608:
This is a corrected version of https://github.com/k2-fsa/icefall/pull/690.

* (107df3b) First pass: We apply black formatting without the `--experimental-string-processing` feature which was causing some files to get deleted entirely.
* (d31db01) Second pass: I manually went through all the modifications and corrected several of the sort:
```python
            f"expected sample rate: {expected_sample_rate}. "
            f"Given: {sample_rate}"
```
being converted to:
```python
           f"expected sample rate: {expected_sample_rate}. " f"Given: {sample_rate}"
```
and a small number of cases where inline comments made the formatting weird.

At the very least, these changes should not cause anything to break. If there is something I missed, I am hoping they can be corrected as people come across them in the future.


desh2608:
@csukuangfj I would request you to please briefly check the files in the main `icefall/` directory since they are used in all recipes, to ensure there are no breaking changes. (You can skip the changes in `egs`)

.github/workflows/style_check.yml
```diff
@@ -45,17 +45,18 @@ jobs:
 
       - name: Install Python dependencies
         run: |
-          python3 -m pip install --upgrade pip black==21.6b0 flake8==3.9.2 click==8.0.4
-          # See https://github.com/psf/black/issues/2964
-          # The version of click should be selected from 8.0.0, 8.0.1, 8.0.2, 8.0.3, and 8.0.4
+          python3 -m pip install --upgrade pip black==22.3.0 flake8==5.0.4 click==8.1.0
```
csukuangfj, line 48: Could you also update
https://github.com/k2-fsa/icefall/blob/master/docs/source/contributing/code-style.rst

csukuangfj:
Thanks! Looks good to me. Left a minor comment.

.github/workflows/style_check.yml
desh2608: Done!

desh2608:
> Thanks! Looks good to me. Left a minor comment.

I think it's ready for merge?

Issue #535, opened 2022-08-17
Possible bug in <UNK> pronunciation.


videodanchik:
Hi, when I was adapting Librispeech for my needs, I haven't even noticed that I changed 
https://github.com/k2-fsa/icefall/blob/c74cec59e9f6d00e3a5838b4f8d4ace7e2303ad4/egs/librispeech/ASR/local/prepare_lang_bpe.py#L162
to:
`lexicon.append(("<UNK>", [▁, sp.id_to_piece(sp.unk_id())]))`
I just did it as part of my testing and experimenting, noticing that for unknown words `sentencepiece` outputs `▁ <unk>` I just added and forget (and everything seems to work well), and now I realized that it's in the main script for generating BPE lang. So is it a bug or not? Because if it is it seems not to affect recipes with a low number of `<UNK>` in supervision transcripts, but if there are a lot of `<UNK>` like for example in `tedlium3` recipe, there might be issues and I believe there were some...

csukuangfj:
We have done that before. Please have a look at 
https://github.com/k2-fsa/icefall/pull/183#issuecomment-1047583114

> Update some results (decoding-method=greedy_search) when removing <unk> for training bpe model:
when encoding <unk> into a single id, the best result is 6.73%.
when not encoding <unk> into a single id, the best result is 8.78%.
If I don't remove <unk> for training bpe model and don't encode <unk> into a single id, the best result is 11.11%
>
> The above results show that it is necessary to remove <unk> and encode it into a single id.

pkufool:
> We have done that before. Please have a look at [#183 (comment)](https://github.com/k2-fsa/icefall/pull/183#issuecomment-1047583114)
> 
> > Update some results (decoding-method=greedy_search) when removing  for training bpe model:
> > when encoding  into a single id, the best result is 6.73%.
> > when not encoding  into a single id, the best result is 8.78%.
> > If I don't remove  for training bpe model and don't encode  into a single id, the best result is 11.11%
> > The above results show that it is necessary to remove  and encode it into a single id.

I think what he was talking about is different from https://github.com/k2-fsa/icefall/pull/183#issuecomment-1047583114
. He converted`<UNK>` to [`▁`, `<UNK>`], but in https://github.com/k2-fsa/icefall/pull/183#issuecomment-1047583114, `<UNK>` was converted to [`<`, `U`, `N`, `K`, `>`]. 

[...]

PR #183, opened 2022-01-21
[WIP] Add Tedlium3 recipe

luomingshuang:
Will add RNN-T recipes first, and then others.

PR #696, opened 2022-11-19
Tedlium3 conformer ctc2

videodanchik:
This is tedlium3 `conformer_ctc2` recipe it is similar to `librispeech/ASR/conformer_ctc2` though I was mostly looking into `librispeech/ASR/pruned_transducer_stateless5` and therefore there are some differences comparing to `librispeech/ASR/conformer_ctc2`:

1. Warmup added to decoder layers.
2. `RandomCombiner` added to encoder.

I modified `prepare.sh` significantly though it must stay compatible with the previous `tedlium3` recipes. Now we download the pretrained 4-gram (small and big) language models from `http://kaldi-asr.org/models/5`; small is used for `HLG` decoding and big is used for rescoring. I also used `TEDLIUM_release-3/tedlium3/LM/*.en.gz` for training the BPE models and `TEDLIUM_release-3/tedlium3/TEDLIUM.152k.dic` as a lexicon. Also, an example of a decoding script with tuning LM weights for `1best` is presented.

This is not ready for merge yet, I'm still working on some experiments on different `<unk>` encodings (continuing https://github.com/k2-fsa/icefall/issues/535), for tedlium3 these `<unk>` seems to correspond to punctuation, so I want to play around with that. I also have some ideas how to modify the decoder, but it is later. For now, my best WERs for `HLG` and `attention-decoder` decoding methods are:

`1best`
dev WER lm_scale_0.1 6.55
test WER lm_scale_0.1 6.08

`attention-decoder`
dev WER ngram_lm_scale_0.01_attention_scale_0.7 6.33
test WER ngram_lm_scale_0.01_attention_scale_1.5 5.74


egs/librispeech/ASR/conformer_ctc/label_smoothing.py
```diff
@@ -45,6 +45,7 @@ def __init__(
         """
         super().__init__()
         assert 0.0 <= label_smoothing < 1.0
+        assert reduction in ("none", "sum", "mean")
```
csukuangfj, line 48:
```suggestion
        assert reduction in ("none", "sum", "mean"), reduction
```
It is helpful to print out the value of reduction on assertion failure.

egs/librispeech/ASR/local/prepare_lang_bpe.py
```diff
@@ -204,7 +204,9 @@ def main():
 
     words = word_sym_table.symbols
 
-    excluded = ["<eps>", "!SIL", "<SPOKEN_NOISE>", "<UNK>", "#0", "<s>", "</s>"]
+    excluded = [
+        "<eps>", "!SIL", "!sil", "<SPOKEN_NOISE>", "<UNK>", "<unk>" "#0", "<s>", "</s>"
```
csukuangfj, line 208:
```suggestion
        "<eps>", "!SIL", "!sil", "<SPOKEN_NOISE>", "<UNK>", "<unk>", "#0", "<s>", "</s>"
```

egs/tedlium3/ASR/conformer_ctc2/attention.py
```diff
@@ -0,0 +1,198 @@
...
+from scaling import ScaledLinear
+
+
+class MultiheadAttention(torch.nn.Module):
+    """Allows the model to jointly attend to information
+    from different representation subspaces. This is a modified
+    version of the original version of multihead attention 
+    (see Attention Is All You Need <https://arxiv.org/abs/1706.03762>)
+    with replacement of input / output projection layers 
+    with newly itroduced ScaleLinear layer
```
csukuangfj, line 30:
```suggestion
    with newly introduced ScaleLinear layer 
```

egs/tedlium3/ASR/conformer_ctc2/attention.py
```diff
@@ -0,0 +1,198 @@
...
+        
+        super().__init__()
+        
+        self.embed_dim = embed_dim
+        self.num_heads = num_heads
+        self.dropout = dropout
+        self.batch_first = batch_first
+        
+        if embed_dim % num_heads != 0:
+            raise ValueError("embed_dim must be divisible by num_heads")
```
csukuangfj, line 78: Shall we print out the value of `embed_dim` and `num_heads` on error, e.g.,

```python3
raise ValueError(f"embed_dim must be divisible by num_heads. {embed_dim} vs {num_heads}")
```

egs/tedlium3/ASR/conformer_ctc2/combiner.py
```diff
@@ -0,0 +1,248 @@
...
+               weight `final_weight` to the final layer and equal
+               weights to other layers; then add Gaussian noise
+               with variance `stddev` to these log-weights, and normalize
+               to weights (note: the average weight assigned to the
+               final layer here will not be `final_weight` if stddev>0).
+        """
+        super().__init__()
+        assert 0 <= pure_prob <= 1, pure_prob
+        assert 0 < final_weight < 1, final_weight
+        assert num_inputs >= 1
```
csukuangfj, line 73:
```suggestion
        assert num_inputs >= 1, num_inputs
```

egs/tedlium3/ASR/conformer_ctc2/combiner.py
```diff
@@ -0,0 +1,248 @@
...
+        Args:
+          inputs:
+            A list of Tensor, e.g. from various layers of a transformer.
+            All must be the same shape, of (*, num_channels)
+        Returns:
+          A Tensor of shape (*, num_channels). In test mode
+          this is just the final input.
+        """
+        num_inputs = self.num_inputs
+        assert len(inputs) == num_inputs
```
csukuangfj, line 99:
```suggestion
        assert len(inputs) == num_inputs, (len(inputs), num_inputs)
```

egs/tedlium3/ASR/conformer_ctc2/conformer.py
```diff
@@ -0,0 +1,1069 @@
+#!/usr/bin/env python3
```
csukuangfj, line 1: Is it possible to replace this file with a symlink to the one from librispeech's `conformer_ctc2`?

egs/tedlium3/ASR/conformer_ctc2/decode.py
```diff
@@ -0,0 +1,963 @@
...
+    get_texts,
+    load_averaged_model,
+    setup_logger,
+    store_transcripts,
+    str2bool,
+    write_error_stats,
+)
+
+
+def add_model_arguments(parser: argparse.ArgumentParser) -> None:
```
csukuangfj, line 64: Could you import `add_model_arguments` from `train.py` and remove this one from `decode.py`?

egs/tedlium3/ASR/conformer_ctc2/decode.py
```diff
@@ -0,0 +1,963 @@
...
+
+def ctc_greedy_search(
+    nnet_output: torch.Tensor,
+    memory: torch.Tensor,
+    memory_key_padding_mask: torch.Tensor,
+) -> List[List[int]]:
+    """Apply CTC greedy search
+
+     Args:
+         speech (torch.Tensor): (batch, max_len, feat_dim)
```
csukuangfj, line 273: Could you fix the docstring? (The comments do not match the input arguments.)

egs/tedlium3/ASR/conformer_ctc2/decode.py
```diff
@@ -0,0 +1,963 @@
...
+    topk_prob, topk_index = ctc_probs.topk(1, dim=2)  # (B, maxlen, 1)
+    topk_index = topk_index.view(batch_size, maxlen)  # (B, maxlen)
+    topk_index = topk_index.masked_fill_(encoder_mask, 0)  # (B, maxlen)
+    hyps = [hyp.tolist() for hyp in topk_index]
+    scores = topk_prob.max(1)
+    hyps = [remove_duplicates_and_blank(hyp) for hyp in hyps]
+    return hyps, scores
+
+
+def remove_duplicates_and_blank(hyp: List[int]) -> List[int]:
```
csukuangfj, line 294: Could you replace it with
`torch.unique_consecutive`?

It is available since torch >= 1.6.0?
Please see
https://pytorch.org/docs/1.6.0/generated/torch.unique_consecutive.html

You can use a separate call to remove 0s, e.g,
```python3
a = a[a!=0];
```

egs/tedlium3/ASR/conformer_ctc2/decode.py
```diff
@@ -0,0 +1,963 @@
...
+    """
+    batch_size = memory.shape[1]
+    # Let's assume B = batch_size
+    encoder_out = memory
+    encoder_mask = memory_key_padding_mask
+    maxlen = encoder_out.size(0)
+
+    ctc_probs = nnet_output  # (B, maxlen, vocab_size)
+    topk_prob, topk_index = ctc_probs.topk(1, dim=2)  # (B, maxlen, 1)
+    topk_index = topk_index.view(batch_size, maxlen)  # (B, maxlen)
```
csukuangfj, line 286: I would suggest
```python3
topk_index = topk_index.squeeze(-1)
```

egs/librispeech/ASR/local/prepare_lang_bpe.py
videodanchik: Ooops... but I checked, it was introduced after I changed from my local `excluded = ["<eps>", "!sil", "<unk>", "#0", "<s>", "</s>"]` to `excluded = ["<eps>", "!SIL", "!sil", "<SPOKEN_NOISE>", "<UNK>", "<unk>" "#0", "<s>", "</s>"]` to preserve the compatibility with Librispeech and by accident forgot the comma, thanks @csukuangfj good catch!

csukuangfj:
@videodanchik 
Could you please fix the conflicts?

egs/tedlium3/ASR/conformer_ctc2/combiner.py
videodanchik: Done

egs/tedlium3/ASR/conformer_ctc2/combiner.py
videodanchik: Done

egs/tedlium3/ASR/conformer_ctc2/decode.py
videodanchik: With such a modification import of `add_model_arguments` from `train.py` it runs `torch.set_num_interop_threads(1)` twice: first time in `train.py` because of import and second time at the bottom of `decode.py` script. This causes an error for my Pytorch 1.13.
```
terminate called after throwing an instance of 'c10::Error'
  what():  Error: cannot set number of interop threads after parallel work has started or set_num_interop_threads called
Exception raised from set_num_interop_threads at ../aten/src/ATen/ParallelThreadPoolNative.cpp:54 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::string) + 0x3e (0x7f13d5b8486e in /opt/conda/lib/python3.10/site-packages/torch/lib/libc10.so)
frame #1: c10::detail::torchCheckFail(char const*, char const*, unsigned int, char const*) + 0x60 (0x7f13d5b4f469 in /opt/conda/lib/python3.10/site-packages/torch/lib/libc10.so)
frame #2: <unknown function> + 0x196eacf (0x7f1402c4eacf in /opt/conda/lib/python3.10/site-packages/torch/lib/libtorch_cpu.so)
frame #3: <unknown function> + 0x69115a (0x7f142b86215a in /opt/conda/lib/python3.10/site-packages/torch/lib/libtorch_python.so)
<omitting python frames>
frame #15: __libc_start_main + 0xf3 (0x7f144d59a083 in /usr/lib/x86_64-linux-gnu/libc.so.6)

Aborted (core dumped)
```
To prevent this I added an `if` statement in `decode.py`
```
# when we import add_model_arguments from train.py
# we enforce torch.set_num_interop_threads(1) in it,
# so we ended up with setting num_interop_threads to one
# two times: in train.py and decode.py which cause an error,
# that is why added an additional if statement.
if torch.get_num_interop_threads() != 1:
    torch.set_num_interop_threads(1)

```

videodanchik:
@csukuangfj @danpovey 

Hi guys, sorry for a bit slow progress here, but I think I've solved all these questions about `<unk>` encoding.

There seems to be an issue in how you perform `<unk>` encoding during training in `tedlium3` recipes compared to how the BPE model does the encoding of unknown words. I first refer to [this function](https://github.com/k2-fsa/icefall/blob/a0cf85343dad31a678ddaac7652f0bb2bbb4cac2/egs/tedlium3/ASR/local/convert_transcript_words_to_bpe_ids.py#L39) especially considering the examples that provide the explanation how this function works.

> texts = ["\<unk\> this is a sunny day"]
> spm_ids = [[2, 38, 33, 6, 118, 11, 11, 21, 316]]
> 
> texts = ["\<unk\>"]
> spm_ids = [[2]]

This is not how this function should work I believe. If I take these examples and perform the encoding with my BPE model (or any other BPE model you have from any `icefall` recipe) through the aforementioned function (I'm using `sentencepiece==0.1.97` there is an `emit_unk_piece` option appeared in `encode` method) and compare results with direct encoding with `sp.encode`, here what I'm getting:
```
import sentencepiece as spm
from egs.tedlium3.ASR.local.convert_transcript_words_to_bpe_ids import convert_texts_into_ids

sp = spm.SentencePieceProcessor("icefall2/egs/tedlium3/ASR/data/lang_bpe_500/bpe.model")

texts = ["<unk>"]
convert_texts_into_ids(texts, unk_id=2, sp=sp)

>>> [[2]]
```
Now direct `sp.encode` of `texts = ["<unk>"]` as input will result in a split into multiple tokens during processing (actually why `convert_texts_into_ids` was implemented), for this reason, I replace the input to `texts = ["@"]`. My BPE model was not trained to encode this "word" and will output `unk_id`, but actually with small addition:
```
texts = ["@"]
sp.encode(texts, out_type=int), sp.encode(texts, out_type=str, emit_unk_piece=True)

>>> ([[15, 2]], [['▁', '<unk>']])
```
So it encodes with a "space" token before the `<unk>` token. Similar with input `texts = ["<unk> this is a sunny day"]`:
```
convert_texts_into_ids(["<unk> this is a sunny day"], unk_id=2, sp=sp)

>>> [[2, 86, 32, 6, 107, 8, 8, 16, 325]]
```
And with `<unk>` -> `@` replacement in direct encoding we get:
```
texts = ["@ this is a sunny day"]
sp.encode(texts, out_type=int), sp.encode(texts, out_type=str, emit_unk_piece=True)

>>> ([[15, 2, 86, 32, 6, 107, 8, 8, 16, 325]],
>>> [['▁', '<unk>', '▁this', '▁is', '▁a', '▁su', 'n', 'n', 'y', '▁day']])
```
The difference is an additional "space" token in the beginning of each `<unk>`.

There is no difference in how many characters the `<unk>` word consists of:  
```
texts1_direct = ["@@"]
texts2_direct = ["@@ this is a sunny day"]
print(sp.encode(texts1_direct, out_type=int), sp.encode(texts1_direct, out_type=str, emit_unk_piece=True))
print(sp.encode(texts2_direct, out_type=int), sp.encode(texts2_direct, out_type=str, emit_unk_piece=True))

>>> [[15, 2]] [['▁', '<unk>']]
>>> [[15, 2, 86, 32, 6, 107, 8, 8, 16, 325]] [['▁', '<unk>', '▁this', '▁is', '▁a', '▁su', 'n', 'n', 'y', '▁day']]
```
BPE model doesn't output a "space" token before `<unk>` token only if it is attached to the previous word:
```
texts1_direct = ["i hate these fu***ng unks"]
print(sp.encode(texts1_direct, out_type=int), sp.encode(texts1_direct, out_type=str, emit_unk_piece=True))
 
>>> [[23, 141, 84, 314, 49, 34, 2, 116, 165, 47, 3]] 
>>> [['▁i', '▁ha', 'te', '▁these', '▁f', 'u', '<unk>', 'ng', '▁un', 'k', 's']]
```
BTW if you feed something like `text=["<unk> this is a sunny day @"]` to `convert_texts_into_ids` you will end up with:
```
convert_texts_into_ids(["<unk> this is a sunny day @"], unk_id=2, sp=sp)

>>> [[2, 86, 32, 6, 107, 8, 8, 16, 325, 15, 2]]
```
where you don't have space before first <unk> and naturally output it before encoding `@`. 

Now the outcome of all that's been said is that the BPE model naturally emits a "space" (`▁`) token in most cases before `<unk>` token when it fails to encode the input. With the current `pruned_transducer_stateless` and `transducer_stateless` recipes, all "space" tokens before `<unk>` token were removed and models were trained to output `<unk>` token "attached" to the previous token and you ended up with (see p.2 in https://github.com/k2-fsa/icefall/pull/183#issuecomment-1048550862). I suggest preserving the natural way of encoding `<unk>` with the "space" token inserted before and modifying https://github.com/k2-fsa/icefall/blob/a0cf85343dad31a678ddaac7652f0bb2bbb4cac2/egs/tedlium3/ASR/local/convert_transcript_words_to_bpe_ids.py#L61 to `y_ids.extend(id_segments[i] + [sp.piece_to_id("▁"), sp.unk_id()])`

Currently, in this PR it's a nasty hack, look at line `536` of `train.py`: `texts = [t.replace("<unk>", "@") for t in texts]`, I will change it shortly.

Now a separate question is how do we create `lexicon.txt` in all recipes. If you look at any `lexicon.txt` of any recipe in `icefall` you can notice that each word transcription in it starts from `▁` character whether it's a separate token or a part of another token that begins with this character. All word transcriptions are like this except the transcription of `<UNK>` **word**: currently it is `<UNK> <unk>` see https://github.com/k2-fsa/icefall/blob/a0cf85343dad31a678ddaac7652f0bb2bbb4cac2/egs/librispeech/ASR/local/prepare_lang_bpe.py#L160 I'm also suggesting to add a "space" token before `<unk>` token to make it similar with the rest of the lexicon. As it's easy to recompile the `HLG.pt` upon the lexicon where the `<UNK>` word is encoded with or without the `▁` token in the beginning, I present decoding results for both options. 

I have trained 2 final models after some experimentation, training command is the following:
```
./conformer_ctc2/train.py \
    --world-size 4 \
    --num-epochs 40 \
    --use-fp16 true \
    --max-duration 350 \
    --num-workers 4
```
 In the first model, I preserved replacement `<unk> -> @` and feed the resulting transcript to `sp.encode` making it output `sp.unk_id()` **with inserted "space" token** before, each time it faces `@`. Next, I performed the decoding with `"ctc-decoding", "1best", "whole-lattice-rescoring", "attention-decoder"` methods preserving the best results after tuning. For `"1best", "whole-lattice-rescoring", "attention-decoder"` methods I performed decoding using two versions of `HLG.pt`, where lexicon compiled with and without `▁` token before `<unk>` token for `<UNK>` word transcription. I got the following results:

method | epoch | avg | language model weight | attention scale | WER_dev | WER_test | WER_avg | with ▁ token in lexicon
-- | -- | -- | -- | -- | -- | -- | -- | --
ctc-decoding | 40 | 30 | NaN | NaN | 6.37 | 5.92 | 6.145 | doesn't matter
1best | 40 | 30 | 0.1 | NaN | 5.88 | 5.46 | 5.670 | True
1best | 32 | 22 | 0.1 | NaN | 6.21 | 5.67 | 5.940 | False
whole-lattice-rescoring | 40 | 27 | 0.3 | NaN | 5.90 | 5.48 | 5.690 | True
whole-lattice-rescoring | 24 | 15 | 0.1 | NaN | 6.28 | 5.65 | 5.965 | False
attention-decoder | 38 | 26 | 0.01 | 3.00 | 5.60 | 5.33 | 5.465 | True
attention-decoder | 29 | 18 | 0.01 | 0.70 | 6.06 | 5.48 | 5.770 | False

So when I don't add `▁` token I'm getting worse results, when I looked at WER reports for the second and third lines from the table but didn't find any patterns. The `tensorboard` log is available here: https://tensorboard.dev/experiment/5NQQiqOqSqazfn4w2yeWEQ/
Notice the validation ctc and attention losses, it starts to diverge significantly after some time, I blame the absence of `<unk>` words in both dev and test sets.

My second model was trained with the complete removal of `<unk>` words from reference transcripts.

method | epoch | avg | language model weight | attention scale | WER_dev | WER_test | WER_avg | with ▁ token in lexicon
-- | -- | -- | -- | -- | -- | -- | -- | --
ctc-decoding | 40 | 30 | NaN | NaN | 6.46 | 6.05 | 6.255 | doesn't matter
1best | 37 | 28 | 0.1 | NaN | 5.95 | 5.61 | 5.780 | True
1best | 37 | 28 | 0.1 | NaN | 5.94 | 5.59 | 5.765 | False
whole-lattice-rescoring | 34 | 25 | 0.2 | NaN | 6.02 | 5.62 | 5.820 | True
whole-lattice-rescoring | 34 | 25 | 0.2 | NaN | 6.01 | 5.60 | 5.805 | False
attention-decoder | 38 | 28 | 0.08 | 3.00 | 5.78 | 5.50 | 5.640 | True
attention-decoder | 38 | 28 | 0.10 | 4.00 | 5.73 | 5.46 | 5.595 | False

Here decoding with or without  `▁` token doesn't make much difference, but both results are worse than the model trained with `<unk>` and with `▁` token in lexicon before `<unk>` token. Tensorboard available here: https://tensorboard.dev/experiment/wLAAaGoCQBuaSkpakaT67g/

I still observe diverging of validation losses but way less compared to the first model where we didn't exclude `<unk>` from the reference, it's could be an overfitting sign, but at the same time, when I perform decoding with `--avg 1` and `1best` method I do not observe overfitting signs like WER increase:
image

Anyway, to summarize all these results, I believe that we should proceed with the model that was trained **WITH** `<unk>` in transcripts + that is using `▁ <unk>` as an `<UNK>` word transcript in the lexicon, as it seems to produce the best WER results.

I should have probably tried to train a separate model without `▁` token before `<unk>` token during training, which corresponds to `<unk>` "attached" to the previous token, something that we currently have in `pruned_transducer_stateless` and `transducer_stateless` recipes, but I decided to skip this set up as it seems to be unnatural to train such models.

csukuangfj:
@videodanchik 
Thanks for your detailed analysis and comparison.

>  (actually why convert_texts_into_ids was implemented)

If I recall correctly, the first attempt treated `<unk>` as a normal word and encoded it into 5 pieces (https://github.com/k2-fsa/icefall/pull/183#discussion_r806451724) and then we fixed it by following (https://github.com/k2-fsa/icefall/pull/183#discussion_r806450892) and finally we kept it since the result shown in https://github.com/k2-fsa/icefall/pull/183#issuecomment-1047583114 was promising.

> Currently, in this PR it's a nasty hack, look at line 536 of train.py: texts = [t.replace("<unk>", "@") for t in texts], I will change it shortly.

I think we need to handle the case when `@` is a valid BPE token. Can we modify the implementation of `convert_texts_into_ids` to insert a blank before `<unk>` manually?

> I'm also suggesting to add a "space" token before <unk> token to make it similar with the rest of the lexicon

Yes, that is a good idea.

> I believe that we should proceed with the model that was trained WITH <unk> in transcripts + that is using ▁ <unk> as an <UNK> word transcript in the lexicon, as it seems to produce the best WER results.

Agreed.

videodanchik:
@csukuangfj I performed the final fixes you asked for, see: https://github.com/k2-fsa/icefall/pull/696/commits/d9f46092b9086bf267eee50df9ecc99288b9b248

I believe this is ready for merge. What else should be done to make it happen?

egs/tedlium3/ASR/conformer_ctc2/conformer.py
videodanchik: @csukuangfj Not sure really, as I mentioned in the first comment to this PR, I've added `RandomCombine` and `aux_layers` to `ConformerEncoder`, not sure that this is compatible with librispeech `conformer_ctc2`

csukuangfj:
Thanks!

> I believe this is ready for merge. What else should be done to make it happen?

(1) Could you fix the CI test and style check issues?
Please have a look at https://k2-fsa.github.io/icefall/contributing/code-style.html
for the version of black and flake8 we are using

(2) Also, could you update
https://github.com/k2-fsa/icefall/blob/master/egs/tedlium3/ASR/RESULTS.md
and upload your pre-trained models and decoding results to a hugging face repository?

test/test_lexicon.py
```diff
@@ -112,7 +112,7 @@ def uniq_lexicon_test():
     # But there is no word "ca" in the lexicon, so our
     # implementation returns the id of "<UNK>"
     print(token_ids, expected_token_ids)
-    assert token_ids.tolist() == [[sp.unk_id()]]
+    assert token_ids.tolist() == [[sp.piece_to_id("▁"), sp.sp.unk_id()]]
```
csukuangfj, line 115:
```suggestion
    assert token_ids.tolist() == [[sp.piece_to_id("▁"), sp.unk_id()]]
```

videodanchik:
@csukuangfj Should be ready now.

csukuangfj:
@videodanchik 

Thanks! Could you fix the test error reported by GitHub actions?
https://github.com/k2-fsa/icefall/actions/runs/3672409740/jobs/6208556493#step:8:295

```
=================================== FAILURES ===================================
_____________________ test_with_torch_label_smoothing_loss _____________________

    def test_with_torch_label_smoothing_loss():
        if torch_ver < LooseVersion("1.10.0"):
            print(f"Current torch version: {torch_ver}")
            print("Please use torch >= 1.10 to run this test - skipping")
            return
        torch.manual_seed(20211105)
        x = torch.rand(20, 30, 5000)
        tgt = torch.randint(low=-1, high=x.size(-1), size=x.shape[:2])
        for reduction in ["none", "sum", "mean"]:
            custom_loss_func = LabelSmoothingLoss(
                ignore_index=-1, label_smoothing=0.1, reduction=reduction
            )
            custom_loss = custom_loss_func(x, tgt)
    
            torch_loss_func = torch.nn.CrossEntropyLoss(
                ignore_index=-1, reduction=reduction, label_smoothing=0.1
            )
            torch_loss = torch_loss_func(x.reshape(-1, x.size(-1)), tgt.reshape(-1))
>           assert torch.allclose(custom_loss, torch_loss)
E           RuntimeError: The size of tensor a (5000) must match the size of tensor b (600) at non-singleton dimension 1

test_label_smoothing.py:44: RuntimeError
=========================== short test summary info ============================
FAILED test_label_smoothing.py::test_with_torch_label_smoothing_loss - RuntimeError: The size of tensor a (5000) must match the size of tensor b (600) at non-singleton dimension 1
========================= 1 failed, 8 passed in 2.22s ==========================
Error: Process completed with exit code 1.
```

csukuangfj:
@videodanchik 

I just updated https://huggingface.co/spaces/k2-fsa/automatic-speech-recognition
to include your pre-trained models (from https://huggingface.co/videodanchik/icefall-asr-tedlium3-conformer-ctc2)

Screen Shot 2022-12-12 at 11 47 15 AM

Screen Shot 2022-12-12 at 11 48 20 AM

csukuangfj:
> @csukuangfj I wonder how does it work with decoding with greedy_search and modified_beam_search methods upon the Conformer CTC model?

greedy search and modified beam search are applicable only for transducer-based models.

That page should update the decoding methods when a CTC model is selected.

(By default, it uses a CTC topology for decoding when a CTC model is selected).

csukuangfj (approved):

csukuangfj:
Thanks!

videodanchik:
> greedy search and modified beam search are applicable only for transducer-based models.
> 
> That page should update the decoding methods when a CTC model is selected.
> 
> (By default, it uses a CTC topology for decoding when a CTC model is selected).
@csukuangfj I see thanks, but it looks like there is a bug or something as I still see transducer decoding methods for my model.
image

csukuangfj:
> @csukuangfj I see thanks, but it looks like there is a bug or something as I still see transducer decoding methods for my model.


Yes, it can be considered a bug. Sorry, I don't have time to fix it.

PR #717, opened 2022-11-30
Convert conv-emformer to ncnn

csukuangfj:
It is working now.  

## TODOs
- [ ] Add doc
- [x] Add CI tests
- [ ] Code refactoring
- [x] Add to [sherpa-ncnn](https://github.com/k2-fsa/sherpa-ncnn)


csukuangfj:
Here is a demo for this PR.

The code is at https://github.com/k2-fsa/sherpa-ncnn


https://user-images.githubusercontent.com/5284924/205819670-e76f4f49-d0e9-449e-a754-0e81042c82ed.mov

csukuangfj:
emformer2.py:
(1) It is modified from emformer.py and its sole purpose is for model exporting (supporting only streaming recognition).
(2) We also change the way how `states` is represented. (We use a list of tensors in emformer2.py)

The reason to have a separate version of emformer2.py
(1) emformer.py contains code both for streaming and non-streaming ASR
(2) For model exporting, we only need the streaming part
(3) Making a separate copy instead of changing `emformer.py` in-place is simpler

(I suggest that we don't put streaming and non-streaming in the same class in the future).

---

train2.py:
It is copied from train.py and uses `emformer2.py` inside.


---

Note: I should have used `streaming_emformer.py` to replace `emformer2.py`.

desh2608:
> emformer2.py: (1) It is modified from emformer.py and its sole purpose is for model exporting (supporting only streaming recognition). (2) We also change the way how `states` is represented. (We use a list of tensors in emformer2.py)
> 
> The reason to have a separate version of emformer2.py (1) emformer.py contains code both for streaming and non-streaming ASR (2) For model exporting, we only need the streaming part (3) Making a separate copy instead of changing `emformer.py` in-place is simpler
> 
> (I suggest that we don't put streaming and non-streaming in the same class in the future).
> 
> train2.py: It is copied from train.py and uses `emformer2.py` inside.
> 
> Note: I should have used `streaming_emformer.py` to replace `emformer2.py`.

Thanks for the explanation. I asked because I am working on a transducer-based multi-talker ASR model that would need a streaming encoder, so I was wondering which is the recommended streaming encoder setup in icefall currently.

csukuangfj:
You can use either streaming conformer (pruned_transducer_statelessX)
or
conv-emformer (conv_emformer_pruned_stateless2)

If you use the latter one, you can easily export it to ncnn once you finish your training.

PR #722, opened 2022-12-01
Add Blankskip to Zipformer+CTC

yfyeung:
Accelerating RNN-T Training and Inference Using CTC guidance
https://arxiv.org/abs/2210.16481


kobenaxie:
Hi, yifan, why not reduce frames at middle layers ?

csukuangfj (changes requested):
Please add `.DS_Store` to `.gitignore`.

kobenaxie:
> Hi, yifan, why not reduce frames at middle layers ?

And  another question, is it possible the reduced encoder frames shorter than label sequence(For the case deletion error in CTC) ?

yfyeung:
> Please add `.DS_Store` to `.gitignore`.

OK

Issue #724, opened 2022-12-01
Crash testing "yesno" recipe


iggygeek:
Hello,
I need help.
Testing my installation with  yesno recipe, here is the error I get :

2022-12-01 15:38:57,293 INFO [train.py:481] Training started
2022-12-01 15:38:57,293 INFO [train.py:482] {'exp_dir': PosixPath('tdnn/exp'), 'lang_dir': PosixPath('data/lang_phone'), 'lr': 0.01, 'feature_dim': 23, 'weight_decay': 1e-06, 'start_epoch': 0, 'best_train_loss': inf, 'best_valid_loss': inf, 'best_train_epoch': -1, 'best_valid_epoch': -1, 'batch_idx_train': 0, 'log_interval': 10, 'reset_interval': 20, 'valid_interval': 10, 'beam_size': 10, 'reduction': 'sum', 'use_double_scores': True, 'world_size': 1, 'master_port': 12354, 'tensorboard': True, 'num_epochs': 15, 'seed': 42, 'feature_dir': PosixPath('data/fbank'), 'max_duration': 30.0, 'bucketing_sampler': False, 'num_buckets': 10, 'concatenate_cuts': False, 'duration_factor': 1.0, 'gap': 1.0, 'on_the_fly_feats': False, 'shuffle': True, 'return_cuts': True, 'num_workers': 2, 'env_info': {'k2-version': '1.23.2', 'k2-build-type': 'Release', 'k2-with-cuda': True, 'k2-git-sha1': '4e6b87bb75a6430fb7fcf5cfa4adfa9eabe3ad6c', 'k2-git-date': 'Sat Nov 26 13:34:17 2022', 'lhotse-version': '1.10.0', 'torch-version': '1.13.0', 'torch-cuda-available': False, 'torch-cuda-version': '11.6', 'python-version': '3.9', 'icefall-git-branch': 'master', 'icefall-git-sha1': '2bca703-clean', 'icefall-git-date': 'Thu Dec 1 08:57:43 2022', 'icefall-path': '/home/usera/git/icefall', 'k2-path': '/home/usera/miniconda3/envs/icefall/lib/python3.9/site-packages/k2/__init__.py', 'lhotse-path': '/home/usera/miniconda3/envs/icefall/lib/python3.9/site-packages/lhotse/__init__.py', 'hostname': 'DELL', 'IP address': '127.0.0.1'}}
[...]

PR #725, opened 2022-12-02
Fix for older versions of k2

csukuangfj:
See #724
Older versions of k2 do not support using a bool tensor for indexing.

The change was introduced by https://github.com/k2-fsa/icefall/pull/669

I suggest that we always add test to cover new changes.

### DIFF
--- a/icefall/graph_compiler.py
+++ b/icefall/graph_compiler.py
@@ -81,7 +81,7 @@ class CtcTrainingGraphCompiler(object):
 
         self.ctc_topo._is_repeat_token_ = (
             self.ctc_topo.labels != self.ctc_topo.aux_labels
-        )
+        ).int()
 
         decoding_graph = k2.compose(
             self.ctc_topo, fsa_with_self_loops, treat_epsilons_specially=False


pkufool:
Just want to save some memory by using `bool`,  It is might not the best choice to just convert it to `int`.  This line is not needed for older version code (or it is not needed if we don't want delay-penalty), could we add a flag here?

csukuangfj:
I think the `is_repeat` flag should be added when we create `ctc_topo.`  We don't need to change the icefall code. The `is_repeat` flag should not be visible to users in icefall.

pkufool:
Em, It might be better to set this attr in k2.

pkufool:
I will file a PR, thanks!

pkufool:
But, we do need a flag in icefall, because we create ctc_topo in icefall. We can set the default value to False.

pkufool:
> But, we do need a flag in icefall, because we create ctc_topo in icefall. We can set the default value to False.

I mean, I don't want this extra attr if we don't use delay-penalty.

csukuangfj:
> > But, we do need a flag in icefall, because we create ctc_topo in icefall. We can set the default value to False.
> 
> I mean, I don't want this extra attr if we don't use delay-penalty.

In that case, I suggest that you change `k2.ctc_topo` to take an optional argument, e.g., `need_repeat_flag=False`.

pkufool:
I think add a flag to `CtcTrainingGraphCompiler` is enough.  This is only need for phone-based ctc, it depends on what modeling units you choose in icefall.

PR #727, opened 2022-12-02
Add need_repeat_flag in phone based ctc graph compiler

pkufool:


icefall/graph_compiler.py
```diff
@@ -29,6 +29,7 @@ def __init__(
         lexicon: Lexicon,
         device: torch.device,
         oov: str = "<UNK>",
+        need_repeat_flag: bool = False,
```
csukuangfj, line 32: You may also need to modify the corresponding `train.py` to pass True to this funciton.

pkufool: I think we only have bpe models that support delay-penalty, @yaozengwei  Did you update the phone-based recipes?

csukuangfj: I am meaning that will your current change break any existing recipes?

pkufool: I think it won't break any recipes,  will check it again.

csukuangfj: Previously, the repeat attr is always added. But this PR removes it since the arg is always False.

pkufool: Yes, you are right, we never tried the phone-based models, but it did support in our code. Will fix it.

yaozengwei: > that

Yeah. I also updated the phone-based graph-compiler.

egs/librispeech/ASR/conformer_ctc3/train.py
```diff
@@ -890,6 +890,7 @@ def run(rank, world_size, args):
         graph_compiler = CtcTrainingGraphCompiler(
             lexicon,
             device=device,
+            need_repeat_flag=True if params.delay_penalty > 0 else False,
```
csukuangfj, line 893:
```suggestion
            need_repeat_flag=params.delay_penalty > 0
```
is simpler

csukuangfj (approved):

PR #729, opened 2022-12-02
Update DoubleSwish

yaozengwei:
This PR aims to update the DoubleSwish function. It just replaces `x * sigmoid(x-1)`  with  `x * sigmoid(x-1) - 0.05x`. 

Experimental results on Dan's Zipformer training on `train-clean-100` shows that it can slightly outperform the baseline:

epoch-20-avg-5: 
- Baseline, 6.38 & 17.29, 
- Modified DoubleSwish, 6.33 & 17.15

epoch-25-avg-10:
- Baseline, 6.04 & 16.41
- Modified DoubleSwish, 6.02 & 16.28 

epoch-30-avg-13: 
- Baseline, 6.0 & 15.85
- Modified DoubleSwish, 5.97 & 15.78

- [ ] Make it compatible with existing recipes. 
- [ ] Enable to save and load the `alpha` attribute using the `_metadata` in `torch.nn.Module` (https://github.com/pytorch/pytorch/blob/f62e54df8f04362db96154a94137d6afbdfaa953/torch/nn/modules/module.py#L404)


csukuangfj:
Looks great to me.

csukuangfj (approved):

egs/librispeech/ASR/pruned_transducer_stateless7/scaling.py
```diff
@@ -994,28 +995,60 @@ def forward(ctx, x: Tensor) -> Tensor:
...
         floor = -0.043637
         ceil = 1.2
         d = d * ((ceil - floor) / 255.0) + floor
-        return y_grad * d
+        ans = y_grad * (d + alpha) if alpha != 0 else y_grad * d
+        return ans, None
 
 
 class DoubleSwish(torch.nn.Module):
+    def __init__(self, alpha: float = 0.0):
```
csukuangfj, line 1017: Shall we give it a default value `-0.05`?

csukuangfj:
could you also update

yaozengwei:
> could you also update https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/RESULTS-100hours.md

OK. I will do that.

egs/librispeech/ASR/pruned_transducer_stateless7/scaling.py
yaozengwei: I think we should keep `alpha=0.0` as default. Since for existing repices, the models are trained with `alpha=0.0`, and their saved `state_dict` does not store this attribute.

danpovey:
I'm slightly having second thoughts about merging this, since the improvement was quite small.  It's OK to merge if you're very confident that it's not going to cause us problems in future, e.g. due to mistakes in the back-compatibility code for example.  Or runtime issues, if relevant.
Make sure that anything inside "if torch.jit.is_scripting()" is correct.

PR #732, opened 2022-12-04
 Filter the case of T <  S  for wenetspeech training data

v-yunbin:


egs/wenetspeech/ASR/pruned_transducer_stateless2/train.py
```diff
@@ -861,15 +861,40 @@ def run(rank, world_size, args):
...
         #
-        # Caution: There is a reason to select 15.0 here. Please see
+        # Caution: There is a reason to select 20.0 here. Please see
         # ../local/display_manifest_statistics.py
         #
         # You should use ../local/display_manifest_statistics.py to get
         # an utterance duration distribution for your dataset to select
         # the threshold
-        return 1.0 <= c.duration <= 15.0
+        if c.duration < 1.0 or c.duration > 20.0:
```
csukuangfj, line 872: Could you please replace 20 with 10?

According to
https://github.com/k2-fsa/icefall/blob/master/egs/wenetspeech/ASR/local/display_manifest_statistics.py#L52,
choose 10 here will remove only about ~0.1% of the data.

csukuangfj:
Are you going to reopen it?

PR #737, opened 2022-12-06
Update prepare.sh

ahazned:
- Changed the wrong filenames when shuffling the train cuts.

- Changed special word "[UNK]" into \<UNK\> in order to match with the special word used in local/prepare_lang_bpe.sh


csukuangfj:
Thanks for your contribution!

PR #753, opened 2022-12-09
 Filter the training data of T <  S  for Wenet train recipe

v-yunbin:
In pruned RNN-T, it  require that T >= S where T is the number of feature frames after subsampling and S is the number of tokens in the utterance filter the case of T <  S  for training data.


csukuangfj:
Thanks for your contribution!

csukuangfj (approved):

PR #783, opened 2022-12-23
Optimize frame_reducer.py

yfyeung:
Optimize frame_reducer.py


csukuangfj (approved):

kobenaxie:
Hi @yfyeung , is it possible that the length of frame reduced encoder outputs shorter than label on some speech data ? Generally the deletion error in CTC model . if `x_lens_fr < y_lens`, CTC or RNNT loss computation may raise error `T < S` ?

yfyeung:
> Hi @yfyeung , is it possible that the length of frame reduced encoder outputs shorter than label on some speech data ? Generally the deletion error in CTC model . if `x_lens_fr < y_lens`, CTC or RNNT loss computation may raise error `T < S` ?

@kobenaxie  Thanks for your suggestion, and here is the PR: https://github.com/k2-fsa/icefall/pull/875

Issue #800, opened 2023-01-01
New tedlium3 conformet_ctc recipe problem


AlexandderGorodetski:
Hello guys,

I am trying new  tedlium3 conformet_ctc recipe which was lately added to icefall. I created following script to run training and decoder together:

export CUDA_VISIBLE_DEVICES="0,1,2,3"

./conformer_ctc2/train.py \
    --world-size 4 \
    --num-epochs 40 \
    --exp-dir conformer_ctc2/exp \
    --max-duration 150 \
    --use-fp16 true

epoch=38
avg=26

## ctc decoding
./conformer_ctc2/decode.py \
  --method ctc-decoding \
  --exp-dir conformer_ctc2/exp \
  --lang-dir data/lang_bpe_500 \
  --result-dir conformer_ctc2/exp \
  --max-duration 500 \
  --epoch $epoch \
  --avg $avg

Its behaviour is weird. After 3 days of training, and after the training was finished I got following messages. It seems that directory ~/icefall/egs/tedlium3/ASR/conformer_ctc2/exp was completed deleted. And all outputs that were created during the training were deleted. Could you help please.

2022-12-30 08:21:00,987 INFO [train.py:990] (3/4) Done!
2022-12-30 08:21:01,016 INFO [train.py:990] (2/4) Done!
2022-12-30 08:21:01,028 INFO [train.py:990] (1/4) Done!
2022-12-30 08:21:01,030 INFO [checkpoint.py:75] (0/4) Saving checkpoint to conformer_ctc2/exp/epoch-40.pt
2022-12-30 08:21:04,603 INFO [train.py:990] (0/4) Done!
fatal: detected dubious ownership in repository at '/fsx_new/Alex/tmp1/icefall'
To add an exception for this directory, call:

        git config --global --add safe.directory /fsx_new/Alex/tmp1/icefall
fatal: detected dubious ownership in repository at '/fsx_new/Alex/tmp1/icefall'
To add an exception for this directory, call:

        git config --global --add safe.directory /fsx_new/Alex/tmp1/icefall
fatal: detected dubious ownership in repository at '/fsx_new/Alex/tmp1/icefall'
To add an exception for this directory, call:

        git config --global --add safe.directory /fsx_new/Alex/tmp1/icefall
2022-12-30 08:21:08,389 INFO [decode.py:646] Decoding started
[...]

PR #801, opened 2023-01-01
Not removing result_dir in tedlium3 conformer ctc2

videodanchik:
This is to address https://github.com/k2-fsa/icefall/issues/800 + several small fixes.


csukuangfj:
Thanks!

PR #833, opened 2023-01-11
Add export to ONNX using in Triton server for Zipformer

wd929:
Hi,
This PR supports exporting offline Zipformer model to ONNX format which can be used in Triton server.
Thanks!

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless7/onnx_wrapper.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless7/onnx_wrapper.py
@@ -15,6 +15,7 @@
 import torch
 from torch import nn
 
+
 class TritonOnnxDecoder(nn.Module):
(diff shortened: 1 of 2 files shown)


egs/librispeech/ASR/pruned_transducer_stateless7/onnx_wrapper.py
```diff
@@ -0,0 +1,76 @@
...
+
+class TritonOnnxJoiner(nn.Module):
+    def __init__(
+        self,
+        model,
+    ):
+        super().__init__()
+
+        self.model = model
+        self.encoder_proj = model.encoder_proj
```
yuekaizhang, line 58: The two lines may could be removed.

yuekaizhang:
Could you please add a --fp16 option? We could export onnx fp16 model using onnxmltools.

PR #835, opened 2023-01-12
Use dynamic threshold to reduce frame, and small fix for lconv

yfyeung:


egs/librispeech/ASR/pruned_transducer_stateless7_ctc_bs/frame_reducer.py
```diff
@@ -65,7 +68,9 @@ def forward(
         """
 
         padding_mask = make_pad_mask(x_lens)
-        non_blank_mask = (ctc_output[:, :, blank_id] < math.log(0.9)) * (~padding_mask)
+        non_blank_mask = (ctc_output[:, :, blank_id] < math.log(threshold)) * (
```
csukuangfj, line 71: Please add a command line argument to enable using dynamic threshold. Its default value should be False so that it won't break any existing recipes.

yfyeung: OK

PR #843, opened 2023-01-16
Filter uneven-sized batch 

yaozengwei:
During training, for the uneven-sized batch, the total duration after padding would possibly cause OOM. Hence, for each batch, which is sorted descendingly by length, we simply drop the last few shortest samples, so that the retained total frames (after padding) would not exceed `allow_max_frames`:
`allow_max_frames = int(max_frames * (1.0 + allow_excess_duration_ratio))`,
 where `max_frames = max_duration * 1000 // frame_shift_ms`.
 We set allow_excess_duration_ratio=0.1.

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless7/train.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless7/train.py
@@ -82,7 +82,13 @@ from icefall.checkpoint import (
 from icefall.dist import cleanup_dist, setup_dist
 from icefall.env import get_env_info
 from icefall.hooks import register_inf_check_hooks
-from icefall.utils import AttributeDict, MetricsTracker, setup_logger, str2bool
+from icefall.utils import (
+    AttributeDict,
+    MetricsTracker,
+    filter_uneven_sized_batch,
+    setup_logger,
+    str2bool,
+)
 
 LRSchedulerType = Union[torch.optim.lr_scheduler._LRScheduler, optim.LRScheduler]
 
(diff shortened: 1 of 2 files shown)


egs/librispeech/ASR/pruned_transducer_stateless7/train.py
```diff
@@ -642,6 +666,11 @@ def compute_loss(
      warmup: a floating point value which increases throughout training;
         values >= 1.0 are fully warmed up and have all modules present.
     """
+    if params.filter_uneven_sized_batch:
+        max_frames = params.max_duration * 1000 // params.frame_shift_ms
+        allow_max_frames = int(max_frames * (1.0 + params.allow_excess_duration_ratio))
```
csukuangfj, line 671:
```suggestion
        allowed_max_frames = int(max_frames * (1.0 + params.allow_excess_duration_ratio))
```

csukuangfj (approved):
Thanks! Looks good to me.

danpovey:
Guys, what I was suggesting RE the option was not to have it True by default but not to have the option at all, as I don't see a circumstance where we'd really want it.  OK to keep it in the short term for testing, but in the longer term we need to be a bit disciplined about removing unnecessary options and just hardcoding them.  For experiments you can always just change the code, and I think supporting too many different options makes the code less readable and is not necessary unless people would actually want to change the option frequently.

danpovey:
BTW, this makes a huge difference to maximum memory consumption.   In a full-libri experiment I am doing, the maximum memory usage (at a certain point in the program) changes from 22.0G to 28.4G.   (tag   @pzelasko  for his info)

pzelasko:
Thanks, this is interesting! If I understand correctly this approach can be described as oversampling and discarding a portion of the cuts that are too short. I wonder if you separately tried to tweak num_buckets and how does it compare?

In general I'm interested to improve the quality of this sampler, do you know of a way I can measure actual CUDA memory usage for a single forward + backward call? I could spend some time tweaking it..

danpovey:
You can use torch.cuda.max_memory_allocated(), and there is a way to reset the "max" counter that it uses.
BTW I only view this method as a quick and dirty fix.
We didn't try tweaking num_buckets, IDK what the value we are using is, but I don't view that as a scalable solution regardless.

kli017:
@yaozengwei Hello, I have a question about the uneven-sized batch. Why does the last few shortest samples affect the total memory usage? During the padding operation, dose it padding the short sequence according to the longest one? Does that means we should limit the longest one in the batch?

PR #848, opened 2023-01-17
New recipe: tiny_transducer_ctc

wangtiance:
This recipe is intended for streaming ASR on very low cost devices, with model parameters in the range of 1-2M and less than 0.1GOPS . It uses a small convolutional net as the encoder. It is trained with combined transducer and CTC losses, and supports both phone and BPE lexicons. The encoder consists of 2 subsampling layers followed by a stack of Conv1d-batchnorm-activation-causal_squeeze_excite blocks, with optional skip add. It's a bit similar to CitriNet and ContextNet, but even smaller.

For WER and more details see README.md. Note that transducer decoding does NOT require external LM, therefore the WER looks higher than CTC.


wangtiance:
I get noticeable WER improvements just by enabling tf32 (see https://pytorch.org/docs/stable/notes/cuda.html#tensorfloat-32-tf32-on-ampere-devices). Will update the WER once all models' results are ready.

yaozengwei:
Thanks. Could you also support streaming decoding with cached left context for the causal convolutional modules, like [`streaming_decode.py`](https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/conv_emformer_transducer_stateless2/streaming_decode.py) in recipe `conv_emformer_transducer_stateless2`?

wangtiance:
> Thanks. Could you also support streaming decoding with cached left context for the causal convolutional modules, like [`streaming_decode.py`](https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/conv_emformer_transducer_stateless2/streaming_decode.py) in recipe `conv_emformer_transducer_stateless2`?

Looks like there's some work involved, will get to it when I have the time!

JinZr:
Hi wangtiance, it looks like some of the files need to be black formatted to match the requirements, would you please look into that? 

Thank you!

wangtiance:
> Hi wangtiance, it looks like some of the files need to be black formatted to match the requirements, would you please look into that?
> 
> Thank you!

Reformatted. Now all checks have passed.

JinZr (approved):
LGTM

PR #852, opened 2023-01-19
Edit Zipformer Comments

teowenshen:
A few simple edits on the comments of zipformer, which I found to have helped in my understanding of the codes.

On particular point which I'm not sure - I couldn't find `zipformer_subsampling_factor` anywhere in the codes, so I suspect it's wrong. However, I couldn't understand where (or why) exactly the mask values were repeated. Anyhow I suspect that it's `zipformer_downsampling_factor`, so that was what I edited to.


csukuangfj:
We will be slow in response during Chinese New Year.

yaozengwei:
Thanks.

yaozengwei (approved):

PR #854, opened 2023-01-27
hardcode --filter-uneven-sized-batch

yaozengwei:
See https://github.com/k2-fsa/icefall/pull/843#issuecomment-1386454863


csukuangfj:
Thanks!

PR #856, opened 2023-01-28
fix expired links

weimeng23:
edit README in egs/{aishell,librispeech,timit,yesno}/ASR


csukuangfj:
Thanks for your contribution!

PR #868, opened 2023-02-02
Add pruned_transducer_stateless9

yfyeung:
Keeping a separate vector as an nn.Parameter in the model, and adding it times k/(1+k) to the embedding before the ReLU during training and decoding.

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless9/beam_search.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless9/beam_search.py
@@ -1,0 +1,2382 @@
+# Copyright    2021  Xiaomi Corp.        (authors: Fangjun Kuang
+#                                                  Xiaoyu Yang)
+#
+# See ../../../../LICENSE for clarification regarding multiple authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import warnings
+from dataclasses import dataclass, field
+from typing import Dict, List, Optional, Tuple, Union
+
+import k2
+import sentencepiece as spm
+import torch
+from model import Transducer
+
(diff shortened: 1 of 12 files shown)


egs/librispeech/ASR/pruned_transducer_stateless9/decoder.py
```diff
@@ -0,0 +1,143 @@
...
+          embedding_out:
+            A tensor of shape (N, U, decoder_dim).
+          k:
+            A tensor of shape (N, U).
+            Should be (N, S + 1) during training.
+            Should be (N, 1) during inference.
+        Returns:
+          Return a tensor of shape (N, U, decoder_dim).
+        """
+        return embedding_out + torch.matmul(
```
danpovey, line 97: not very elegant to use matmul here, I think it would be more natural to use multiplication with broadcasting as I think there is no real matrix multiplication going on

egs/librispeech/ASR/pruned_transducer_stateless9/beam_search.py
```diff
@@ -0,0 +1,2416 @@
...
+            timestamp.append(t)
+            decoder_input = torch.tensor([hyp[-context_size:]], device=device).reshape(
+                1, context_size
+            )
+
+            c = torch.tensor([hyp[-context_size - 1 :]], device=device).reshape(
+                1, context_size + 1
+            )
+
+            k = torch.sum(
```
danpovey, line 595: Hm, this does not seem right to me, if you are just computing k from "context_size + 1" elements.  k needs to be computed recursively from the previous k, or it may not have the correct value.   It should be potentially unbounded, here, I don't see how it could have a value more than context_size.

egs/librispeech/ASR/pruned_transducer_stateless9/train.py
```diff
@@ -0,0 +1,1308 @@
...
+    if params.best_train_epoch == params.cur_epoch:
+        best_train_filename = params.exp_dir / "best-train-loss.pt"
+        copyfile(src=filename, dst=best_train_filename)
+
+    if params.best_valid_epoch == params.cur_epoch:
+        best_valid_filename = params.exp_dir / "best-valid-loss.pt"
+        copyfile(src=filename, dst=best_valid_filename)
+
+
+def compute_k(
```
danpovey, line 630: I don't see how this formula can be correct because it is not recursive, i.e. I don't see where the value of k depends on the previous value of k.

egs/librispeech/ASR/pruned_transducer_stateless9/decoder.py
yfyeung: OK

PR #876, opened 2023-02-05
Get (start, end) timestamps for CTC models

yaozengwei:
This PR aims to support getting (start, end) timestamps of each word for CTC models.


yaozengwei:
You could see `test/test_parse_timestamp.py` for examples. Now we can get <start, end> frame indexes of each word.

yaozengwei:
Now in the function `parse_fsa_timestamps_and_texts` in `icefall/utils.py` supports two cases:
- ctc-decoding, the attribtutes `labels` and `aux_labels` are both BPE tokens. 
- HLG-based 1best, the attribtute `labels` is the prediction unit, e.g., phone or BPE tokens; attribute `aux_labels` is the word index.

It has been used in `conformer-ctc3/decode.py` with decoding methods of `ctc-decoding` and `1-best`. Taken `ctc-decoding` as example, the (start, end) timestamps (in seconds) of each words would be saved along with the predicted transcripts, like:
```
1089-134686-0000-459-0: hyp=['HE', 'HOPED', 'THERE', 'WOULD', 'BE', 'STEW', 'FOR', 'DINNER', 'TURNIPS', 'AND', 'CARROTS', 'AND', 'BRUISED', 'POTATOES', 'AND', 'FAT', 'MUTTON', 'PIECES', 'TO', 'BE', 'LADLED', 'OUT', 'IN', 'THICK', 'PEPPERED', 'FLUR', 'FATTENED', 'SAUCE']
1089-134686-0000-459-0: timestamp_hyp=[(0.640, 0.680), (0.840, 1.080), (1.160, 1.200), (1.280, 1.320), (1.440, 1.480), (1.760, 1.960), (2.000, 2.040), (2.120, 2.400), (2.760, 3.160), (3.240, 3.280), (3.520, 3.840), (3.920, 3.960), (4.160, 4.480), (4.560, 5.080), (5.200, 5.240), (5.440, 5.600), (5.840, 6.160), (6.240, 6.560), (6.680, 6.720), (6.760, 6.800), (7.040, 7.400), (7.400, 7.440), (7.720, 7.760), (8.000, 8.160), (8.400, 8.720), (8.960, 9.160), (9.360, 9.720), (9.920, 10.120)]
```

huangruizhe:
Question: how would you decide the end time of a word? In CTC model, we are able to get the start time of a word/token by the frame id of the labels or aux_labels. It looks like the position of the end time is less obvious to decide.

yaozengwei:
> Question: how would you decide the end time of a word? In CTC model, we are able to get the start time of a word/token by the frame id of the labels or aux_labels. It looks like the position of the end time is less obvious to decide.

The decoded fsa has the attributes: 
* token start time -> the arc whose `label` and `aux_label` are both non-epsilon. 
* token end time -> the last arc with non-epsilon (non-zero) `label` (i.e., the last repeating arc), before the start arc of next token. 

To get the (start, end) frame indexes pairs of each word: 
(1) If `aux_labels` are BPE tokens, we can utilize the start-token '_' to indentify the start of word, e.g., : 
* text: "HELLO WORLD"
* bpe tokens: ['_HE', 'LL', 'O', '_WORLD']
* bpe token ids: [22, 58, 24, 425]
* label: [22, 22, 58, 24, 0, 0, 425, 425, 425, 0, 0]
* aux_label: [22, 0, 58, 24, 0, 0, 425, 0, 0, 0, 0]
* (start, end) pairs: [(0, 3), (6, 8)]

(2) If `aux_labels` are word ids, the start of word is at the arc whose `label` and `aux_label` are both non-epsilon, e.g., : 
* text: "HELLO WORLD"
* bpe tokens: ['_HE', 'LL', 'O', '_WORLD']
* bpe token ids: [22, 58, 24, 425]
* label: [22, 22, 58, 24, 0, 0, 425, 425, 425, 0, 0]
* aux_label: [[79677], [], [], [], [], [], [196937], [], [], [], []]
* (start, end) pairs: [(0, 3), (6, 8)]

huangruizhe:
Got the idea! Thanks for the really nice explanation. That makes sense.

For CTC, the only concern is that the posterior can be peaky. That is, the repetition of [425, 425, 425 ] can be [425, 0, 0] instead. In this case, the end time of a word might be predicted earlier than the actual end time.
What do you think about this?

Here is an observation from my previous experiments on getting end time from the CTC model. 
The utterance here is "THANK YOU VERY MUCH"  ~  "▁THAN  K ▁YOU ▁VERY ▁MUCH"

The end time of "MUCH" got predicted too early, missing the last "CH" part as shown in the wav plot above.

If this issue can be or is already solved, that would be very useful.

yaozengwei:
> Got the idea! Thanks for the really nice explanation. That makes sense.
> 
> For CTC, the only concern is that the posterior can be peaky. That is, the repetition of [425, 425, 425 ] in the example could also be [425, 0, 0] instead. In this case, the end time of a word might be predicted earlier than the actual end time. What do you think about this?
> 
> Here is an observation from my previous experiments on getting end time from the CTC model. The utterance here is "THANK YOU VERY MUCH" ~ "▁THAN K ▁YOU ▁VERY ▁MUCH"  The end time of "MUCH" got predicted too early, missing the last "CH" part as shown in the wav plot above.
> 
> If this issue can be or is already solved, that would be very useful.

Thanks. The intent is not to align the audios, just to get decoding timestamps for CTC models, and show the symbol delay.

PR #879, opened 2023-02-06
Refactor onnx export for streaming zipformer

### LINKED ISSUE #682: ONNX exported model count
The ONNX exporting process currently exports five separate models:
  - ``encoder.onnx``
  - ``decoder.onnx``
  - ``joiner.onnx``
  - ``joiner_encoder_proj.onnx``
  - ``joiner_decoder_proj.onnx``

However, these models are just called sequentially, it doesn't look like there's much reason to keep them separate. It appears these models could be merged into just two .onnx models like so during the export:
* "encoder": `encoder`+`joiner_encoder_proj`
* "joiner": `decoder`+`joiner_decoder_proj`+`joiner`

Some sample code to illustrate what I mean:
```py
class OnnxEncoder(nn.Module):
    def __init__(self, encoder, joiner):
        super().__init__()
        self.encoder = encoder
        self.encoder_proj = joiner.encoder_proj

    def forward(self, x, h, c):
        x_lens = torch.tensor([9], dtype=torch.int64)
        x, _, new_states = self.encoder(x, x_lens, (h, c), 1.0)
        x = self.encoder_proj(x)

        return x, new_states[0], new_states[1]

class OnnxJoiner(nn.Module):
    def __init__(self, decoder, joiner):
        super().__init__()
        self.decoder = decoder
        self.joiner = joiner

    def forward(self, context, encoder_out):
        decoder_out = self.decoder(context, False)
        decoder_out = self.joiner.decoder_proj(decoder_out)

        joiner_out = self.joiner(encoder_out, decoder_out, False)

        return joiner_out
```

This could greatly reduce the amount of code and complexity required to perform inference, from loading and calling 5 models to just 2. Is there any reason they're separate and not merged like that?

csukuangfj:
Closes #682 

## TODOs
- [ ] Add documentation describing how to use it


csukuangfj:
It would be nice if someone can apply changes from this PR to
https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/pruned_transducer_stateless3/export.py

PR #882, opened 2023-02-07
Add generate_averaged_model.py

yfyeung:
Generate the averaged model from exp-dir.

Usage:
(1) use the checkpoint exp_dir/epoch-iter.pt
./pruned_transducer_stateless7/generate_averaged_model.py \
    --epoch 28 \
    --avg 15 \
    --exp-dir ./pruned_transducer_stateless7/exp

(2) use the checkpoint exp_dir/checkpoint-iter.pt
./pruned_transducer_stateless7/generate_averaged_model.py \
    --iter 22000 \
    --avg 5 \
    --exp-dir ./pruned_transducer_stateless7/exp

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless7/generate_averaged_model.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless7/generate_averaged_model.py
@@ -1,0 +1,200 @@
+#!/usr/bin/env python3
+#
+# Copyright 2021-2022 Xiaomi Corporation (Author: Fangjun Kuang,
+#                                                 Zengwei Yao,
+#                                                 Xiaoyu Yang)
+#
+# See ../../../../LICENSE for clarification regarding multiple authors
+#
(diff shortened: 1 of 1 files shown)


egs/librispeech/ASR/pruned_transducer_stateless7/generate_averaged_model.py
```diff
@@ -0,0 +1,199 @@
+#!/usr/bin/env python3
+#
+# Copyright 2021-2022 Xiaomi Corporation (Author: Fangjun Kuang,
```
csukuangfj, line 3: please change it to yours.

yfyeung: OK

csukuangfj (approved):
Thanks!

danpovey:
I think it would be better if it had the --use-averaged-model option, so that it can do the averaged-over-checkpoints version like the decode script.

PR #890, opened 2023-02-08
Librispeech, validate_manifest.py

vesis84:
- modifying the validation script
   - the Supervision start is relative to Cut
   - let's assume the data is prepared with 'trim-to-supervision' (i.e. Supervision.begin is close to 0.0)
   - the end of supervision should not fall out of cut
   - a tolerance of 2ms is itroduced
   - finally, the validation 'validate_for_asr' from 'lhotse.dataset.speech_recognition' is performed

- this script will be generic also for imported kaldi data folders (for both types 'one segment per wav' and 'many segments per wav')
   - actually I am building a generic recipe for training from imported kaldi data folders
   - would it be interesting to contribute it to icefall ?

### DIFF
--- a/egs/librispeech/ASR/local/validate_manifest.py
+++ b/egs/librispeech/ASR/local/validate_manifest.py
@@ -35,6 +35,7 @@ from pathlib import Path
 
 from lhotse import CutSet, load_manifest_lazy
 from lhotse.cut import Cut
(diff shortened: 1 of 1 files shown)


csukuangfj:
Thanks for your contribution!

vesis84:
Thank you for merging!

PR #895, opened 2023-02-09
fix mmi graph compiler bug.

emilyluj:
Fix one mmi graph compiler issue.


csukuangfj:
Thanks for your contribution!

PR #907, opened 2023-02-14
Conformer CTC & RNN-T with delay penalty

yaozengwei:
This PR aims to implement streaming Conformer with CTC loss and RNN-T loss, where both losses support delay penalty. 
It is built based on `pruned_transducer_stateless4`.
- For RNN-T branch, it supports getting start timestamps for each decoded word, and calculating symbol delay. 
- For CTC branch, it supports getting (start, end) timestamps for each decoded word, and calculating symbol delay.


yfyeung (approved):

egs/librispeech/ASR/pruned_transducer_stateless4_ctc/train.py
```diff
@@ -0,0 +1,1228 @@
...
+    valid_dl: torch.utils.data.DataLoader,
+    world_size: int = 1,
+) -> MetricsTracker:
+    """Run the validation process."""
+    model.eval()
+
+    tot_loss = MetricsTracker()
+
+    for batch_idx, batch in enumerate(valid_dl):
+        loss, loss_info = compute_loss(
```
pkufool, line 785: It is better to add  ` with torch.cuda.amp.autocast(enabled=params.use_fp16):` here too.

egs/librispeech/ASR/pruned_transducer_stateless4_ctc/train.py
```diff
@@ -0,0 +1,1228 @@
...
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        supervision_segments, token_ids = encode_supervisions(
+            supervisions,
+            subsampling_factor=params.subsampling_factor,
+            token_ids=token_ids,
+        )
+
+    # Works with a BPE model
+    decoding_graph = k2.ctc_graph(token_ids, modified=False, device=device)
```
pkufool, line 712: Add `with torch.cuda.amp.autocast(enabled=False):` here,  we should skip fp16 when calculating ctc loss, because `k2.DenseFsaVec` requires `ctc_output` to be `torch.float32`. 

It is safer to construct `k2.DenseFsaVec` as follows (adding `.float()` to ctc_output).  It works well now because the last operation in `get_ctc_output` is `LogSoftmax` which does not support fp16. 
```
dense_fsa_vec = k2.DenseFsaVec(
        ctc_output.float(),
        supervision_segments,
        allow_truncate=params.subsampling_factor - 1,
    )
```

yaozengwei: Thanks. Fixed that.

PR #936, opened 2023-03-08
remove key from result filename

pehonnet:
This results in shorter filenames (too long file can cause errors).


csukuangfj:
Thanks for your contribution!

Issue #920, opened 2023-02-22
LM shallow fusion and LODR in aishell


li563042811:
Hi,
    When will you support modified beam search with LM shallow fusion and LODR in Chinese corpus, such as aishell?

csukuangfj:
> Hi, When will you support modified beam search with LM shallow fusion and LODR in Chinese corpus, such as aishell?

We already have the code. I think it can be reused for aishell.

li563042811:
For librispeech the rnnlm can be trained with bpe dictionary. But for aishell, the best is trained in char. 
The rnnlm training code in https://github.com/k2-fsa/icefall/tree/master/icefall/rnn_lm is for bpe and it has some problem to train a rnnlm with the same token table with a well-trained asr model.

marcoyang1998:
> When will you support modified beam search with LM shallow fusion and LODR in Chinese corpus, such as aishell?

This is on our schedule.

The first step will be preparing training data in char. Once we figure out the best data format, we can use the RNNLM training code.

I will update here as soon as we make some progress.

li563042811:
Thank you for your reply.
I also tried to break the bind between token tables of the ASR model and LM model. What I do is that mapping the token_list to char_list with ASR token_table and then mapping the char_list to token_list with LM token_table.  After this process, the token_list can be used for LM inference.

li563042811:
Thank you for your rely.
I also tried to break the bind between token_tables of ASR and LM.

li563042811:
Thank you for getting back to me.
I also tried to break the bind between token_tables of ASR and LM so that the LM training can be more flexible.
I map the token_list to char_list with ASR token_table, then map the char_list to token_list with LM token_table so that the token_list can be used for LM inferring.
But this method will increase a lot of time-consuming decoding. So I think maybe training LM with the same token_table still is the best way for LM fusion.

li563042811:
I modified the generation process of the Chinese language model training corpus. Instead of using the sentencepiece model for mapping, I changed it to use the specified tokens.txt for char-level mapping. However aishell has too little training text, and the RNNLM model tends to overfit. The ppl of the trained RNN language model on the test set corpus is around 70, and the shallow fusion does not reduce the CER. However, if more training corpus(aishell2) is added, the ppl will decrease significantly, and the reduction in CER will be pronounced after shallow fusion(9% relative decrease).

marcoyang1998:
That's great! 

[...]

PR #945, opened 2023-03-14
Add char-based language model training process for aishell.

li563042811:
Add prepare_char_lm_training_data.py in local/ for char-based lm train data preparation process.
Add steps 9,10,11 in prepare.sh for lm data preparation and RNN lm training.
https://github.com/k2-fsa/icefall/issues/920

### DIFF
--- a/egs/aishell/ASR/prepare.sh
+++ b/egs/aishell/ASR/prepare.sh
@@ -7,7 +7,7 @@ set -eou pipefail
 
 nj=15
 stage=-1
-stop_stage=10
+stop_stage=11
 
 # We assume dl_dir (download dir) contains the following
 # directories and files. If not, they will be downloaded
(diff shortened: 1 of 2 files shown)


marcoyang1998 (approved):

marcoyang1998:
Thanks for your contribution!

Do you have any WERs to report with LM fusion? It would be great to have some numbers here.

egs/aishell/ASR/prepare.sh
```diff
@@ -219,3 +219,91 @@ if [ $stage -le 8 ] && [ $stop_stage -ge 8 ]; then
...
+  log "Stage 10: Sort LM training data"
+  # Sort LM training data by sentence length in descending order
+  # for ease of training.
+  #
+  # Sentence length equals to the number of tokens
+  # in a sentence.
+
+  out_dir=data/lm_training_char
+  mkdir -p $out_dir
+  ./local/sort_lm_training_data.py \
```
marcoyang1998, line 277: Could you add a soft link for this file? I don't think it's present under `local`

marcoyang1998 (changes requested):

egs/aishell/ASR/prepare.sh
li563042811: Ok, I will add it as soon as possible.

marcoyang1998 (approved):

marcoyang1998:
LGTM!

Merging it now.

minmin0:
good job! Can we use SF in sherpa-onnx now? I found kaldi and  wenet already support LM in C++  runtime, but in sherpa-onnx or sherpa-ncnn  i don't know how to use extern LM.

marcoyang1998:
I would recommend using RNNLM rescoring in #1002 in Sherpa as it is much faster than shallow fusion while yielding similar WERs. This is on our to-do list.

minmin0:
> RNNLM

Thanks, but the decoding code in icefall is not suitable for use in the production environment.I think sherp-onnx maybe one option, but unfortunately, it is not supported LM.

csukuangfj:
> > RNNLM
> 
> Thanks, but the decoding code in icefall is not suitable for use in the production environment.I think sherp-onnx maybe one option, but unfortunately, it is not supported LM.

@minmin0 Can you help to test it in sherpa-onnx if it is supported in sherpa-onnx?

minmin0:
> onnx

yes

minmin0:
> > > RNNLM
> > 
> > 
> > Thanks, but the decoding code in icefall is not suitable for use in the production environment.I think sherp-onnx maybe one option, but unfortunately, it is not supported LM.
> 
> @minmin0 Can you help to test it in sherpa-onnx if it is supported in sherpa-onnx?

yes， I am willing to do  that

PR #958, opened 2023-03-22
Modify ScaledAdam for parameter quantization

yaozengwei:
This PR aims to modify the ScaledAdam optimizer to make parameter quantization work better. 
The idea is to prevent absolute-values of any given weight tensor from being over a certain percentile of the distribution of that parameter tensor's absolute values, e.g. over 80% or 90% or 95%.  (This will, in effect, cause the distribution of absolute values to end in a peak).
We do this randomly with a probability, say, `p_limit_values=0.1`, for each batch of parameter tensors (ScaledAdam batches the parameters of the same shape for speed).

The algorithm is as follows:
```
Initialize the optimizer state with some default/guessed 90th percentile value, 1.0, 
say, `stored_percentile`, this is not critical.

On each iteration that we do this type of update:
If >10% of the parameter abs() value are greater than our `stored_percentile`:
    Increase the stored_percentile by 1.0 + learning rate / p_limit_values.
Else:
    Decrease the stored_percentile by 1.0 + learning rate / p_limit_values. 
    (but don't let it go t o zero, e.g. floor to 1.0e-20)

For all parameters whose abs() values are greater than the stored_percentile: 
    we modify the sign of the update for this iteration so that it points 'inward'.
```

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless7/optim.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless7/optim.py
@@ -164,11 +164,9 @@ class ScaledAdam(BatchedOptimizer):
                    of the parameter tensor.  This is provided to save a little time
                    in the update.
      clipping_update_period: if clipping_scale is specified, this is the period
-    p_limit_values: The probability (e.g., 0.1) to modify the update sign so as to prevent
-                    absolute-values of any weight tensor from being over a certain percentile of
-                    the distribution of that parameter tensor's absolute values.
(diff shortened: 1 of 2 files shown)


danpovey:
My feeling is that this may not be a great idea, it seems to harm training more than I expected and this is with the addition of an extra condition "p_abs > 10" so that it only applies to the largest parameter tensors.   Let's revisit this kind of thing later.  For example we could try periodically (quite rarely) truncating parameter values to their 99th percentile absolute-value within each tensor.

PR #965, opened 2023-03-24
remove if-branch at downsample pad in zipformer for onnx-export compatibility

kamirdin:
this if-branch can't trace in onnx-exporting when we use tensor which T-dim is an even number , throw RuntimeError when we run the onnx model :
> The input tensor cannot be reshaped to the requested shape. Input shape:{219,1,384}, requested shape:{110,2,1,384}

even though it can export successfully without any error

we can allow he number of pads to be 0 to make this an unconditional execution.


yaozengwei (approved):

yaozengwei:
Thanks.

PR #975, opened 2023-03-30
add option of using full attention for streaming model decoding

yaozengwei:
For the streaming conformer models trained with dynamic chunk-wise attention mask, we can validate its performance with full attention using:
```
./pruned_transducer_stateless4/decode.py \
    --epoch 17 \
    --avg 5 \
    --simulate-streaming 1 \
    --causal-convolution 1 \
    --decode-chunk-size -1 \
    --left-context -1 \
    --exp-dir pruned_transducer_stateless4/exp \
    --max-duration 600 \
    --decoding-method greedy_search
```

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless/decode.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless/decode.py
@@ -107,6 +107,7 @@ Usage:
 
 import argparse
 import logging
+import math
 from collections import defaultdict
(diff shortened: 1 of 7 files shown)


yaozengwei:
For the non-streaming Zipformer in pruned7, it use full attention. For the streaming version in pruned7_streaming, it did not support full attention for decoding. But it is easy to modify that similarily.

PR #980, opened 2023-04-01
Support long audios recognition

yaozengwei:
This PR aims to support long audios recognition. The process is as follows (see `long_file_recog.sh` for detail): 
1) Split long audios into chunks with overlaps.
2) Perform speech recognition on chunks, getting tokens and timestamps.
3) Merge the overlapped chunks into utterances acording to the timestamps.


desh2608:
> BTW just wanted to mention that I tried this method for decoding full GigaSpeech dev and test sets, and got the following results:
> 
> Model	DEV	TEST
> Conformer-transducer greedy	10.53 / 12.16	10.74 / 13.20
> Conformer-transducer beam search	10.43 / 11.95	10.50 / 12.81
> The WERs are <segment / full>. I used the pre-trained pruned_transducer_stateless2 model.

@danpovey Here are the WERs.

egs/librispeech/ASR/long_file_recog/merge_chunks.py
```diff
@@ -0,0 +1,181 @@
...
+            cur_end = right
+
+            assert len(cut.supervisions) == 1, len(cut.supervisions)
+            for ali in cut.supervisions[0].alignment["symbol"]:
+                t = ali.start + cut.start
+                if left <= t < right:
+                    alignments.append(ali.with_offset(cut.start))
+
+        old_sup = supervisions[rec.id]
+        assert old_sup.recording_id == rec.id, (old_sup.recording_id, rec.id)
```
desh2608, line 113: Given the token-level alignments, if I want to get corresponding word-level alignments (for example, to create a CTM file), is there a way to do that? At the moment, I get the full output text as:
```python
        # Decode the BPE tokens to text
        hyp = [ali.symbol for ali in alignments]
        hyp_text = sp.decode(hyp)
        hyp_text = asr_text_post_processing(hyp_text)
```
Basically, the problems comes down to knowing the mapping that `sp.decode` creates, but I don't know if this mapping is accessible.

desh2608: Tagging @csukuangfj since you may be familiar with sentencepiece.

desh2608: Found this relevant function: https://github.com/k2-fsa/icefall/blob/30bde4b788f0ad05110f6981b630b745ce6ffb82/egs/librispeech/ASR/transducer_stateless/alignment.py#L239

yaozengwei: You could use this https://github.com/k2-fsa/icefall/blob/30bde4b788f0ad05110f6981b630b745ce6ffb82/icefall/utils.py#L1392. It would cover more cases.

desh2608: @yaozengwei Thanks! This is exactly what I needed :)

PR #983, opened 2023-04-03
Remove simulate streaming from stateless7

yfyeung:


yaozengwei (dismissed):
You could also remove the args "--decode-chunk-size" and "--left-context"

yfyeung:
> You could also remove the args "--decode-chunk-size" and "--left-context"

Ok, I have removed them.

yaozengwei (approved):

PR #986, opened 2023-04-05
Training with byte level BPE (AIShell)

pkufool:
This PR tries to use the byte level BPE as modeling uints, the main purpose is to reduce the vocab-size by spliting CJK characters, so as to speed up the training.

To find out how we use byte level BPE, the best way is to read the code in egs/aishell/ASR/local/train_bbpe_model.py


pkufool:
The best results I have gotten are:

Vocab size | Greedy search(dev & test) | Modified beam search(dev & test) | Fast beam search (dev & test)  | Fast beam search LG (dev & test) | comments
-- | -- | -- | -- | -- | --
500 | 4.31 & 4.59 | 4.25 & 4.54 | 4.27 & 4.55 |  4.07 & 4.38 | --epoch 48 --avg 29
1000 | 4.94 & 5.31 | 4.73 & 5.11 | 4.81 & 5.17 |    | --epoch 50 --avg 36

The training command:

```
export CUDA_VISIBLE_DEVICES="4,5,6,7"

./pruned_transducer_stateless7_bbpe/train.py \
  --world-size 4 \
  --num-epochs 50 \
  --start-epoch 1 \
  --use-fp16 1 \
  --max-duration 800 \
  --bbpe-model data/lang_bbpe_500/bbpe.model \
  --exp-dir pruned_transducer_stateless7_bbpe/exp \
  --lr-epochs 6 \
  --master-port 12535
```

The decoding command:

```
for m in greedy_search modified_beam_search fast_beam_search fast_beam_search_LG; do 
    ./pruned_transducer_stateless7_bbpe/decode.py \                                                                                                                                                                                              
      --epoch 48 \                                                                                                                                                                                                                         
      --avg 29 \                                                                                                                                                                                                                           
      --exp-dir ./pruned_transducer_stateless7_bbpe/exp \                                                                                                                                                                                  
      --max-sym-per-frame 1 \                                                                                                                                                                                                              
      --ngram-lm-scale 0.25 \                                                                                                                                                                                                            
      --ilme-scale 0.2 \                                                                                                                                                                                                                 
      --bbpe-model data/lang_bbpe_500/bbpe.model \                                                                                                                                                                                         
      --max-duration 2000 \                                                                                                                                                                                                                
      --decoding-method $m
done
```

egs/aishell/ASR/pruned_transducer_stateless7_bbpe/export.py
```diff
@@ -0,0 +1,320 @@
...
+        "--exp-dir",
+        type=str,
+        default="pruned_transducer_stateless7/exp",
+        help="""It specifies the directory where all training related
+        files, e.g., checkpoints, log, etc, are saved
+        """,
+    )
+
+    parser.add_argument(
+        "--bbpe-model",
```
marcoyang1998, line 158: Can we use `--bpe-model` here? Then we could use a soft link for this file?

marcoyang1998: This also applies to a couple of other files.

egs/aishell/ASR/pruned_transducer_stateless7_bbpe/onnx_check.py
marcoyang1998, line 1: Can we use a soft link for this file?

pkufool:
A pretrained model is available here: https://huggingface.co/pkufool/icefall_asr_aishell_pruned_transducer_stateless7_bbpe

yaozengwei (approved):

pkufool:
I am going to merge this PR, so as to be more convenient to create tal_csasr and wenetspeech recipes, if you guys have any suggestions will fix in the seperate PR.

egs/aishell/ASR/local/train_bbpe_model.py
```diff
@@ -0,0 +1,112 @@
...
+    if not model_file.is_file():
+        spm.SentencePieceTrainer.train(
+            input=train_text,
+            vocab_size=vocab_size,
+            model_type=model_type,
+            model_prefix=model_prefix,
+            input_sentence_size=input_sentence_size,
+            character_coverage=character_coverage,
+            user_defined_symbols=user_defined_symbols,
+            unk_id=unk_id,
```
kobenaxie, line 101: hi @pkufool , if using byte level bpe, maybe need not <unk_id> anymore ?

armusc:
Hi

I'm probably asking a naive question, but here it is. I'm developing a Korean ASR and I'd like to try with the bbpe approach because the number of different characters in the training corpus is quite high, anyway higher than what I'd have in other latin-char based languages, something like 1300 characters

THe thing is in 
tokenize_by_CJK_char
we split all the CJK chars while we leave unsplitted the words written in latin characters (like the example in the function code:
    Example:
      input = "你好世界是 hello world 的中文"
      output = "你 好 世 界 是 HELLO WORLD 的 中 文"
)
by doing this, after training the bbpe model with sentencepiece, I have a lexicon like this:

가로지르는 ▁ƑŖġ ▁ƒņŁ ▁Ɠōġ ▁ƒŋ Ś ▁ƒīķ
가로축 ▁ƑŖġ ▁ƒņŁ ▁ƓŜ ĸ
가루 ▁ƑŖġ ▁ƒňŎ
가루를 ▁ƑŖġ ▁ƒňŎ ▁ƒŋŢ

and basically all tokens starting with ▁

because all the Korean words were splitted in single CJK chars, so the word boundary info would be lost for the sentencepiece model (and each word would be tokenized by "word-beginning" tokens)
is that the expected training unit? because then in decoding I'm not sure how words would be reconstructed, or are we supposing to decode isolated chars and evaluate by CER?

otherwise we could simply avoid splitting the CJK, and we could have a lexicon like:
가게를 ▁ƑŖġ ƑŘĭ ƒŋŢ
가게에서 ▁ƑŖġ ƑŘĭ ƓĺıƓĥŁ
가게에선 ▁ƑŖġ ƑŘĭ Ɠĺı ƓĥŅ
가겠 ▁ƑŖġ ƑŘ Ņ
가겠구나 ▁ƑŖġ ƑŘ Ņ ƑśŒ ƒģĻ

and only 142 out of 500 tokens as "word-beginning" tokens

I can see that my understanding of this approach is possibly quite limited, thanks in advance if you can help me a bit in understanding

pkufool:
Actually, we already tried kinds of splitting,  1) splitting by single char 2) splitting by word (for Chinese might be several chars) 3) no splitting.  From our experiments, the performance is  1 > 2 > 3, but difference is minor. We finally choose "splitting by single char" because we don't want one bpe have two bytes that belong to different chars (like English, one bpe won't have two ascii chars that belong to two different words).

pkufool:
> I'm not sure how words would be reconstructed, or are we supposing to decode isolated chars and evaluate by CER?

When decoding, the decoded sequence would be convert to byte sequence first, then decode to CJK char words.
Yes, we evaluate by CER on AIshell. but for bilingual model, we evaluate it by MER (WER for english, CER for Chinese).

armusc:
thanks for the anwser

PR #990, opened 2023-04-09
Add averaged model && change start from 0 to 1 && fix typo for gigaspeech

yfyeung:
1. Add support for use-averaged-model
2. Change start-epoch from 0 to 1
3. Fix some typo


yaozengwei:
This might cause conflicts when loading uploaded checkpoints without `model_avg`. But it is ok using `--use-averaged-model=False`.  I think we can merge it first. @csukuangfj

yaozengwei (approved):

yfyeung:
> This might cause conflicts when loading uploaded checkpoints without `model_avg`. But it is ok using `--use-averaged-model=False`. I think we can merge it first. @csukuangfj

Yes, I have tried to load one without model-avg using this PR.

PR #1005, opened 2023-04-17
Fix torch.jit.script() export for streaming zipformer.

csukuangfj:
It was causing errors for `jit_trace_export.py`.

image


whaozl:
it's ok, Thank you very much, I update this method.

PR #1013, opened 2023-04-20
Add compute_ppl.py and ngram_entropy_pruning.py

yfyeung:
`compute_ppl.py` can compute the perplexity of the giving ngram LM on the specify dataset, usage:
```shell
./pruned_transducer_stateless7/compute_ppl.py \
    --ngram-lm-path ./download/lm/3gram_pruned_1e7.arpa
```

`ngram_entropy_pruning` is from kalid `egs/wsj/s5/utils/lang/ngram_entropy_pruning.py`, which is an implementation of ``Entropy-based Pruning of Backoff Language Models''
in the same way as SRILM, usage:
```shell
./ngram_entropy_pruning.py \
    -threshold 1e-8 \
    -lm download/lm/4gram.arpa \
    -write-lm download/lm/4gram_pruned_1e8.arpa
```


csukuangfj:
Thanks!

PR #1027, opened 2023-04-28
MUCS recipe

bloodraven66:
Hi,
This is a recipe for [MUCS](https://navana-tech.github.io/MUCS2021/data.html) dataset. 
I've taken librispeech conformer_ctc and modified it for this corpus.

Currently, on hi-en corpus, it obtains 28.37 WER. The espnet baseline reported in [leaderboard](https://navana-tech.github.io/MUCS2021/leaderboard.html) is 25.53. I've used LM with sentences from training, and tokens as BPE 200.
In the next few days, I'll clean up the codes, and add results for other datasets part of MUCS.

Meanwhile, if anyone can suggest best practices to improve the performance with Icefall, that would be helpful.


desh2608:
Thanks. If you have added some *.py files directly from another recipe, it is better to symlink it than copy as a new file.

bloodraven66:
Sure, I believe I have made small changes in all files, changing the names of cuts, etc. Will replace with symlink when possible

desh2608:
Cool. Usually we only change train.py and asr_datamodule.py. Of course, the files in local/ don't have to be symlinked since they are "local" to the recipe.

bloodraven66:
symlinked the required files, and modified it to support both code switched datasets part of MUCS, here are the results with training for 60 epcohs, decoded with ep50-59 chk average, num_paths=20 

<google-sheets-html-origin>

  | ctc decoding | 1best | nbest | nbest-rescoring | whole-lattice-rescoring | attention-decoder
-- | -- | -- | -- | -- | -- | --
bn-en | 31.72 | 28.05 | 27.92 | 27.22 | 27.24 | 26.46
hi-en | 31.43 | 28.48 |   |   |   | 28.16

csukuangfj:
@bloodraven66 

Thanks for your contribution!

Could you follow the librispeech recipe to create a RESULTS.md to document your results and describe where to find the pre-trained models, training logs, and decoding results?

csukuangfj:
> @bloodraven66
> 
> Thanks for your contribution!
> 
> Could you follow the librispeech recipe to create a RESULTS.md to document your results and describe where to find the pre-trained models, training logs, and decoding results?

@bloodraven66  Do you have time to update the results?

egs/mucs/ASR/local/compile_hlg.py
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/env python3
```
csukuangfj, line 1: Could you replace it with a symlink to the one in the librispeech recipe?

csukuangfj:
Thanks! Left a minor comment. Otherwise looks good to me.

Could you please use `export.py` to export your pretrained model and upload it to huggingface? You can find that in all other RESULTS.md we have uploaded pretrained models for each experiment to huggingface. The intention is that other users can easily use your pre-trained model for inference without re-training.

---

> Meanwhile, if anyone can suggest best practices to improve the performance with Icefall, that would be helpful.

I suggest that you try our latest [zipformer](https://github.com/k2-fsa/icefall/tree/master/egs/librispeech/ASR/zipformer), which is the best model we have so far.

The training commands can be found at
https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/RESULTS.md#zipformer-zipformer--pruned-stateless-transducer

From our past experience, transducer-based models produce lower WER than ctc-based models.

PR #1028, opened 2023-04-28
Fix typo in librispeech OpenFST-based HLG preparation script

sailordiary:
`compile_hlg_using_openfst.py` expects `L_disambig.fst` instead of `disambig_L.fst`.


csukuangfj:
Thanks!

PR #1038, opened 2023-05-06
Add Context biasing

pkufool:
This is a draft version, the documents might not be clear.
I am implementing a version using Aho-Corasick automata (the history states managment is a little complicate in current implementation based on openfst).


pkufool:
This the results of my self-constructed test set (test-book),  WERs decrease from 11.16 & 5.95 to 3.9 & 3.88 without affecting the WERs of test-clean & test-other.

image

Also one of the inner test set for wenetspeech dataset, CERs decrease from 16.37 to 10.74.

image

Need more testing, but I think it is promising.

pkufool:
I re-implement the ContextGraph with an Aho-Corasick, the performance is the same as the openfst one, but it is much more efficient, here are the decoding time for whether using context biasing.

Biasing method | Decode time (second) for test-clean(libri) | overhead | Decode time (second) for test-net (wenetspeech) | overhead
-- | -- | -- | -- | --
No Biasing | 222 | - | 1467 | -
Aho-Corasick | 224 | Almost zero | 1437 | smaller?(might be noise)
Openfst | 328 | 46% | 1794 | 22%

pkufool:
Although there are still some code for my own experiments, you can have a look first.

pkufool:
@huangruizhe  Could you try this Aho-Corasick implementation on your test-sets?

dophist:
Great work, Aho-Corasick is the right call, and this might be the first "correct" design of phrase-based context biasing in open-source domain.

Just one reminder about the implementation: real-life context biasing can be dynamic, where phrases need to be added from context to context. Typical finite state machine implementation (such as OpenFST's VectorFST, where states are stored continuously in memory) is not suitable for such senarios.

You guys may consider a hash table implementation, coupled with an incremental hash function for ngrams, like [KenLM's incremental hash](https://github.com/kpu/kenlm/blob/4e6ac85c8d01ac91cb61dfbdc76cd652158c5969/lm/search_hashed.hh#L25).

The idea is, Aho-Corasick algorithm can be seen as a sparse NGram LM(where N = infinity),  and failure arcs are just NGram's backoff(except it can step back more than one token due to the sparsity).  This not only adds the flexibility to add phrases efficiently, but also simplifies the implementation (without any FST framework dependency, and the entire graph is nothing more than a global hash-table of ngrams). The tradeoff made here is that hash-table lookup might not be necessarily faster than the binary search of a trie, especially when the context graph is tiny. But when the graph becomes large, hash-table is more stable. Again, [KenLM's probing hash table](https://github.com/kpu/kenlm/blob/master/util/probing_hash_table.hh) is a good reference. 

Just sharing this thought, actually this is more relavent to K2's C++ runtime implementation, not particularly to this python PR.

Looking forward to this new feature, cheers.

icefall/context_graph.py
```diff
@@ -14,243 +14,209 @@
...
+        Returns:
+          Return a tuple of score and next state.
+        """
+        # token matched
+        if token in state.next:
+            node = state.next[token]
+            score = node.score
+            # if the matched node is the end of a word/phrase, we will start
+            # from the root for next token.
+            if node.is_end:
```
kobenaxie, line 141: `return self.root` only happens when `node` is a leaf, here should be `if node.is_end and node.next == {}` ?
otherwise, if there are two phrases `周杰` and `周杰伦`, it seems `周杰伦` can not be boost ?

icefall/context_graph.py
```diff
@@ -0,0 +1,222 @@
...
+                    node.next[token] = ContextState(
+                        token=token,
+                        score=self.context_score,
+                        # The total score is the accumulated score from root to current node,
+                        # it will be used to calculate the score of fail arc later.
+                        total_score=node.total_score + self.context_score,
+                        is_end=i == len(tokens) - 1,
+                    )
+                node = node.next[token]
+            self._fill_fail()
```
dophist, line 119: @pkufool I think the building of failure arcs should be global, not per-phrase basis. The idea is to scan the entire list, figuring out the "proper suffix" relations between all trie nodes. If the failure arcs are built per-phrase basis, in the middle of the graph building, existing nodes can not see nodes from unseen(not-built-yet) phrases.

icefall/context_graph.py
```diff
@@ -0,0 +1,222 @@
...
+        Returns:
+          Return a tuple of score and next state.
+        """
+        # token matched
+        if token in state.next:
+            node = state.next[token]
+            score = node.score
+            # if the matched node is the end of a word/phrase, we will start
+            # from the root for next token.
+            if node.is_end:
```
dophist, line 141: This seems not right, the search set (phrase list) may contain a string A(say "ab") that is a prefix of a string B(say "abc"). Even the 'b' of "ab" is end(algorithmically, we often call this an "output node" instead of "end", which is misleading), the match should be able to continue along the path. Actually, the correct matching behavior should follow "either a match or a failure arc" strictly. I'll recommand some links as references in the main thread later.

dophist:
Just had a skim of this implementation, there possibly might be some misunderstanding of the Aho–Corasick algorithm. I recommand a paper [here](https://se.inf.ethz.ch/~meyer/publications/string/string_matching.pdf), the author gave comprehensive seudo code and explicitly explained all invariants of the algorithms, it is pretty old-school and exceedingly well-written. 

The algorithm is non-trivial, not easy to get it implemented correctly, I hope the above paper may be a helpful resource.

icefall/context_graph.py
pkufool: Oh, yes, the indent is wrong, it is very strange that such a error did not affect the performance. Thanks!

icefall/context_graph.py
pkufool: Yes, I know that the Aho-Corasick does not match this way,  my current implementation can match shortest phrase if some phrases have common prefixs. I tried the original matching method, but it gets worse performance. I suspect it is because the wrong ``fill_fail``. Will re-run the experiments.

pkufool:
> Just had a skim of this implementation, there possibly might be some misunderstanding of the Aho–Corasick algorithm. I recommand a paper [here](https://se.inf.ethz.ch/~meyer/publications/string/string_matching.pdf), the author gave comprehensive seudo code and explicitly explained all invariants of the algorithms, it is pretty old-school and exceedingly well-written.
> 
> The algorithm is non-trivial, not easy to get it implemented correctly, I hope the above paper may be a helpful resource.

Thanks for your comments, I think I did not misunderstanding the Aho-Corasick, I was too lazy to do more complex tests and so made such foolish mistakes (I mean the wrong fill_fail). Thanks again！ I will read the paper to improve my algorithm.

dophist:
No worries, it's just not a straight forward algorithm to get it 100% correct for the first try : )

icefall/context_graph.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        queue = []
+        for token, node in self.root.next.items():
+            node.fail = self.root
+            queue.append(node)
+        while queue:
+            current_node = queue.pop(0)
+            current_fail = current_node.fail
+            for token, node in current_node.next.items():
+                fail = current_fail
```
dophist, line 86: Another possible pitfall, my feeling is that there should be some sort of recursive(or looping) fall-back here. For example, to fill the failure arc of C, you might look at its parent B, and possibly also B's failure node A, and so on...

pkufool: It is strange the experiment results keep the same after fixing the `fill_fail`. I think the phrases are complex enough (it has 800 phrases) to make some difference. There indeed are some bugs in the code, I will try visualizing the constructed graph tomorrow.

pkufool: > Another possible pitfall, my feeling is that there should be some sort of recursive(or looping) fall-back here. For example, to fill the failure arc of C, you might look at its parent B, and possibly also B's failure node A, and so on...

Yes, you are right.

icefall/context_graph.py
pkufool: It is expected to match both phrases like ("ab", "abc"),  the difficulty is tracking the scores, will first make sure the graph constructed is correct, than let's consider the searching.

pkufool:
I think the graph is working as expected now.
Here is the graph for  "HE / SHE / SHELL / HIS / HERS / HELLO" , the red arcs are the fail arcs.

image

Here is the test code:
**Note:** If two contexts sharing same suffix, the longer one will be matched, for example, if a query contains "SHE" it will match "SHE" rather than "HE".

```python
    queries = ["HERSHE", "HISHE", "SHED", "HELL", "HELLO", "DHRHISQ"]
    # HERSHE mathes "HERS" and "SHE", gets 7
    # HISHE matches "HIS" and "SHE", gets 6
    # SHED matches "SHE", gets 3
    # HELL matches "HE", get 2
    # "HELLO" matches "HELLO", get 5
    # "DHRHISQ" matches "HIS", get 3
    expected_scores = [7, 6, 3, 2, 5, 3]
    for i, query in enumerate(queries):
        total_scores = 0
        state = context_graph.root
        for q in query:
            score, state = context_graph.forward_one_step(state, ord(q))
            total_scores += score
        score, state = context_graph.finalize(state)
        assert state.token == -1, state.token
        total_scores += score
        assert total_scores == expected_scores[i], (
            total_scores,
            expected_scores[i],
            query,
        )
```

dophist:
Great progress.

I have a major concern after a quick skim of the improved implementation (I might be wrong, just start the discussion to make sure everything is by design).

Let's start with an example:
Say you have a search set: { "BC",  "ABCDE" }, and rescoring hypothesis starts with "ABCDF...", I guess "BC" will be shadowed by "ABCDE", the entire match will yields nothing in the end. In practice, this will result to zombie entries(that can never take effect) in your phrase list.
I find a 5-min video that may be helpful to understand the pitfall. [hot words search shadowing](https://www.youtube.com/watch?v=OFKxWFew_L0)

As developers from other speech communities might have said, "oh there should be only one active path during the search", well that doesn't make sense to me at all. The root cause for this is: Theoretically there are 3 major components in Aho-Corasick algorithm, this PR now only implements two of them:
1. The "Goto Function" (i.e. as you called "forward_one_step")
2. The "Failure Function" (as you called the "fail" arc)
3. The "Output Table" (which is missing)

The key to understand this is that there are multiple parallel paths during the search, and the search head essentially encapsulates two linked list:
a. one encodes the frontier nodes of the searched region
b. the other encodes a list of hits(phrases) at current location (i.e. the Output Table). List b is a subset of list a.

Since output table can contain multiple entries, this implies the biasing weights should be placed on end nodes of these hits, not on the single arc that the search head just traversed.

My explanation is far from clear compared to the paper I recommanded above.

Another suggestion I would give is that, don't rely too much on the numbers from your current test set (which contains only a few hundreds of phrases). In practice, you may have giant list containing tens of thousands of musics, company names, POIs, and the correctness of the implementation may be crucial to get certain entries correctly recognized.

pkufool:
Em, yes! "search shadowing" is a problem, will try to fix this issue.

> Since output table can contain multiple entries, this implies the biasing weights should be placed on end nodes of these hits, not on the single arc that the search head just traversed.

I put the score on the single arc rather than the end node of a phrase, because we need to output the score for each step to make the "prefix-matched" path survive during beam search, if a phrase is very long, the corresponding path might be pruned output before reaching the end node of the phrase (I did not have any experiment result, though, will try it).

dophist:
Yeah, typically we want to avoid applying the biasing score too early, which may be detrimental to base model behavior, ideally we want to place the score at the end of each phrases. On the other hand we also want to avoid phrases falling out of the decoder beam. For such consideration, i think there is a simple yet elegant solution, say here is the phrase-registration format: 
```
# end weight
the old man and the sea:5.0
# distributed weights
the old man and:1.0 the:1.0 sea:3.0
```
It enforces specific prefix requirements for biasing, and minimizes human interferences to the base model. And at the same time, outside users can freely control how to distribute the weights if falling-out-of-beam is a concern.

In this design, the behavior of aho-corasick algorithm provides a solid machnism that can be clearly articulated in single sentence: "a match is a match, a biasing score will be enforced", just as simple as that. It's easy to reason about, users don't need to worry about relations such as "phrase A is a prefix/substring/suffix of phrase B, and shadowing etc", they can just check the presence of strings explicitly, and tune their weights and distribution accordingly. Hence setting the phrases becomes a user-space policy that core engine deveopers don't need to attend to.

I personally believe this is the right design for phrase-based context biasing in ASR. Hope the idea is helpful.

pkufool:
I added a output arc to fix the "search shadowing", here is the graph for " S / HE / SHE / SHELL / HIS / HERS / HELLO / THIS / THEM", red arcs are the fail arc, green arcs are the output arcs.

image

```python
queries = {
        "HEHERSHE": 14,  # "HE", "HE", "HERS", "S", "SHE", "HE"
        "HERSHE": 12,  # "HE", "HERS", "S", "SHE", "HE"
        "HISHE": 9,  # "HIS", "S", "SHE", "HE"
        "SHED": 6,  # "S", "SHE", "HE"
        "HELL": 2,  # "HE"
        "HELLO": 7,  # "HE", "HELLO"
        "DHRHISQ": 4,  # "HIS", "S"
        "THEN": 2,  # "HE"
    }
    for query, expected_score in queries.items():
        total_scores = 0
        state = context_graph.root
        for q in query:
            score, state = context_graph.forward_one_step(state, ord(q))
            total_scores += score
        score, state = context_graph.finalize(state)
        assert state.token == -1, state.token
        total_scores += score
        assert total_scores == expected_score, (
            total_scores,
            expected_score,
            query,
        )
```

pkufool:
> In practice, you may have giant list containing tens of thousands of musics, company names, POIs.

My plan currently is to support small set of phrase lists (such as, different sequences have different phrases, something like the on-screen words  recognition).  We could definitely apply this to large set of phrase, but I think we need a more memory-efficient trie implementation (for example double array trie).

Yes, the "distributed weights" is a good idea!  I was thinking we could merge this version first (after checking the correctness of it, of course), as a lot of people were asking this feature. I will try the "distributed weights" and also the incremental method (in the suggested paper) in the following versions.

dophist:
Closer, but not 100% correct to me yet. A correct implementation should be much simpler and cleaner than your current implementation. More specifically:
1. In the middle of the search, all hits from the output table need to be aggregated as single biasing score. Intuitively, this specifies what to do with "multiple hits at the same time".
For example: say, you have a search set { "BC":0.1, "ABC":1.0 } and a hypothesis starts with "ABC...", after the token "C", both phrases will be succesfully matched(in output table), so what should be the weight to apply to the hypothesis? sum 1.1? or average (0.55)? or max or longest (1.0)? This is a key design choice to make if you want consistency and compatibility for the sake of users. If that's my call, I tend to choose sum or average.
2. It feels quite convolved that scores are distributed on both arcs and nodes, and it seems that you add a "score cancelling" machnism that is dubious compared to standard Aho-Corasick implementation.

I haven't spent time to do a line-by-line review of your implementation, my suggestion is that we should first collect or borrow some string-matching test cases, to make sure the Aho-Corasick implementation is precise and correct. 
But please feel free to merge and organize your priorities, my response might be slow and I don't want to be the one blocking any of your development.

pkufool:
> For example: say, you have a search set { "BC":0.1, "ABC":1.0 } and a hypothesis starts with "ABC...", after the token "C", both phrases will be succesfully matched(in output table), so what should be the weight to apply to the hypothesis? sum 1.1? or average (0.55)? or max or longest (1.0)?

I think I used the sum.

> It feels quite convolved that scores are distributed on both arcs and nodes, and it seems that you add a "score cancelling" machnism that is dubious compared to standard Aho-Corasick implementation.

About the "score cancelling", it is because I give out a score for each step, that means I give scores for some prefixes that will not actually reach the end of phrase, so when falling back to some state or at the last token of a sequence I have to cancel them out. If we only output score at the end of the phrase, the "score cancelling" is not necessary.  I don't want to change the policy for now, as it was validated in my early openfst implementaion.  Will change this when I find some convincing results that show outputing scores only at the end of phrase works fine or even better. I need to find some good test sets open-sourced or internally first.

> I haven't spent time to do a line-by-line review of your implementation, my suggestion is that we should first collect or borrow some string-matching test cases, to make sure the Aho-Corasick implementation is precise and correct.

Of course, checking the correctness is the most important stuff.


Thank you! You help me a lot for this PR.

dophist:
> About the "score cancelling", it is because I give out a score for each step, that means I give scores for some prefixes that will not actually reach the end of phrase, so when falling back to some state or at the last token of a sequence I have to cancel them out. If we only output score at the end of the phrase, the "score cancelling" is not necessary. I don't want to change the policy for now, as it was validated in my early openfst implementaion. Will change this when I find some convincing results that show outputing scores only at the end of phrase works fine or even better. I need to find some good test sets open-sourced or internally first.

This smells like a over-design to me, lots of people do this wrong. Just reconsider the problem from first principles, don't follow other's buggy implementation.  The paper I recommand provides a solid ground, the first part of that paper is all about precompile search set, and the author did a great job explaining the invariants need to be satisfied during the search.

pkufool:
> This smells like a over-design to me, lots of people do this wrong.

Might be, the graphs in several of Google's paper have scores on each arc.  Will try your suggestions, thank you!

kobenaxie:
I think “score cancelling” is necessary for context biasing，but how to deal with the case below ？

> 

set `beam_width=2`，one of the biasing list is `“石家庄”`，when saying a word `“时间”`，there are two valid paths， `“时：0.6” `and `“石：0.5”`，while path `“石：0.5”` will be boosted with biasing score 0.5 for example，now path `“石：1.0”` has a greater score，even though at the next decoding step，the biasing score 0.5 will be cancelled，but the path `“时：0.5”` might have died at last step，resulting in the final path `“石鉴”` rather than `“时间”`.

dophist:
No, not necessary at all. It's just an over-design that is theretically problematic and make things unnecessarily complicated.

Google did this right and publish their work in 2015 under the keyword of "salient ngrams biasing" in WFST framework, other people just don't get it, and copy their solution in a wrong way. If you guys read their paper carefully, they explicitly stated that there are two types of arcs, and the failure arcs carry no weights.

What we are doing here is essentially establishing an equvalent implementation under the well established Aho-Corasick algorithm, instead of poking around with WFST composition customization and proper semiring selection for "output table".

 @kobenaxia, for your case, the biasing weights should be pushed to the end of the phrase at the first place, i.e.
```
石 家 庄:1.0
```
and tune it only when it is absolutely necessary with the following setup:
```
石 家:0.1 庄:0.9
```
I shall not explain futher on this topic, and leave the decision to you guys.

pkufool:
> No, not necessary at all. It's just an over-design that is theretically problematic and make things unnecessarily complicated.

I think it is very easy to implement your suggestion in my current code, just deleting some code, will try it tomorrow. I agree that your suggestion is simpler and eaiser to understand, actually it is hard for me to write a clear documentation for my current code, if the performance is ok, I am very glad to choose your idea.  Thanks again.

pkufool:
Again，I think "distributed weights" is a good idea, one question is if two phrases share one node, how to combine their weights? Do you have any thoughts?

ezerhouni:
@pkufool About an open source dataset, maybe :
https://github.com/revdotcom/speech-datasets/tree/main/earnings21
They provide a list words and the model that they used (for one of their paper) has been trained on GigaSpeech.

Thank you for the implementation. I have been struggling to do so.

ezerhouni:
> Again，I think "distributed weights" is a good idea, one question is if two phrases share one node, how to combine their weights? Do you have any thoughts?

Maybe I am mistaken, but wenet implementation has the sum of the arc of each contextual word being equal, but that means that some token/char might have a large weight if it is shared a lot accross contextual words.

pkufool:
> Maybe I am mistaken, but wenet implementation has the sum of the arc of each contextual word being equal, but that means that some token/char might have a large weight if it is shared a lot accross contextual words.

Wenet uses openfst, I think they determinize the graph with log semiring.

pkufool:
> @pkufool About an open source dataset, maybe : https://github.com/revdotcom/speech-datasets/tree/main/earnings21 They provide a list words and the model that they used (for one of their paper) has been trained on GigaSpeech.
> 
> Thank you for the implementation. I have been struggling to do so.

Thank you, will have a look at the test set.

dophist:
I just wrote an exampler draft so that you can play with it. Compare it with your implementation when you have more test cases,
and make sure the core functionality is well-designed.

The core logic takes less then 100 LOC, should be working in Python 3.10+, I do not guarentee it's bug-free, but should be more than enough to demonstrate the idea.
```
import sys, argparse
from dataclasses import dataclass, field
from collections import deque

@dataclass
class Phrase:
    tokens: list[str] = field(default_factory=list)
    scores: list[float] = field(default_factory=list)

    def __len__(self):
        assert(len(self.tokens) == len(self.scores))
        return len(self.tokens)

@dataclass
class TrieNode:
    id: int = 0
    parent: int = 0
    hist: list[str] = field(default_factory=list)
    token: str = ''
    oarcs: dict[str, int] = field(default_factory=dict) # (this node)--token-->(child node)
    score: float = 0.0
    backoff: int = 0
    output_link: int = 0

    def IsOutputNode(self) -> bool:
        return self.score != 0.0

    def Context(self) -> str:
        return ''.join(self.hist) + self.token


class Trie:
    def __init__(self, phrases: list[Phrase]):
        root = TrieNode(parent = -1, backoff = -1, output_link = -1)
        nodes = [root] # root node index = 0

        # Stage 1: build the trie
        for phrase in phrases:
            node = root
            for k in range(len(phrase)):
                t, s= phrase.tokens[k], phrase.scores[k]
                if t in node.oarcs: 
                    child = nodes[node.oarcs[t]]
                    if (s != 0.0):
                        if (child.score != 0.0):
                            print('Warning: found collision of biasing scores.', file=sys.stderr)
                        child.score += s # choose to resolve collision with sum / average / overwrite
                        #child.score = s # overwrite
                else:
                    child = TrieNode(id = len(nodes), parent = node.id, hist = phrase.tokens[:k], token = t, score = s)
                    node.oarcs[t] = child.id
                    nodes.append(child)
                node = child

        # Stage 2: setup failure/output links
        q = deque()
        q.append(root)
        bfs_visitor = []
        while q:
            node = q.popleft()
            for child in node.oarcs.values():
                bfs_visitor.append(nodes[child])
                q.append(nodes[child])
        print(f'BFS visitor: {[ node.id for node in bfs_visitor ]}', file=sys.stderr)

        for node in bfs_visitor:
            if len(node.hist) == 0:
                node.backoff, node.output_link = root.id, root.id
                continue

            p = nodes[node.parent].backoff 
            while p != -1: 
                q = nodes[p].oarcs.get(node.token)
                if q:
                    node.backoff = q
                    if nodes[q].IsOutputNode():
                        node.output_link = q
                        break
                p = nodes[p].backoff

        self.nodes = nodes


    def Goto(self, p, token):
        nodes = self.nodes
        q = nodes[p].oarcs.get(token)
        if q:
            return q

        if p == 0: # This mimics the <UNK> looping arount root node
            return 0
        else:
            return self.Goto(nodes[p].backoff, token)

    def Match(self, tokens: list[str]):
        nodes = self.nodes
        matches = []
        p = 0
        for k, t in enumerate(tokens):
            q = self.Goto(p, t)

            # 1. check if the current search head is an output node
            if nodes[q].IsOutputNode():
                matches.append((q, k))
            # 2. iterate over the remaining output table(a linked list)
            o = nodes[q].output_link
            while o > 0:
                matches.append((nodes[o].id, k))
                o = nodes[o].output_link

            p = q
        return matches
    

if __name__ == '__main__':
    parser = argparse.ArgumentParser()
    parser.add_argument('--hints', type=str, default='hints.txt')
    parser.add_argument('hypothesis', type=str)
    args = parser.parse_args()

    phrases = []
    with open(args.hints) as f:
        # line format: "TOKEN:SCORE" pairs seperated by whitespace, one phrase per line
        #   "A:0.1 B:0.2 ... "
        # ignored scores default to 0.0
        #   "A B ... Y:0.5 Z:1.0"
        for line in f:
            cols = line.strip().split()
            phrase = Phrase()
            for col in cols:
                if ':' in col:
                    t, s = col.strip().split(':')
                else:
                    t, s = col.strip(), None
                phrase.tokens.append(t)
                phrase.scores.append(float(s) if s else 0.0)
            phrases.append(phrase)
    for x in phrases:
        print(f'{x}', file=sys.stderr)

    trie = Trie(phrases)
    for x in trie.nodes:
        print(x, file=sys.stderr)

    def Tokenize(text) -> list[str]:
        return list(text)

    text = args.hypothesis
    matches = trie.Match(Tokenize(text))

    print('-' * 40)
    print(text)
    for m, pos in matches:
        s = trie.nodes[m].score
        print(f'{trie.nodes[m].Context().rjust(pos+1)}:{s}')
    
```

and here is a random example of hints.txt:
```
A B C:1.0
A B:1.0
C:1.0 D:1.0
C E:1.0
B C:1.0
```

kobenaxie:
Hi, @dophist ,

> node.backoff = q

The failure link(`backoff` here) of `node` will be overridden by the output link which might not be the longest proper suffix of current `node`, should be change to 
```
if node.backoff == 0:
    node.backoff = q
```
as  when we have found the  longest proper suffix of some node, there is no need update it anmore.

One example can be found [wiki Aho–Corasick algorithm](https://en.wikipedia.org/wiki/Aho%E2%80%93Corasick_algorithm), The failure link of `node(bca)` should be `node(ca)`, while in your implementation the failure link will be overridden by `node(a)` which is the output link of `node(bca)`.
220px-A_diagram_of_the_Aho-Corasick_string_search_algorithm svg

kobenaxie:
Hi, @dophist ,

 > node.backoff = q

 The failure link(`backoff` here) of `node` will be overridden by the output link which might not be the longest proper suffix of current `node`, here might be changed to
 ```
 if node.backoff == 0:
     node.backoff = q
 ```
 as when we have found the longest proper suffix of some node, there is no need update it anymore.

One example can be found [wiki Aho–Corasick algorithm](https://en.wikipedia.org/wiki/Aho%E2%80%93Corasick_algorithm), The failure link of `node(bca)` should be `node(ca)`, while in your implementation the failure link will be overridden by `node(a)` which is the output link of `node(bca)`.
220px-A_diagram_of_the_Aho-Corasick_string_search_algorithm svg

Another question is about `IsOutputNode`, for the phrase `"C:1.0 D:1.0"`, the `node(C)` is not an output node actually, but the `IsOutputNode()` function will return True, as the score of `node(C)` is 1.0, should be modified here ?

dophist:
> Hi, @dophist ,
> 
> > node.backoff = q
> 
> The failure link(`backoff` here) of `node` will be overridden by the output link which might not be the longest proper suffix of current `node`, here might be changed to
> 
> ```
> if node.backoff == 0:
>     node.backoff = q
> ```
> 
> as when we have found the longest proper suffix of some node, there is no need update it anymore.
> 
> One example can be found [wiki Aho–Corasick algorithm](https://en.wikipedia.org/wiki/Aho%E2%80%93Corasick_algorithm), The failure link of `node(bca)` should be `node(ca)`, while in your implementation the failure link will be overridden by `node(a)` which is the output link of `node(bca)`. 220px-A_diagram_of_the_Aho-Corasick_string_search_algorithm svg
> 
> Another question is about `IsOutputNode`, for the phrase `"C:1.0 D:1.0"`, the `node(C)` is not an output node actually, but the `IsOutputNode()` function will return True, as the score of `node(C)` is 1.0, should be modified here ?

The first point is a bug indeed, I just updated the code snippet with a guard, and the provided `hints.txt` is now a precise demonstration of the graph from wiki.

For the second point, note that:
```
A B C:0.1 D:1.0
```
is a shortcut of two registrations:
```
A B C:0.1
A B C D:1.0
```

pkufool:
It is being a while, I re-test this feature with more solid test sets here https://github.com/pkufool/open-biasing (under construction),  in short, the test sets were created by extracting the NERs in original test sets.

Here is the results, you can see that it can improve the contexts subset without affecting the remaining part.

For AIshell (has 1073 context phrases):

  | test | dev | contexts subset | (test + dev) - contexts subset |  
-- | -- | -- | -- | -- | --
No contexts | 4.81 | 4.4 | 13.47 | 3.81 |  
With contexts | 4.68 | 4.29 | 11.95 | 3.81 | Context-score = 1.5
improvement | 2.7% | 2.5% | 11.28% | - |  

For Librispeech (has 487 context phrases):

  | clean | other | contexts subset | (clean + other) - contexts subset |  
-- | -- | -- | -- | -- | --
No contexts | 2.62 | 6.57 | 10.59 | 3.74 |  
With contexts | 2.44 | 6.22 | 8.41 | 3.75 | Context-score = 1.5
improvement | 6.87% | 5.3% | 20.58% | - |  

Here are some cases:

AIshell:

image
image

LibriSpeech:

image
image

pkufool:
@csukuangfj @danpovey  I am thinking we can merge this version first,  hopefully, the community guys will help to do more testing.

csukuangfj:
> @csukuangfj @danpovey  I am thinking we can merge this version first,  hopefully, the community guys will help to do more testing.

Yes, +2

PR #1044, opened 2023-05-08
make egs/timit work according to the documentation

rouseabout:
* prepare.sh: restore working directory after git lfs pull
* set execute permisons on python scripts called by prepare.sh


csukuangfj:
Thanks for your contribution!

PR #1051, opened 2023-05-11
[Not for merge] Zipformer with cross attention

marcoyang1998:
This PR is **not** for merge. It's only for illustration purposes.

This PR adds cross-attention in the zipformer.


danpovey:
BTW, the conflicts in the 'pruned2_knowledge' directory... we should delete that directory from the branches we are using and also the master.  It was merged by accident to the master, I believe, and was never something that worked.

marcoyang1998:
OK. Shall I remove it and push the changes to #1052?

PR #1055, opened 2023-05-11
fix rnn_lm/train.py usage

arbs-gpu:
fix rnn_lm/train.py usage


csukuangfj:
Thanks for your contribution!

PR #1061, opened 2023-05-15
Generate supervision from text and recordings.

csukuangfj:
When importing data from Kaldi, lhotse requires that there must exist a `segments` file in order to generate a supervision file. However, not all datasets from kaldi have such a file, e.g., the yesno recipe.

This PR provides a standalone script to generate supervisions from given `text` and `recordings.jsonl.gz`.

### DIFF
--- a/icefall/shared/text_to_supervision.py
+++ b/icefall/shared/text_to_supervision.py
@@ -5,9 +5,9 @@
 This script takes the following two files as inputs:
 
 - text
-- recordings.jsonl.gz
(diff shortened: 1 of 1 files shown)


csukuangfj:
@pzelasko 

Do you think it is reasonable to remove the constraint that `segments` is required to generate supervisions in lhotse?

pzelasko:
Yea Lhotse should support these cases out of the box. Could you modify it?

PR #1095, opened 2023-05-24
Update pre-commit isort package to v5.11.5

rouseabout:
Earlier versions of isort are not compatible with current version of poetry-core. This causes "pre-commit install" to fail. See: https://github.com/PyCQA/isort/issues/2077


csukuangfj:
Thanks!

PR #1104, opened 2023-05-31
Add averaged model and new-bob scheduler to rnnlm 

yfyeung:
- [x] Add NewBob scheduler
- [x] Add support for averaged model
- [x] Fix typo 

### Result on GigaSpeech XL transcription
| method | ppl |
|-- |-- |
| Adam | |
| + NewBobScheduler | |


icefall/rnn_lm/train.py
```diff
@@ -589,10 +663,20 @@ def run(rank, world_size, args):
         lr=params.lr,
         weight_decay=params.weight_decay,
     )
+    scheduler = NewBobScheduler(optimizer)
```
marcoyang1998, line 666: Could you make it possible for users to pass some arguments to the scheduler?

yfyeung: Sure.

videodanchik:
Any results improvements with averaged model + Scheduler?

danpovey:
sorry, we are now using very different learning rate schedules and a different optimizer.
we need to revisit the RNNLM work in the near future, possibly using a transformer-based teacher model.
I think this recipe is a little outdated.

PR #1105, opened 2023-05-31
Remove multidataset from librispeech/pruned_transducer_stateless7

yfyeung:
This PR removes multi-dataset support from egs/librispeech/pruned_transducer_stateless7.
Note the following recipes still support multi dataset.
- lstm_transducer_stateless2
- pruned_transducer_stateless3
- pruned_transducer_stateless7_streaming_multi
- pruned_transducer_stateless8
- transducer_stateless_multi_datasets


yaozengwei:
To support the multi-dataset training for the lastest recipe `zipformer`, I suggest to create a new directory namely `multidataset_en` instead (We will also create `multidataset_cn` similarly). This would make the directory structure clearer. It would also make the data preparation less confusing for other users, since I notice the code `prepare_multidataset` also do prepare the librispeech dataset. 
What do you think @csukuangfj @pkufool ?

yfyeung:
> To support the multi-dataset training for the lastest recipe `zipformer`, I suggest to create a new directory namely `multidataset_en` instead (We will also create `multidataset_cn` similarly). This would make the directory structure clearer. It would also make the data preparation less confusing for other users, since I notice the code `prepare_multidataset` also do prepare the librispeech dataset. What do you think @csukuangfj @pkufool ?

Sure

yaozengwei (approved):

PR #1116, opened 2023-06-05
Add zipformer to peoples_speech

yfyeung:


yfyeung:
Thanks for your train and decode code. Currently, we're in need of trained models and decoding results. This is the reason this PR is stuck.

PR #1117, opened 2023-06-07
verify have installed ffmpeg

SarahSmitho:
Add verification to ensure that ffpmeg is installed to prevent audio loading errors such as audioread exceptions. NoBackendError


csukuangfj:
Thanks for your contribution!

PR #1123, opened 2023-06-08
Use symlinks for best epochs

rouseabout:
* Saves precious disk space.
* Convenient to discover best epoch weights by running `ls -l` on the experiment directory.


yfyeung (approved):
Thanks!

PR #1129, opened 2023-06-14
Add multi_en

yfyeung:


egs/multi_en/ASR/zipformer/train.py
```diff
@@ -0,0 +1,1394 @@
...
+            params=params,
+        )
+
+    scaler = GradScaler(enabled=params.use_fp16, init_scale=1.0)
+    if checkpoints and "grad_scaler" in checkpoints:
+        logging.info("Loading grad scaler state dict")
+        scaler.load_state_dict(checkpoints["grad_scaler"])
+
+    for epoch in range(params.start_epoch, params.num_epochs + 1):
+        scheduler.step_epoch(epoch - 1)
```
yaozengwei, line 1255: I think we should remove this line here.

yfyeung: You are right.

PR #1157, opened 2023-06-30
Zipformer2 logaddexp onnx safe

MicKot:
I've encountered a situation where the last `SwooshR` in `Subsampling.conv` produces inf due to naive `logaddexp` operation for tracing, so this PR is supposed to fix it


csukuangfj:
Thanks for your contribution!

danpovey:
We should actually test whether torch.logaddexp works in jit tracing mode.  I think the if-statement was probably added when we had some kind of custom kernel there that wouldn't work in tracing mode.

csukuangfj:
> We should actually test whether torch.logaddexp works in jit tracing mode. I think the if-statement was probably added when we had some kind of custom kernel there that wouldn't work in tracing mode.

@danpovey  Please see #1158

PR #1160, opened 2023-07-03
Fix ONNX export for the latest non-streaming zipformer.

csukuangfj:
See https://github.com/k2-fsa/icefall/commit/c3e23ec8d2a3ed2547bd94dee7280bd3f193a47e#commitcomment-120481693

# TODO
- [ ] Add GitHub actions to covert ONNX export-related scripts in separate pull requests.

@joazoa

Please try this pull request. I have tested it locally that  both `zipformer/export-onnx.py` and `zipformer/export-onnx-streaming.py`  work as expected with this change. Also, the exported non-streaming and streaming models are able to decode files with `onnx_pretrained.py` and `onnx_pretrained-streaming.py`, respectively.

### DIFF
--- a/egs/librispeech/ASR/zipformer/scaling.py
+++ b/egs/librispeech/ASR/zipformer/scaling.py
@@ -25,6 +25,11 @@ import math
 import torch.nn as nn
 from torch import Tensor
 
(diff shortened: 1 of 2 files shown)


joazoa:
This solved he problem, thank you for fixing it (and apologies for not clarifying earlier that this was the offline export).

PR #1161, opened 2023-07-03
Add start-batch option for RNNLM training

nshmyrev:
If epochs are long, it is helpful to start from batch just like other trainings


csukuangfj:
> First-time contributor

Thanks for your contribution!

PR #1176, opened 2023-07-20
disable speed perturbation by default

JinZr:


danpovey:
We should probably do this, even though it may make some of the recipes not fully reproducible (i.e. it would require to change the learning rate schedules and num-epochs).
It might make sense to actually change the LR schedules and epochs at the same time.  This would not be fully
foolproof, since some people will have previously-dumped non-perturbed features on disk.  But it will ensure that
from a fresh checkout things would be basically reproducible.
(if README files have example usages, we could perhaps just write the alternative command lines that you would use
if the features were not dumped with speed perturbation, with the num-epochs and if applicable lr-epochs multiplied by 3.)

pzelasko:
It doesn’t help anymore?

JinZr:
Not really. The thing is when the amount of data reaches a certain point,
speed perturbation stops bringing performance improvement to the system and
would be a waste of space and time. So we figure that we can leave a switch
and let users decide whether they want it or not.

pkufool:
> It doesn’t help anymore?

We did experiments on librispeech dataset, the performance of  speed-pertur is the same as no speed-pertur (3 times epochs).

yfyeung:
> We did experiments on librispeech dataset, the performance of speed-perturb is the same as no speed-perturb (3 times epochs).

Somehow slightly worse than with perturb

yfyeung (approved):

yfyeung (changes requested):
Maybe you can use:

if speed_perturb and "train" in partition:

JinZr:
applied requested changes to the latest commit

danpovey:
There is already a PR (sorry IDK the PR#) to change the arg to false in the
prepare.sh scripts.

egs/aidatatang_200zh/ASR/local/compute_fbank_aidatatang_200zh.py
```diff
@@ -109,7 +109,12 @@ def get_args():
         default=80,
         help="""The number of mel bins for Fbank""",
     )
-
+    parser.add_argument(
+        "--speed-perturb",
+        type=bool,
```
csukuangfj, line 114: Please follow
https://github.com/k2-fsa/icefall/blob/00256a766921dd34a267012b0e2b8ff7d538f0e6/egs/librispeech/ASR/tdnn_lstm_ctc/asr_datamodule.py#L87
to use `str2bool`.

csukuangfj (commented):
Thanks! Left a minor comment. Otherwise, it looks good to me.

egs/aidatatang_200zh/ASR/local/compute_fbank_aidatatang_200zh.py
JinZr: thx! fixed this in the latest commit.

csukuangfj (approved):
Thanks!

PR #1181, opened 2023-07-25
Fix text normalization bug in CommonVoice.

kobenaxie:
`re.sub(r"[^a-zA-Z\s]", "", utt).upper()` will remove `'`, for example, `isn't` will be `ISNT` after text normalizetion.


danpovey:
It's not clear to me that that is what we want.   What is the goal here?

danpovey:
Oh sorry I misunderstood the change, I thought you were doing the opposite; you are making it keep the '.
Yes I agree that this makes sense.  In fact these days I am leaning towards doing everything with completely
unnormalized text (except maybe scoring).  But that's a larger discussion.

yfyeung:
I agree with you. Still, there are many non-sense ' in the CommonVoice transcript.

yfyeung (approved):

PR #1202, opened 2023-08-07
[Not for merge] Add Bengaliai speech

yfyeung:
Kaggle Competition: https://www.kaggle.com/competitions/bengaliai-speech/overview
The goal of this competition is to recognize Bengali speech from out-of-distribution audio recordings. You will build a model trained on the first Massively Crowdsourced (MaCro) Bengali speech dataset with 1,200 hours of data from ~24,000 people from India and Bangladesh. The test set contains samples from 17 different domains that are not present in training


npovey:
hi,
I tried this code but the duration in manifests for **all** audios are zero.

 ...."duration": 0.0,...

not sure if the output below is the cause:

2023-08-14 23:47:41,483 INFO [audio.py:137] The user overrided the global setting for whether to use ffmpeg-torchaudio to compute the duration of audio files. Old setting: True. New setting: False.

yfyeung:
> hi, I tried this code but the duration in manifests for **all** audios are zero.
> 
> ...."duration": 0.0,...
> 
> not sure if the output below is the cause:
> 
> 2023-08-14 23:47:41,483 INFO [audio.py:137] The user overrided the global setting for whether to use ffmpeg-torchaudio to compute the duration of audio files. Old setting: True. New setting: False.
> 
> PS: added to stage 0 the line below
> 
> ```
>   unzip bengaliai-speech.zip -d download/bengaliai_speech
> ```

Using ffmpeg-torchaudio to compute the duration of `.mp3` files leads to %CPU more than 100%, see issue: https://github.com/lhotse-speech/lhotse/issues/1026
So we disable ffmpeg-torchaudio when the type of audio files is `mp3`.
I don't think this is the reason causing `the duration in manifests for all audios are zero.`

danpovey:
@yfyeung so do you have a theory why the durations might be zero?

npovey:
The problem is fixed.
Actually, I was having a problem on my text_search project. The m4a duration was calculated wrong and @pzelasko advised to have pytorch 2.0 or above see here:https://github.com/lhotse-speech/lhotse/issues/1121
After running
```
pip3 install torch torchvision torchaudio
```
This fixed this problem.

I had:
torch: 1.12.1+cu113
torchaudio: 0.12.1+cu113
lhotse: 1.16.0

Now I have:
python -c "import torch; print(torch.__version__)"
2.0.1+cu117
python -c "import torchaudio; print(torchaudio.__version__)"
2.0.2+cu117

PR #1215, opened 2023-08-13
Speed up yesno training to finish in ~10s on CPU

pzelasko:
Setting `persistent_workers=True` when creating DataLoader removes a significant overhead at the start of each epoch and validation. Thanks to this, it finishes in ~10s vs a couple of minutes.


csukuangfj:
Wow! Thanks!

danpovey:
good catch!

PR #1222, opened 2023-08-17
[NOT FOR MERGE] Add Blank Skip to Zipformer2

ezerhouni:
I will double check that I am not missing something.

cc @csukuangfj

### DIFF
--- a/egs/librispeech/ASR/zipformer/train.py
+++ b/egs/librispeech/ASR/zipformer/train.py
@@ -538,6 +538,7 @@ def get_params() -> AttributeDict:
             "reset_interval": 200,
             "valid_interval": 3000,  # For the 100h subset, use 800
             # parameters for zipformer
+            "ctc_beam_size": 10,
             "feature_dim": 80,
             "subsampling_factor": 4,  # not passed in, this is fixed.
             "warm_step": 2000,
@@ -783,6 +784,7 @@ def compute_loss(
     sp: spm.SentencePieceProcessor,
     batch: dict,
     is_training: bool,
+    warmup: float,
 ) -> Tuple[Tensor, MetricsTracker]:
     """
     Compute loss given the model and its inputs.
@@ -823,9 +825,12 @@ def compute_loss(
             x=feature,
             x_lens=feature_lens,
             y=y,
+            supervisions=supervisions,
             prune_range=params.prune_range,
             am_scale=params.am_scale,
             lm_scale=params.lm_scale,
+            ctc_beam_size=params.ctc_beam_size,
+            warmup=warmup,
         )
 
         loss = 0.0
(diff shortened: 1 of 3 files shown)


yfyeung:
Thanks!

egs/librispeech/ASR/zipformer/lconv.py
```diff
@@ -0,0 +1,113 @@
...
+
+from typing import Optional
+
+import torch
+import torch.nn as nn
+from acoustic_model.utils_py.scaling_zipformer import Balancer, ScaledConv1d
+
+
+class LConv(nn.Module):
+    """A convolution module to prevent information loss."""
```
danpovey, line 25: should probably clarify that this relates to blank skipping and maybe reference the paper.

egs/librispeech/ASR/zipformer/train.py
```diff
@@ -620,16 +640,24 @@ def get_model(params: AttributeDict) -> nn.Module:
...
         decoder=decoder,
         joiner=joiner,
+        lconv=lconv,
+        frame_reducer=frame_reducer,
         encoder_dim=max(_to_int_tuple(params.encoder_dim)),
         decoder_dim=params.decoder_dim,
         vocab_size=params.vocab_size,
         use_transducer=params.use_transducer,
         use_ctc=params.use_ctc,
+        use_bs=params.use_bs,
```
danpovey, line 660: I think it would be better to call this use_blank_skip.

danpovey (commented):
A couple comments (not necessarily promising to merge this, must discuss with the guys.)

Also: since you are only comparing warmup with 2.0, and warmup is now not used elsewhere, it would be better to have a config "blank_skip_start_batch" and just pass in a bool that says whether we do blank slipping.

What happens in test time?  And I wonder whether this would require changes to sherpa.

The guys would have to comment before we were  to merge this, my comments can be ignored if we are not going to merge.

yaozengwei (commented):
I remember that @yfyeung said removing the Lconv module gets better result.

ezerhouni:
@csukuangfj @danpovey If merging this PR is not very urgent and useful (>2 weeks), I can run it librispeech with all the changes needed (from review this PR) update Results.md

danpovey:
We also need to see whether this might cause a problem for streaming systems, e.g. does it support streaming_forward().

csukuangfj:
> @csukuangfj @danpovey If merging this PR is not very urgent and useful (>2 weeks), I can run it librispeech with all the changes needed (from review this PR) update Results.md

That would be great. We have not used blank skipping with the latest zipformer.

ezerhouni:
@danpovey It should work with streaming_forward 
@csukuangfj I will try to schedule that asap. In the meantime, any review would be welcome (so that I can test the reviewed code)

Thank you

PR #1226, opened 2023-08-21
[Do not merge] example of using LibriSpeech + Lhotse Shar

pzelasko:
This PR shows how to use Lhotse Shar data format to train models in Icefall. I only added the necessary adaptation to the basic TDNN LSTM recipe as I don't have a free GPU at hand to run anything bigger. If it's useful, feel free to adapt this to more recent recipes (very little change is required as most of the changes come inside the datamodule).

### DIFF
--- a/egs/librispeech/ASR/tdnn_lstm_ctc/asr_datamodule.py
+++ b/egs/librispeech/ASR/tdnn_lstm_ctc/asr_datamodule.py
@@ -29,10 +29,12 @@ from lhotse.dataset import (  # noqa F401 for PrecomputedFeatures
     CutConcatenate,
     CutMix,
     DynamicBucketingSampler,
+    IterableDatasetWrapper,
     K2SpeechRecognitionDataset,
     PrecomputedFeatures,
     SingleCutSampler,
     SpecAugment,
+    make_worker_init_fn,
 )
 from lhotse.dataset.input_strategies import (  # noqa F401 For AudioSamples
     AudioSamples,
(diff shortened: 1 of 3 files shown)


danpovey:
We will try to apply this to other recipes and see what performance improvements we get.

egs/librispeech/ASR/tdnn_lstm_ctc/asr_datamodule.py
```diff
@@ -473,3 +486,240 @@ def test_other_cuts(self) -> CutSet:
...
+    @lru_cache()
+    def train_all_shuf_cuts(self) -> CutSet:
+        logging.info(
+            "About to get the shuffled and sharded cuts for full (960h) LibriSpeech using Lhotse Shar"
+        )
+        # Below we'll figure out which files to read.
+        # Since we only use either (precomputed) features or recordings,
+        # we shouldn't iterate over both at the same time.
+        shar_dir = Path(self.args.shar_dir)
+        fields = {"cuts": sorted(shar_dir.glob("cuts.*.jsonl*"))}
```
danpovey, line 680: I am a little confused about this glob stuff, is it just grabbing everything in the shar directory, and how does it work if we just want train-clean-100 or the test set?

pzelasko: Yeah it only supports full LibriSpeech. But you can shar-export every piece separately and then here you’d concatenate the result of globs in each exported directory.

danpovey: OK.  IMO it might be a better practice to give the shar-dir a more specific name, like shar_train_960 or something, to indicate it is a specific dataset.

PR #1232, opened 2023-08-30
Missing definitions in scaling.py added

desh2608:
This is a fix for a bug in the SURT recipe. I had made `scaling.py` as a symbolic link earlier to the one in LibriSpeech zipformer. But the recipe needs additional components (such as ScaledLSTM) which are not present in that version. So I copy it here and add those components.

This is a small change and shouldn't affect anything else.


csukuangfj:
Thanks!

PR #1240, opened 2023-09-04
Fix for `RuntimeError: received o items of ancdata` during WeNetSpeech feature extraction

JinZr:


csukuangfj:
Could you give more context about the solution?

JinZr:
This solution comes from an issue https://github.com/pytorch/pytorch/issues/973#issuecomment-604473515 of the PyTorch repo.

PR #1261, opened 2023-09-19
Libriheavy recipe (zipformer)

pkufool:


pkufool:
@pingfengluo from community wants to try this experiment quickly, so, if there is no major errors, I suggest to merge it as soon as possible, I will upload more models and update the results in the following RP.

JinZr (approved):
looks good to me

a minor update has been applied to incorporate https://github.com/k2-fsa/icefall/pull/1269

PR #1269, opened 2023-09-24
fixes for init value of `diagnostics.TensorDiagnosticOptions`

JinZr:
Replace `2 ** 22` with `512` as the default value of `diagnostics.TensorDiagnosticOptions`

also black formatted some scripts


csukuangfj (approved):
Thanks!

PR #1272, opened 2023-09-25
Fix docs for MVQ

marcoyang1998:
Typo fix

### DIFF
--- a/docs/source/recipes/Non-streaming-ASR/librispeech/distillation.rst
+++ b/docs/source/recipes/Non-streaming-ASR/librispeech/distillation.rst
@@ -47,7 +47,7 @@ The data preparation contains several stages, you can use the following two
 options:
 
   - ``--stage``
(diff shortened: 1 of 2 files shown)


docs/source/recipes/Non-streaming-ASR/librispeech/distillation.rst
```diff
@@ -47,7 +47,7 @@ The data preparation contains several stages, you can use the following two
 options:
 
   - ``--stage``
-  - ``--stop-stage``
+  - ``--stop_stage``
```
csukuangfj, line 50: I think --stop-stage and --stop_stage don't make a difference for prepare.sh

marcoyang1998: Ah, I just noticed this, so it should work in both ways. Anyway, this fixes the problem.

PR #1277, opened 2023-09-27
check bbpe model exists in advance.

yaguanghu:


csukuangfj:
Thanks for your contribution!

PR #1278, opened 2023-09-27
Compatible with new lhotse versions.

yaguanghu:
See [CutMix](https://github.com/lhotse-speech/lhotse/blob/5c617ee800376418c9c40acbf24db70540882266/lhotse/dataset/cut_transforms/mix.py#L14C13-L14C13) for more details.


csukuangfj:
Thanks.

https://github.com/lhotse-speech/lhotse/pull/617 changed that argument.

PR #1292, opened 2023-10-07
Minor fixes to the `optim.py`

JinZr:
"It reduces the chance of nan gradients ruining the model, and also it makes the clipping scale actually get applied; there was a bug before, where grad was not used because it was not passed into the other functions directly." -- from Dan

### DIFF
--- a/egs/librispeech/ASR/zipformer/optim.py
+++ b/egs/librispeech/ASR/zipformer/optim.py
@@ -491,6 +491,12 @@ class ScaledAdam(BatchedOptimizer):
                 if self.show_dominant_parameters:
                     assert p.shape[0] == len(param_names)
                     self._show_gradient_dominating_parameter(tuples, tot_sumsq)
+            if ans != ans:  # e.g. ans is nan
+                ans = 0.0
+            if ans == 0.0:
+                for p, state, param_names in tuples:
+                    p.grad.zero_()  # get rid of infinity()
+
             return ans
 
     def _show_gradient_dominating_parameter(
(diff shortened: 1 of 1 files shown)


JinZr:
This PR is pending a result on full-libri dataset hence has been converted to draft.

JinZr:
30th Epoch
|                       | test-clean | test-other |
|-----------------------|------------|------------|
| baseline from Icefall |    2.27    |    5.10    |
| w. fixed optim.py     |    2.36    |    5.14    |

40th Epoch
|                       | test-clean | test-other |
|-----------------------|------------|------------|
| baseline from Icefall |    2.23    |    4.96    |
| w. fixed optim.py     |    2.28    |    5.02    |

danpovey:
In the training logs there are things like this:
```
2023-10-08 20:19:58,841 INFO [optim.py:468] (0/4) Clipping_scale=2.0, grad-norm quartiles 1.470e+02 1.796e+02 1.998e+02 2.262e+02 3.299e+02, threshold=3.995e+02, percent-clipped=0.0
```
(occassionally 1% is clipped.) This indicates that the clipping was for the most part not actually happening in this run, except for up to a few percent clipped in epoch 1.  So I think any WER differences here are just random noise, and it should be OK to merge this.
We should pay attention to these logs though... it's possible that there are situations where this kind of clipping is too aggressive, and  we should use a larger value of clipping_scale (e.g. 10.0) or not set it at all.

yaozengwei (approved):

PR #1303, opened 2023-10-11
Fix padding issues

wd929:
Fixed issues mentioned in https://github.com/k2-fsa/icefall/pull/670 for Conformer and zipformer

### DIFF
--- a/egs/librispeech/ASR/pruned_transducer_stateless2/conformer.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless2/conformer.py
@@ -158,7 +158,7 @@ class Conformer(EncoderInterface):
         if not is_jit_tracing():
             assert x.size(0) == lengths.max().item()
 
-        src_key_padding_mask = make_pad_mask(lengths)
+        src_key_padding_mask = make_pad_mask(lengths, x.size(0))
 
         if self.dynamic_chunk_training:
             assert (
--- a/egs/librispeech/ASR/pruned_transducer_stateless7/zipformer.py
+++ b/egs/librispeech/ASR/pruned_transducer_stateless7/zipformer.py
@@ -281,7 +281,7 @@ class Zipformer(EncoderInterface):
 
         lengths = (x_lens - 7) >> 1
         assert x.size(0) == lengths.max().item(), (x.shape, lengths, lengths.max())
-        mask = make_pad_mask(lengths)
+        mask = make_pad_mask(lengths, x.size(0))
 
         outputs = []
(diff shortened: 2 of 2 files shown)


wd929:
@csukuangfj would you mind reviewing this PR? CC @yuekaizhang

egs/librispeech/ASR/pruned_transducer_stateless2/conformer.py
```diff
@@ -158,7 +158,7 @@ def forward(
         if not is_jit_tracing():
             assert x.size(0) == lengths.max().item()
 
-        src_key_padding_mask = make_pad_mask(lengths)
+        src_key_padding_mask = make_pad_mask(lengths, x.size(0))
```
csukuangfj, line 161: Does `x.size(0)` equals to `lengths.max()`?

wd929: When we do inference with fixed-length and batched inputs, the x.size(0) might not match the lengths.max()
You can see how we send audios in our client: https://github.com/yuekaizhang/Triton-ASR-Client/blob/main/client.py#L340
If we pad the audios with large size of lengths.max(), it will give errors.
So we create the PR to fix the issue in this scenario.

csukuangfj: Line 159 ensures that x.size(0) == lengths.max().

When you pad the samples in triton server, could you also update `lengths` accordingly?

wd929: The lengths should always be the actual length of input audios. We pad the audio to the nearest 10 seconds then send it to the server. This will not affect training. 
For example, we have audios lengths = [9.5, 6.0, 7.6] seconds and we will pad it to 10 seconds. The x.size(0)=10, but the lengths.max()=9.5. This will give errors when we do inference.

csukuangfj: I see. Thanks for the explanation.

csukuangfj (approved):

egs/librispeech/ASR/pruned_transducer_stateless2/conformer.py
yuekaizhang: Yeah,  we have cases that x.size(0) > lengths.max().item() during triton inference, since audios from different users has to be padded to the same length in order to do batch inference. So we have to pad them at client side.

PR #1322, opened 2023-10-19
small enhancements

vesis84:
- add extra check of 'x' and 'x_lens' to earlier point in Transducer model
- specify 'utf' encoding when opening text files for writing (recogs, errs)


csukuangfj:
Thanks!

PR #1325, opened 2023-10-19
Update test_decode.py

Tanmay-Tiwaricyber:
I've removed unnecessary comments and condensed the code to make it more concise while retaining its functionality.

### DIFF
--- a/test/test_decode.py
+++ b/test/test_decode.py
@@ -1,31 +1,11 @@
 #!/usr/bin/env python3
-# Copyright      2021  Xiaomi Corp.        (authors: Fangjun Kuang)
-#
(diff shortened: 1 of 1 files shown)


test/test_decode.py
```diff
@@ -44,15 +24,9 @@ def test_nbest_from_lattice():
         lattice=lattice,
         num_paths=10,
         use_double_scores=True,
-        nbest_scale=0.5,
+        nbest_scale=0.5
     )
-    # each lattice has only 4 distinct paths that have different word sequences:
```
csukuangfj, line 49: Is there a reason to remove these comments?

PR #1333, opened 2023-10-24
Fix CI tests

csukuangfj:
- Replace https://k2-fsa.org/nightly/ with https://k2-fsa.github.io/k2/cpu.html
- Update k2 to the latest version


csukuangfj:
The remaining errors are due to the inaccessibility of http://www.openslr.org/

Issue #1339, opened 2023-10-25
transcribe utterances


AlexandderGorodetski:
Hello guys... Do you have some script that receives list of utterances and performs K2 based transcriptions on the input utterances...

AlexandderGorodetski:
By list of utterances I mean list of short waves (No VAD is needed)

JinZr:
Hi, thanks for your interest.

You can have a look at the `pretrain.py` [here](https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/zipformer/pretrained.py).

AlexandderGorodetski:
I am using tedlium/zipformer recipe as my starting point. I see that tedlium/zipformer directory does not contain export.py file. Could you please guide me how can I create export.py file.

JinZr:
ah you dont need to create it by yourself, there's already a softlink of
the `export.py` for t
he tedlium3/zipformer recipe here:
https://github.com/k2-fsa/icefall/blob/master/egs/tedlium3/ASR/zipformer/export.py
you can use it directly with
the `tokens.txt` you used to train the model.

return to me at any time if you have further questions, thx!

AlexandderGorodetski:
Hello. Unfortunately this export.py is not working for Tedlium recipe. It seems that models for Tedlium and Librispeech are different. I got following error

Traceback (most recent call last):
  File "./zipformer/export.py", line 170, in <module>
    from train import add_model_arguments, get_model, get_params
ImportError: cannot import name 'get_model' from 'train' (/workspace/inputs/alexg/asr/src/models/k2/icefall/egs/tedlium3/ASR_FA/zipformer/train.py)

JinZr:
sorry to hear that, seems like there might be some issues with the tedlium
recipe.

i’ll look into that rn, a fix should be done by weekend.


Best Regards
Jin


AlexandderGorodetski ***@***.***>于2023年10月26日 周四14:22写道：

desh2608:
I think you just need to replace that with `get_asr_model`. The function name seems to have changed recently but the implementation is the same.

JinZr:
this PR https://github.com/k2-fsa/icefall/pull/1345 should fix the issue, please update your codebase to match the master branch and try again. 

thx!

PR #1345, opened 2023-10-27
fix an `ImportError` in the tedlium3/zipformer recipe

JinZr:
ref: https://github.com/k2-fsa/icefall/issues/1339


csukuangfj (approved):
thanks!

PR #1361, opened 2023-10-31
Created auto_comment.yml

Killer2OP:


Killer2OP:
@csukuangfj

csukuangfj:
Thank you for your time. However, I don't think we need such a feature. Users who are watching this repo will receive lots of unrelated messages if we merge this PR.

By the way, please help improve our documentation if you have more time.

Thanks again!

PR #1364, opened 2023-11-02
Update the parameter 'vocab-size'

wnywbyt:
Update the parameter 'vocab-size'


csukuangfj:

PR #1366, opened 2023-11-03
updated grad check for `optim.py`

JinZr:


danpovey:
I think this is OK to merge as long as you can run
python3 ./optim.py
and the test seems to run without crashing

JinZr:
the test finished without crushing, i'll see if fangjun is available to do
the reviewing, thanks!

danpovey (commented):
its OK

danpovey:
its OK, reviewed & merged myself

PR #1372, opened 2023-11-06
Add a TTS recipe VITS on LJSpeech dataset

yaozengwei:
This PR adds a TTS baseline in icefall. 
* model: VITS, [Conditional Variational Autoencoder with Adversarial Learning for End-to-End Text-to-Speech](https://arxiv.org/pdf/2106.06103.pdf). 
* dataset: [LJSpeech](https://keithito.com/LJ-Speech-Dataset/)

The model related codes are mostly copied from [espnet](https://github.com/espnet/espnet/tree/master/espnet2/gan_tts/vits). 

TODO: 
- [x] Support model exporting.
- [x] Upload checkpints and training logs. 
- [x] Add a document.  

Will add a recipe on VCTK dataset later.

PR #1380, opened 2023-11-10
A TTS recipe VITS on VCTK dataset

JinZr:
akin to PR https://github.com/k2-fsa/icefall/pull/1372


yaozengwei:
If possible, could make files be soft links from the `ljspeech` recipe instead of copy, like model files. They are compatible for multi-speakers TTS.

JinZr:
> If possible, could make files be soft links from the `ljspeech` recipe instead of copy, like model files. They are compatible for multi-speakers TTS.

Noted, thanks!

I plan to replace duplicated files with softlinks after #1372 is merged to master branch.

egs/vctk/TTS/prepare.sh
```diff
@@ -0,0 +1,132 @@
...
+if [ $stage -le 7 ] && [ $stop_stage -ge 7 ]; then
+  log "Stage 7: build monotonic_align lib"
+  if [ ! -d vits/monotonic_align/build ]; then
+    cd vits/monotonic_align
+    python setup.py build_ext --inplace
+    cd ../../
+  else 
+    log "monotonic_align lib already built"
+  fi
+fi
```
yaozengwei, line 132: Could you add this part to the script `prepare.sh` in `ljspeech` recipe, and also correspondingly modify the doc?

requirements_tts.txt
```diff
@@ -0,0 +1,6 @@
+# for TTS recipes
+matplotlib==3.8.2
+cython==3.0.6
+numba==0.58.1
+g2p_en==2.1.0
+espnet_tts_frontend==0.0.3
```
yaozengwei, line 6: Could guide users to install these dependencies in doc (also in `ljspeech` recipe)

yaozengwei (approved):

PR #1391, opened 2023-11-24
further refinement to the swbd recipe

JinZr:
including a narrowband setup

### DIFF
--- a/egs/swbd/ASR/local/compute_fbank_eval2000_nb.py
+++ b/egs/swbd/ASR/local/compute_fbank_eval2000_nb.py
@@ -66,14 +66,8 @@ def get_args():
     parser.add_argument(
         "--perturb-speed",
         type=str2bool,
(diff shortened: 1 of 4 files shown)


huangruizhe:
Thanks for the efforts! Do you know what's the WER of the new recipe, and will there be a trained model released? Thank you.

JinZr:
hi!

I have updated the WERs of narrowband systems in the latest commit. This PR
should be merged soon, thanks for the waiting!

huangruizhe:
Thanks. Is it true that the WER improves quite a lot after the narrow band setup?

JinZr:
Yes, indeed. I am a bit confused about the results as well, I think we can certainly arrange a similar experiment using ESPnet with a few tweaks to the setup (they upsample the data to 16kHz as a default) and see how the results change 🧐.

Issue #1356, opened 2023-10-31
Zipformer streaming model: repeated single-syllable words/characters stop abruptly


rouseabout:
I have been experimenting with the pre-trained zipformer streaming models: 

* https://huggingface.co/marcoyang/icefall-libri-giga-pruned-transducer-stateless7-streaming-2023-04-04
* https://huggingface.co/pkufool/icefall-asr-zipformer-streaming-wenetspeech-20230615

A problem I observe is the models do not recognise more than two single-syllable words/characters in a sequence. The recogniser stop emitting tokens after the first two words or characters.

For example:
* English: Speaking ONE ONE ONE is recognised only as ONE ONE. Speaking ONE over and over again produces no more tokens.
* 中文： Speaking 对对对对 is recognised only as 对对. Speaking more 对's produces no more tokens.

The problem goes away if single syllable words/characters are mixed with other single syllable words/characters. For example:
* ONE TWO ONE TWO ONE TWO ONE
* 对的对的对的

rouseabout:
Problem also occurs with non-streaming models, e.g.:

* https://huggingface.co/yfyeung/icefall-asr-multidataset-pruned_transducer_stateless7-2023-05-04
* https://huggingface.co/pkufool/icefall-asr-zipformer-wenetspeech-20230615

Given this happens across different datasets and training runs, is there something about the zipformer model architecture that intentionally limits output of many identical single-syllable word/character tokens?

[audio-samples-en-zh.zip](https://github.com/k2-fsa/icefall/files/13234437/audio-samples-en-zh.zip)

joazoa:
I also experience this issue.

danpovey:
Can it relate to the stateless transducer context length of 2?  It doesn't
know how many copies of the symbol it has already emitted and guesses it
has already output them.
At one point we experimented with augmenting the context with a
repeat-count, but we didn't end up merging it.

rouseabout:
Thanks for your comments Dan. I retrained librispeech with larger context lengths and ran some experiments.

> Can it relate to the stateless transducer context length of 2?  It doesn't
> know how many copies of the symbol it has already emitted and guesses it
> has already output them.

Yes. Increasing context length helps.

> At one point we experimented with augmenting the context with a
> repeat-count, but we didn't end up merging it.
> […](#)

Was it this patch: https://github.com/k2-fsa/icefall/pull/868 ?

joazoa:
[...]

PR #1393, opened 2023-11-27
Zipformer-DictinctK: Differentiating Repeated Characters

yfyeung:
Keeping a separate vector as an nn.Parameter in the model, and adding it times k/(1+k) to the embedding before the ReLU during training and decoding.

It only supports greedy_search and greedy_search_batch.

### DIFF
--- a/egs/aishell/ASR/zipformer_distinctK/__init__.py
+++ b/egs/aishell/ASR/zipformer_distinctK/__init__.py
@@ -1,0 +1,1 @@
+../zipformer/__init__.py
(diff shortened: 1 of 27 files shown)


rouseabout:
Will this model change help with the kind of problem raised in issue https://github.com/k2-fsa/icefall/issues/1356 ? Thanks.

yfyeung:
> Will this model change help with the kind of problem raised in issue #1356 ? Thanks.

Perhaps no. This is designed to address the endless repeat loops caused by repeated characters.

PR #1217, opened 2023-08-14
Update padding modified beam search

ezerhouni:
This PR follows the discussion in https://github.com/k2-fsa/icefall/issues/1206

PR #1398, opened 2023-12-04
fix initial states

LoganLiu66:
To be consistent with https://github.com/k2-fsa/icefall/pull/1217


yaozengwei:
> To be consistent with #1217

Thanks a lot!

PR #1403, opened 2023-12-07
Fix Blankskip bug.

kobenaxie:


yfyeung:
**1 1** 1 0

**1 1 1** 1

**1** 1 0 0

max reduced frames for each utt are [2, 1, 3] instead of [1, 1, 1]

kobenaxie:
> **1 1** 1 0
> 
> **1 1 1** 1
> 
> **1** 1 0 0
> 
> limit_lens, which is the maximum reduced frames for each utt, is [2, 1, 3] instead of [1, 1, 1]

Yes, but `limit_lens + y_lens` can not greater than `x_lens`.
For example, if `x_lens = [10, 12], y_lens = [1, 2]`, then `T = 12`, and `limit_lens = T - y_lens = [11, 10]` , where `limit_lens[0] = 11 > x_lens[0] = 10`, which is ilegal.

---
Besides, frame reduce should be performed on valid frames(without padding), so should perform mask `non_blank_mask = non_blank_mask & ~padding_mask`, as the ctc peak in padding part is not an actual peak.

yfyeung:
> For example, if x_lens = [10, 12], y_lens = [1, 2], then T = 12, and limit_lens = T - y_lens = [11, 10] , where limit_lens[0] = 11 > x_lens[0] = 10, which is ilegal.
limit_lens[0] = 11 < T = 12, which is legal there.

PR #1422, opened 2023-12-21
Update prepare.sh

Zth9730:
fix the bug in line 251:
1、 del the additional blank
2、correct the spell error of "new_vocab_size"


csukuangfj:

PR #1430, opened 2023-12-25
Update TTS export-onnx.py scripts for handling variable token counts

ahazned:
TTS/VITS training examples assume a hardcoded and constant token count of 79 when exporting the output models to onnx. Although this works for these recipes, I think it's better to make these counts variable for users training on different phone sets.


yaozengwei:
Thanks.

yaozengwei (approved):

PR #1428, opened 2023-12-25
Recipes for open vocabulary keyword spotting

pkufool:
This is a initial version of decoder for open vocabulary keyword spotting system, the idea is almost the same as the context biasing system we proposed before, I improve the `ContextGraph`  to make users can trade off `recall` and  `precision` easily.

I also trained some small zipformer models (around 3M parameters) on gigaspeech (for English) and wenetspeech (for Chinese) for keyword spotting purpose, will update the results and models in the following commits soon.

PR #1443, opened 2023-12-29
Fix finalize error in ContextGraph implementation.

maxwellzh:
For Aho-Corasicks algorithm, at the end of beam search, we should subtract scores of edges `state.node_score` only if current state is not an end of word.
Fix is simple.
cc @pkufool

### DIFF
--- a/icefall/context_graph.py
+++ b/icefall/context_graph.py
@@ -240,8 +240,11 @@ class ContextGraph:
           the score is zero, otherwise the score is the score of a implicit fail arc
           to root. The next state is always root.
         """
-        # The score of the fail arc
-        score = -state.node_score
+        if state.is_end:
+            score = 0.0
+        else:
+            # The score of the fail arc
+            score = -state.node_score
         return (score, self.root)
 
     def draw(


pkufool:
@maxwellzh  Thanks!, but the fix is not correct.  I have some major improvements in this PR https://github.com/k2-fsa/icefall/pull/1428 .

maxwellzh:
@pkufool plz correct me if I'm wrong, 
1. node.node_score is the cumulative bonus of all edges from root to current node. This score is used to encourage beam search going along word token path
2. And there is an output_score as bonus for word completion. This score encourages beam search to get completed word.

So at finalize(), why you would like to subtract all bonus of edges for a completed word?

e.g.
For a graph with single edge (1, 2, 3), weight 1.0

At finalize(), path [1,2,3] is expected to have 1.0+1.0 bonus, which without the fix would be 1.0 instead. Or is this by design?

pkufool:
@maxwellzh  Sorry, I am not quite follow your example. The idea is in current implementation all bonus are contained in the `node.output_score` already, you can have a look at the test cases in `context_graph.py`,  it will draw the graph into pdf format, you can go through all cases step by step.

maxwellzh:
@pkufool Sry I misunderstood the implementation. Thanks for the clarification :)

PR #1448, opened 2024-01-04
streaming_decode.py, relax the audio range from [-1,+1] to [-10,+10]

vesis84:
- some AudioTransform classes produce audio signals out of range [-1,+1]
   - Resample produced 1.0079
   - The range [-10,+10] was chosen to still be able to reliably distinguish from the [-32k,+32k] signal...
- this is related to : https://github.com/lhotse-speech/lhotse/issues/1254

### DIFF
--- a/egs/aishell/ASR/pruned_transducer_stateless7_streaming/streaming_decode.py
+++ b/egs/aishell/ASR/pruned_transducer_stateless7_streaming/streaming_decode.py
@@ -342,7 +342,12 @@ def decode_dataset(
         assert audio.dtype == np.float32, audio.dtype
 
         # The trained model is using normalized samples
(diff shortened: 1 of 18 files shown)


csukuangfj:
Thanks! Looks good to me.

vesis84:
fixed with `black`

ezerhouni:
@csukuangfj Should we propagate this fix to Sherpa ?

csukuangfj:
> @csukuangfj Should we propagate this fix to Sherpa ?

Could you remind me whether we have such a check in sherpa?

ezerhouni:
@csukuangfj My bad, ```normalize_samples``` in sherpa is used for kaldi style features

PR #1450, opened 2024-01-05
Use shuffled LibriSpeech cuts instead

JinZr:


JinZr:
|     Sys. ID       | test-clean | test-other |       comment     |
|-------------------|------------|------------|-------------------|
|1. Baseline System |    2.27    |    5.10    |--epoch 30 --avg 9 |
|2. Baseline System |    2.23    |    4.96    |--epoch 40 --avg 16|
|3. Shuffled Cuts.  |    2.25    |    5.15    |--epoch 30 --avg 10|
|4. Shuffled Cuts.  |    2.24    |    5.02    |--epoch 40 --avg 17|

All systems were decoded using greedy search with the default decoding configuration.

danpovey:
OK.  
I'd rather leave the old code in comments though, in case anyone is using an older setup, and to make them aware of the possibility of doing this kind of thing.

JinZr:
noted, the latest pr should do that

yaozengwei (approved):

PR #1468, opened 2024-01-19
Fix buffer size of DynamicBucketingSampler

yfyeung:
When `max_duration` is larger than 1500, DynamicBucketingSampler will cause `ZeroDivisionError: division by zero`.


JinZr (approved):
LGTM 🤔

PR #1482, opened 2024-01-30
Add doc about FST-based CTC forced alignment.

csukuangfj:
It is based on [CTC FORCED ALIGNMENT API TUTORIAL](https://pytorch.org/audio/main/tutorials/ctc_forced_alignment_api_tutorial.html) from [torchaudio](https://github.com/pytorch/audio), but we are using
an FST-based approach.

I can produce identical output with [torchaudio](https://github.com/pytorch/audio) using https://github.com/k2-fsa/kaldi-decoder.
Screenshot 2024-01-30 at 19 31 36

I am refactoring the code and will prepare at least two colab notebooks.


whaozl:
Hi, The align tool can  make the word time stamp is accurate on the begin and end postion ?

csukuangfj:
> Hi, The align tool can make the word time stamp is accurate on the begin and end postion ?

It depends on what model you use.

You can have a look at https://pytorch.org/audio/main/tutorials/ctc_forced_alignment_api_tutorial.html
We can produce identical results with torchaudio using the same model.

lifeiteng:
@csukuangfj Will it be completed soon?

csukuangfj:
> @csukuangfj Will it be completed soon?

Yes. I am working on it now.

lifeiteng:
@csukuangfj  Has k2-based approach been forgot?
截屏2024-06-12 19 50 39

csukuangfj:
No, it is `TODO`.
Please use the first approach at present or you can add the second approach with k2 by yourself.
All APIs you need are there. You only need to combine them.

cageyoko:
@csukuangfj  Sorry to bother you, and look forward to your reply. Recently I compared several different alignment methods, such as TorchAudio(ctc), whisperX and funasr, and found that none of them were as good as Kaldi-based alignment. The conclusion is consistent with this paper "https://www.isca-archive.org/interspeech_2024/rousso24_interspeech.pdf", do you have some advice or todo in alignment in k2-fas project.

danpovey:
Kaldi's TDNN systems have limited context, which may give better alignment.  (Or GMMs may give even more precise alignments as they have even less context).  I'm not sure that alignment is a super big priority in k2-fsa right now, is there a specific type of application  you have in mind?

cageyoko:
Thanks to Daniel for solving a problem I have had for a long time (less context is more important for alignment?). Yes. I have been working in spoken pronunciation scoring, which relies heavily on the accuracy of phoneme or word alignment.

PR #1485, opened 2024-02-01
Fixing torch.ctc err

teowenshen:
`F.ctc_loss` throws error when batch size `N` is `1`. Using `torch.long` as recommended by the documentation fixes this error. 

```python
import torch
N = 1
T = 50
S = 31
V = 20
log_probs = torch.randn(T, N, V, device="cuda:0").log_softmax(2)
targets = torch.randint(1, 20, (S,), device="cuda:0", dtype=torch.int32)
input_lengths = torch.full((N,), T, device="cuda:0", dtype=torch.int32)
target_lengths = torch.tensor([S,], device="cuda:0", dtype=torch.int32)

torch.nn.functional.ctc_loss(log_probs, targets, input_lengths, target_lengths)
```

throws:
```python
RuntimeError: Expected tensor to have CPU Backend, but got tensor with CUDA Backend (while checking arguments for cudnn_ctc_loss)
```

### DIFF
--- a/egs/librispeech/ASR/zipformer/model.py
+++ b/egs/librispeech/ASR/zipformer/model.py
@@ -164,9 +164,9 @@ class AsrModel(nn.Module):
 
         ctc_loss = torch.nn.functional.ctc_loss(
             log_probs=ctc_output.permute(1, 0, 2),  # (T, N, C)
-            targets=targets.to(torch.long),
-            input_lengths=encoder_out_lens.to(torch.long),
-            target_lengths=target_lengths.to(torch.long),
+            targets=targets.cpu(),
+            input_lengths=encoder_out_lens.cpu(),
+            target_lengths=target_lengths.cpu(),
             reduction="sum",
         )
         return ctc_loss
--- a/egs/librispeech/ASR/zipformer/model.py
+++ b/egs/librispeech/ASR/zipformer/model.py
@@ -164,9 +164,9 @@ class AsrModel(nn.Module):
 
         ctc_loss = torch.nn.functional.ctc_loss(
             log_probs=ctc_output.permute(1, 0, 2),  # (T, N, C)
-            targets=targets,
-            input_lengths=encoder_out_lens,
-            target_lengths=target_lengths,
+            targets=targets.to(torch.long),
+            input_lengths=encoder_out_lens.to(torch.long),
+            target_lengths=target_lengths.to(torch.long),
             reduction="sum",
         )
         return ctc_loss


egs/librispeech/ASR/zipformer/model.py
```diff
@@ -164,9 +164,9 @@ def forward_ctc(
 
         ctc_loss = torch.nn.functional.ctc_loss(
             log_probs=ctc_output.permute(1, 0, 2),  # (T, N, C)
-            targets=targets,
-            input_lengths=encoder_out_lens,
-            target_lengths=target_lengths,
+            targets=targets.to(torch.long),
```
csukuangfj, line 167: could you try
```
targets=targets.cpu(),
input_lengths=encoder_out_lens.cpu(),
target_lengths=target_lengths.cpu(),
```

It is a pity that PyTorch does not document clearly that only log_probs is required to be on GPU.

csukuangfj:
The error is from the following line
https://github.com/pytorch/pytorch/blob/2b48891e62e5c4b57c8cac92cee5eb71228a203a/aten/src/ATen/native/cudnn/LossCTC.cpp#L146
```
  checkBackend(c, {*targets}, Backend::CPU);
```

You can see that it expects `targets` on CPU.

---

As for the lengths, https://github.com/pytorch/pytorch/blob/2b48891e62e5c4b57c8cac92cee5eb71228a203a/aten/src/ATen/native/cudnn/LossCTC.cpp#L221

```
  Tensor ilc = input_lengths.to(Device(at::kCPU), at::kLong).contiguous();
  Tensor tlc = target_lengths.to(Device(at::kCPU), at::kLong).contiguous();
```

They are converted to `kLong` and moved to `CPU`.

So the only thing that needs to be changed is to move `targets` to CPU.

yfyeung:
Disable cudnn also works for me.
```python
        with torch.backends.cudnn.flags(enabled=False):
            ctc_loss = torch.nn.functional.ctc_loss(
                log_probs=ctc_output.permute(1, 0, 2),  # (T, N, C)
                targets=targets,
                input_lengths=encoder_out_lens,
                target_lengths=target_lengths,
                reduction="sum",
            )
```

teowenshen:
I have followed @csukuangfj 's advice and moved the targets and lengths to CPU. There were some comments that say cudnn is the only deterministic implementation in PyTorch. 

Surprisingly, in my case the error wasn't thrown until when I had a last batch in the validation loop that had only 1 sample. But it doesn't seem to be related to the batch size after all.

csukuangfj:
Thanks!

PR #1503, opened 2024-02-19
Fixed softlink for `ljspeech` recipe

JinZr:


csukuangfj:
Thanks!

Looks like we need a CI test for tts.

csukuangfj (approved):

PR #1512, opened 2024-02-20
Add recipe for fine-tuning Zipformer with adapter

marcoyang1998:
This PR supports fine-tuning a Zipformer transducer model with adapters (https://arxiv.org/pdf/1902.00751.pdf). The original model parameters are untouched and we only update the parameters in the adapters on the new domain.

The following table shows the WERs of different models on GigaSpeech test sets in a domain adaptation setting.

| Model config | trainable params | dev | test | 
| ----------- | ------ | ---------- | ---------- |
| pretrained (without fine-tuning) |  - |   20.06 | 19.27 |
| fine-tune whole model | 66M |  13.56 | 13.64 | 
| fine-tune with adapters | 1.5M |  15.05 | 15.18 | 

By only introducing around **2%** trainable parameters, the model with adapters achieves much lower WERs than the baseline model without adaptation.


egs/librispeech/ASR/zipformer_adapter/train.py
```diff
@@ -1018,6 +1018,12 @@ def train_one_epoch(
         be set to 0.
     """
     model.train()
+    # set modules except adapters to eval mode
+    for name, m in model.named_modules():
+        if "adapter" in name:
+            m.training = True
```
kobenaxie, line 1024: Why using `m.training = True` rather than `m.requires_grad_(False)` here ?

marcoyang1998: This is only to make sure other modules are in `eval()` mode (e.g. Dropout is deactivated). We already set `p.requires_grad = False` after initializing the model.

kobenaxie: Ok, Thank you for your reply ~

marcoyang1998:
After training, the model can be tested on both the original and the new domain.

To deactivate the adapters after training, set `--use-adapters False`. The performance on the original domain should be untouched.

Issue #1531, opened 2024-03-06
Errors while preparing Common Voice dataset


ChrystianKacki:
I'm trying to build zipformer model with Common Voice dataset.
I'm using script `prepare.sh`, located in [https://github.com/k2-fsa/icefall/tree/master/egs/commonvoice/ASR](https://github.com/k2-fsa/icefall/tree/master/egs/commonvoice/ASR).
I downloaded the data manually, because pointing URL is now forbidden and requires to submit e-mail address or to be logged in.
When I execute the script, at `Stage 9: Prepare BPE based lang` get the following error: 

***RuntimeError: Internal: src/trainer_interface.cc(662) [(trainer_spec_.vocab_size()) == (model_proto->pieces_size())] Vocabulary size too high (500). Please set it to a value <= 6.***

Traceback (most recent call last):
- File "/mnt/c/Users/chrys/IdeaProjects/sherpa-onnx-model/model-common_voice/./local/train_bpe_model.py", line 115, in <module>
    main()
- File "/mnt/c/Users/chrys/IdeaProjects/sherpa-onnx-model/model-common_voice/./local/train_bpe_model.py", line 93, in main
    spm.SentencePieceTrainer.train(
- File "/home/chrychoo/.local/lib/python3.10/site-packages/sentencepiece/__init__.py", line 1047, in Train
    SentencePieceTrainer._Train(arg=arg, **kwargs)
- File "/home/chrychoo/.local/lib/python3.10/site-packages/sentencepiece/__init__.py", line 1040, in _Train
    return SentencePieceTrainer._TrainFromMap(new_kwargs)
- File "/home/chrychoo/.local/lib/python3.10/site-packages/sentencepiece/__init__.py", line 985, in _TrainFromMap
    return _sentencepiece.SentencePieceTrainer__TrainFromMap(args)

From the traceback, I think this occurs because `vocab_sizes` variable at the begining of script `prepare.sh` is set to `500`, and then it is passed to `train_bpe_model.py` script. From here, the code in the method `SentencePieceTrainer.train()` of `sentencepiece` library throws this error.
How to deal with this ? Can I simply change the `vocab_sizes` variable value to `6` in `prepare.sh` script ?

JinZr:
Looks like there’s an issue with the text you use to train the bpe model, could you show us whats the text looked like?

ChrystianKacki:
@JinZr Do you mean the text in file `$lang_dir/transcript_words.txt` ?

[...]

PR #1534, opened 2024-03-07
Minor fixes for the `commonvoice` recipe

JinZr:
added text norm for yue language

fixed two errors related to HLG and LG compilation


JinZr:
ref: https://github.com/k2-fsa/icefall/issues/1531

marcoyang1998 (approved):

PR #1537, opened 2024-03-08
Zipformer recipe for Cantonese dataset MDCC

JinZr:
Description of the dataset can be found at: https://arxiv.org/pdf/2201.02419.pdf

Involved PR: https://github.com/lhotse-speech/lhotse/pull/1302


egs/mdcc/ASR/zipformer/train.py
```diff
@@ -0,0 +1,1349 @@
...
+Usage:
+
+export CUDA_VISIBLE_DEVICES="0,1,2,3,4,5,6,7"
+
+./zipformer/train.py \
+  --world-size 8 \
+  --num-epochs 12 \
+  --start-epoch 1 \
+  --exp-dir zipformer/exp \
+  --training-subset L
```
marcoyang1998, line 32: Please update the usage, I think MDCC does not have `--training-subset`.

marcoyang1998 (approved):
Thanks for the PR! Left only one comment, I think we can merge it after you fix it.

PR #1538, opened 2024-03-08
Fix default value for --context-size in aishell.

csukuangfj:
The inconsistent default value of `--context-size` in aishell causes many troubles. This PR fixes that.

### DIFF
--- a/egs/aishell/ASR/pruned_transducer_stateless7_streaming/decode.py
+++ b/egs/aishell/ASR/pruned_transducer_stateless7_streaming/decode.py
@@ -250,7 +250,7 @@ def get_parser():
     parser.add_argument(
         "--context-size",
         type=int,
-        default=1,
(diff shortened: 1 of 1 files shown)


egs/aishell/ASR/pruned_transducer_stateless7_streaming/decode.py
```diff
@@ -250,7 +250,7 @@ def get_parser():
     parser.add_argument(
         "--context-size",
         type=int,
-        default=1,
+        default=2,
```
csukuangfj, line 253: See also
https://github.com/k2-fsa/icefall/blob/ae61bd4090ad3e8c981aa77cc4fc417d095962c4/egs/aishell/ASR/pruned_transducer_stateless7_streaming/train.py#L290-L295

PR #1541, opened 2024-03-11
Support different tts model types.

csukuangfj:
low (quality) -> runs faster. high (quality) -> runs slower

---

Need to test it before merging.

### DIFF
--- a/egs/ljspeech/TTS/vits/export-onnx.py
+++ b/egs/ljspeech/TTS/vits/export-onnx.py
@@ -76,7 +76,8 @@ def get_parser():
     parser.add_argument(
         "--model-type",
         type=str,
-        default="",
+        default="high",
+        choices=["low", "medium", "high"],
         help="""If not empty, valid values are: low, medium, high.
         It controls the model size. low -> runs faster.
         """,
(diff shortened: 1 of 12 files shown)


egs/ljspeech/TTS/vits/vits.py
```diff
@@ -38,6 +39,36 @@
     "hifigan_multi_scale_multi_period_discriminator": HiFiGANMultiScaleMultiPeriodDiscriminator,  # NOQA
 }
 
+LOW_CONFIG = {
```
csukuangfj, line 42: The config values are from
https://github.com/rhasspy/piper/blob/master/src/python/piper_train/__main__.py#L68

egs/ljspeech/TTS/vits/export-onnx.py
```diff
@@ -256,18 +263,6 @@ def main():
...
-    # Generate int8 quantization models
-    # See https://onnxruntime.ai/docs/performance/model-optimizations/quantization.html#data-type-selection
-
-    logging.info("Generate int8 quantization models")
-
-    model_filename_int8 = params.exp_dir / f"vits-{suffix}.int8.onnx"
-    quantize_dynamic(
-        model_input=model_filename,
-        model_output=model_filename_int8,
-        weight_type=QuantType.QUInt8,
```
csukuangfj, line 268: Quantizing using quint8 is very slow at run time, so we removed it.

PR #1556, opened 2024-03-18
allow export of onnx-streaming-models with other than 80dim input features

KarelVesely84:
- i could also add surfacing of `--feature-dim` into train.py, 
- and also feature extraction had to become configurable in [compute_fbank_librispeech.py](https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/local/compute_fbank_librispeech.py) and [streaming_decode.py](https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/zipformer/streaming_decode.py), if you are interested in having the recipe with configurable feature extraction...


csukuangfj:
Thanks! We can change other files if needed.
(The latest ./zipformer folder is more likely to be used than other folders, so it is fine to change only ./zipformer at present.)

csukuangfj (approved):

Issue #1553, opened 2024-03-16
zipformer-adapter streaming_forward without adapters.


kobenaxie:
All adapters are only used in `forward()`, but not in `streaming_forward`, please fix it.
https://github.com/k2-fsa/icefall/blob/2dfd5dbf8be72fce434445268e415bcdfa4b3028/egs/librispeech/ASR/zipformer_adapter/zipformer.py#L933

marcoyang1998:
Thanks for pointing it out. I will try to add the support for `streaming_forward` this week.

kobenaxie:
Hi @marcoyang1998 , another question, when finetuning `adapter` or `lora`, is there any difference between `optimizer with whole model parameters` and `optimizer with only adapter/lora module paramters` ?

marcoyang1998:
They should be equivalent as only the parameters in `lora` or `adapter` will have gradients, but the latter could be more efficient (not tested though) as it won't traverse all model parameters.

marcoyang1998:
Fixed in #1560

PR #1560, opened 2024-03-19
Fix adapter in streaming_forward

marcoyang1998:
Fix for #1553.

### DIFF
--- a/egs/librispeech/ASR/zipformer_adapter/export-onnx.py
+++ b/egs/librispeech/ASR/zipformer_adapter/export-onnx.py
@@ -27,11 +27,13 @@ popd
 
 2. Export the model to ONNX
 
-./zipformer/export-onnx.py \
+./zipformer_adapter/export-onnx.py \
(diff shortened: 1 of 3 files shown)


marcoyang1998:
To finetune a causal zipformer with adapter, just specify `--causal 1` when using `finetune.py`.

The trained model can be exported using the following command (you can adjust the chunk size and left context when exporting the model):
```bash
./zipformer_adapter/export.py \
    --exp-dir ./zipformer_adapter/exp_causal1_giga_finetune1_adapters1_adapter_dim16 \
    --epoch 4 \
    --avg 1 \
    --use-averaged-model 1 \
    --chunk-size 16 \
    --left-context-frames 128 \
    --causal 1 \
    --tokens icefall-asr-librispeech-streaming-zipformer-2023-05-17/data/lang_bpe_500/tokens.txt \
    --use-adapters 1 \
    --adapter-dim 16 \
    --jit 1
```

PR #1582, opened 2024-04-05
ASR SEAME Recipe

AmirHussein96:
This pull request is for the SEAME recipe and includes the following:

1. Standard Zipformer with RNNT loss
2. Zipformer with hybrid autoregressive transducer (HAT) loss
3. Proposed Zipformer with hybrid autoregressive transducer (HAT) with an auxiliary LID branch and blank sharing for synchronization between ASR and LID predictions.

### DIFF
--- a/egs/seame/ASR/local/cer.py
+++ b/egs/seame/ASR/local/cer.py
@@ -9,6 +9,7 @@ This file cer from icefall decoded "recogs" file:
 """
 
 import argparse
+
 import jiwer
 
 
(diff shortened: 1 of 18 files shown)


egs/librispeech/ASR/zipformer/export.py
JinZr, line 1: hi amir, thank you for your contribution!

but im wondering perhaps you want to use a separate `export.py` rather than applying modification to the one used by librispeech?

AmirHussein96: Since both are based on same zipformer architecture, I think they can use the same export, the difference is in the parameters. I could also create a separate one if needed.

JinZr:
A separate one would be better, otherwise backward compatibility might be
broken since you have also modified the default parameters of the export
script.

PR #1584, opened 2024-04-09
fix run.sh script in wenetspeech KWS

yh646492956:
Resolve the issue of finetune. py script parameter passing failure due to insufficient addition of \ in wenetspeech KWS run.sh


csukuangfj:
Thank you for your first-time contribution!

PR #1603, opened 2024-04-23
Clear log handlers before setup

yfyeung:
This PR adds functionality to clear existing log handlers before setting up new ones. This feature addresses issues in environments like `torchrun`, where default loggers cannot be easily overridden. By clearing previous handlers, we enable full customization and flexibility of logging configurations, ensuring that setups can be dynamically adjusted without being constrained by pre-existing environment settings.


JinZr (approved):
LGTM

PR #1607, opened 2024-04-25
fix error in length computing

zzasdf:
This PR fixs the bug in librispeech/SSL, which computes x_lens using padding_mask in a wrong way.
Fortunately, it did not cause any problem.


yfyeung (approved):
LGTM

PR #1617, opened 2024-05-06
Remove duplicate logging initialization logic in utils.py

yfyeung:
Since we already have `force=True` to achieve the same logic, explicitly resetting logging is no longer necessary.


csukuangfj (approved):
Thanks!

PR #1630, opened 2024-05-20
Initialize BiasNorm bias with small random values

yaozengwei:
This aims to fix the inf issue when the input vector is all-zero.


JinZr (approved):
LGTM!

PR #1605, opened 2024-04-24
update speechio whisper ft results

yuekaizhang:
Fine-tuning whisper-large-v2 using multi-hans-zh dataset, exclude datatang 200h (which is not open sourced any more), updating wenetspeech (according to https://github.com/wenet-e2e/WenetSpeech/discussions/54)

Currently, rank 8 according to https://github.com/SpeechColab/Leaderboard

| Rank 排名 | Model 模型 | CER 字错误率 | Date 时间 |
| --- | --- | --- | --- |
| 1 | ximalaya_api_zh | 1.72% | 2023.12 |
| 2 | aliyun_ftasr_api_zh | 1.85% | 2023.12 |
| 3 | microsoft_batch_zh | 2.40% | 2023.12 |
| 4 | bilibili_api_zh | 2.90% | 2023.09 |
| 5 | tencent_api_zh | 3.18% | 2023.12 |
| 6 | iflytek_lfasr_api_zh | 3.32% | 2023.12 |
| 7 | aispeech_api_zh | 3.62% | 2023.12 |
| 8 | **whisper-large-ft-v1** | **4.45%** | 2024.04 |
| 9 | baidu_pro_api_zh | 7.29% | 2023.12 |

|                | Split         | Greedy Search |
|----------------|---------------|---------------|
| Datasets       |               |               |
| alimeeting     | eval          | 23.45         |
| alimeeting     | test          | 25.42         |
| aishell-1      | dev           | 0.78          |
| aishell-1      | test          | 0.83          |
| aishell-2      | dev           | 2.75          |
| aishell-2      | test          | 2.93          |
| aishell-4      | test          | 17.11         |
| magicdata      | dev           | 2.68          |
| magicdata      | test          | 2.33          |
| kespeech-asr   | dev phase1    | 4.97          |
| kespeech-asr   | dev phase2    | 2.02          |
| kespeech-asr   | test          | 6.34          |
| WenetSpeech    | dev           | 5.06          |
| WenetSpeech    | test meeting  | 8.38          |
| WenetSpeech    | test net      | 6.94          |

PR #1648, opened 2024-06-12
add Chinese distill-whisper fine-tuning results 

yuekaizhang:
In #1605, we fine-tuned whisper using 14k hours Chinese data. This PR added decoding results for distill-whisper fine-tuning experiment.

Instead of actually using distillation loss for training, the model structure and parameter initialization method from the distill-whisper paper (https://arxiv.org/abs/2311.00430) were adopted: only the first and last layers of the decoder were retained.

Accuracy:
Distill-whisper is slightly worse comparing with norm whisper.   
| Model  | CER (Average SPEECH_IO_TEST_SET 01-26) | Training Set |
| --- | --- | --- |
| **whisper-large-ft-v1** | **4.32%** | multi-hans-zh (about 14k hours) |
| **whisper-large-ft-v1-distill** | **4.71%** | multi-hans-zh (about 14k hours) |

Speed: 
Every decoding step could acclerate about 4x comparing with the original decoder.
Norm whisper: 32 decoder layers
image
Distill-whisper: 2 decoder layers
image

For a quick test: https://huggingface.co/yuekai/icefall_asr_multi-hans-zh_whisper/blob/main/test_model.py

### DIFF
--- a/egs/multi_zh-hans/ASR/RESULTS.md
+++ b/egs/multi_zh-hans/ASR/RESULTS.md
@@ -6,10 +6,11 @@
 
 Character Error Rates (CERs) listed below are produced by the checkpoint of the second epoch using greedy search.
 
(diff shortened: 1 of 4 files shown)


JinZr (approved):
LGTM, thanks!

PR #1663, opened 2024-06-22
Cast grad_scale in whiten to float

teowenshen:
`grad_scale` in Whiten throws error if it is of type `ScheduledFloat`.


yaozengwei (approved):

yaozengwei:
Thanks!

PR #1689, opened 2024-07-12
Fix usages of returned losses after adding attention-decoder in zipformer

yaozengwei:


pkufool (approved):
LGTM

egs/wenetspeech/KWS/zipformer/finetune.py
```diff
@@ -344,40 +342,6 @@ def compute_loss(
     return loss, info
 
 
-def compute_validation_loss(
```
yaozengwei, line 347: Found duplicated codes here

PR #1691, opened 2024-07-15
Refactor ctc greedy search.

csukuangfj:
Use torch.unique_consecutive() to avoid reinventing the wheel.

See also #1690


yaozengwei:
Fantastic!

PR #1712, opened 2024-08-08
librispeech_decode : split `save_results()` -> `save_asr_output()` + `save_wer_results()`

KarelVesely84:
THE IDEA
- the idea is to support `--skip-scoring` argument passed to a decoding script
- created for Transducer decoding (non-streaming, streaming) and CTC decoding (non-streaming)

OTHER CHANGES
- also added `--label` for extra label in `streaming_decode.py`
- and also added `set_caching_enabled(True)`
   - this has no effect on librispeech, but it leads to faster runtime on DBs with long recordings (assuming `librispeech/zipformer` scripts are the example scripts for other setups)
- replaced some '-' with '_' and vice versa, to get more readable filenames with results

TESTING
- still need to run the code to make sure it works properly (it is meld-ed from other recipe)
- at this point, interested in getting early feedback from you...

Thank you.

### DIFF
--- a/egs/librispeech/ASR/zipformer/ctc_decode.py
+++ b/egs/librispeech/ASR/zipformer/ctc_decode.py
@@ -120,6 +120,7 @@ import sentencepiece as spm
 import torch
 import torch.nn as nn
 from asr_datamodule import LibriSpeechAsrDataModule
+from lhotse import set_caching_enabled
 from train import add_model_arguments, get_model, get_params
 
 from icefall.checkpoint import (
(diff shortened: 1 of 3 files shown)


KarelVesely84:
Hi @csukuangfj ,
would it be ok to merge somethig like this PR ?
(adding `--skip-scoring` to the example decode scripts, it's handy for processing evaluation data)

What should be the next step ?

Best regards,
Karel

csukuangfj:
Looks good to me. Thank you for your contribution!

Issue #1352, opened 2023-10-30
Zipformer2 with CTC is hard to train


joazoa:
I am playing a bit with the CTC option in the zipformer 2, with the largest model from the documentation.
It trained well for a first dataset but when I try another dataset, the training stops.

I have tried reducing the LR, increasing the warmup period, disabling the FP16. changing the max duration, removing any augmentations in lhotse, reducing maximum file duration, removing specaugment and musan and changing the worldsize. 

Do you have any suggestions on what I could try next ?

2023-10-29 23:40:46,833 INFO [train.py:1034] (2/8) Epoch 1, batch 0, loss[loss=8.228, simple_loss=7.432, pruned_loss=7.424, ctc_loss=5.236, over 20568.00 frames. ], tot_loss[loss=8.228, simple_loss=7.432, pruned_loss=7.424, ctc_loss=5.236, over 20568.00 frames. ], batch size: 95, lr: 2.25e-02, grad_scale: 1.0
2023-10-29 23:40:46,833 INFO [train.py:1057] (2/8) Computing validation loss
2023-10-29 23:40:54,294 INFO [train.py:1066] (2/8) Epoch 1, validation: loss=inf, simple_loss=7.453, pruned_loss=7.416, ctc_loss=inf, over 901281.00 frames.
2023-10-29 23:40:54,295 INFO [train.py:1067] (2/8) Maximum memory allocated so far is 22148MB
2023-10-29 23:40:59,516 INFO [scaling.py:979] (2/8) Whitening: name=encoder.encoders.5.encoder.layers.1.nonlin_attention.whiten2, num_groups=1, num_channels=256, metric=7.26 vs. limit=5.0
2023-10-29 23:41:11,380 INFO [scaling.py:199] (2/8) ScheduledFloat: name=encoder.encoders.3.encoder.layers.0.bypass.scale_min, batch_count=0.0, ans=0.9
2023-10-29 23:41:16,030 INFO [scaling.py:199] (2/8) ScheduledFloat: name=encoder.encoders.2.encoder.layers.2.conv_module1.balancer2.min_abs, batch_count=106.66666666666667, ans=0.2016
2023-10-29 23:41:26,737 INFO [scaling.py:199] (2/8) ScheduledFloat: name=encoder.encoders.4.encoder.layers.2.feed_forward1.out_proj.dropout_p, batch_count=106.66666666666667, ans=0.29893333333333333
[...]

PR #1713, opened 2024-08-12
fix the CTC zipformer2 training

KarelVesely84:
- too many supervision tokens
- change filtering rule to `if (T - 2) < len(tokens): return False`
- this prevents inf. from appearing in the CTC loss value (empirically tested)

### DIFF
--- a/egs/librispeech/ASR/zipformer/train.py
+++ b/egs/librispeech/ASR/zipformer/train.py
@@ -1300,9 +1300,11 @@ def run(rank, world_size, args):
         T = ((c.num_frames - 7) // 2 + 1) // 2
         tokens = sp.encode(c.supervisions[0].text, out_type=str)
 
-        if T < len(tokens):
+        # For CTC `(T - 2)  < len(tokens)` is needed. otherwise inf. in loss appears.
+        # For Transducer `T < len(tokens)` was okay.
+        if (T - 2) < len(tokens):
             logging.warning(
-                f"Exclude cut with ID {c.id} from training. "
+                f"Exclude cut with ID {c.id} from training (too many supervision tokens). "
                 f"Number of frames (before subsampling): {c.num_frames}. "
                 f"Number of frames (after subsampling): {T}. "
                 f"Text: {c.supervisions[0].text}. "


KarelVesely84:
workflow: https://github.com/k2-fsa/icefall/actions/runs/10348851808/job/28642009312?pr=1713 
```
fatal: unable to access 'https://huggingface.co/Zengwei/icefall-asr-librispeech-zipformer-2023-05-15/': Recv failure: Connection reset by peer
```

KarelVesely84:
Hi @csukuangfj ,
how about this one ? Is @yaozengwei testing it currently ?

It is solving the issue https://github.com/k2-fsa/icefall/issues/1352 

My theory is that CTC uses 2 extra symbols at beginning/end of label sequence.
So, the label-length limit needs to be lowered by 2 symbols to accomodate that.

Best regards
Karel

csukuangfj:
Sorry for the late reply.


Could you analyze the wave that causes inf loss?
Is it too short?

Does it contain only a single word or does it contain nothing at all?

KarelVesely84:
Hi,
the problematic utterance contained many words:
`(num_embeddings, supervision_length, difference a-b) = (34, 33, 1)`

text:
['▁O', 'f', '▁all', '▁.', '▁P', 'ar', 'li', 'a', 'ment', '▁,', '▁Co', 'un', 'c', 'il', '▁and', '▁Co', 'm', 'm', 'i', 's', 's', 'ion', '▁are', '▁work', 'ing', '▁to', 'ge', 'ther', '▁to', '▁de', 'li', 'ver', '▁.']

It seems like a better set of BPEs could reduce the number of supervision tokens.
Nevertheless, this would only hide the ``inf.'' problem for CTC.

I believe the two extra tokens for the CTC loss are the `<bos/eos>` 
that get (pre-,ap-)pended to the supervision sequence,
hence the `(T - 2)`.

Best regards
Karel

csukuangfj:
> the problematic utterance contained many words:

Thanks for sharing! Could you also post the duration of the corresponding wave file?

KarelVesely84:
This is the corresponding `Cut`:

`MonoCut(id='20180612-0900-PLENARY-3-59', start=557.34, duration=1.44, channel=0, supervisions=[SupervisionSegment(id='20180612-0900-PLENARY-3-59', recording_id='20180612-0900-PLENARY-3', start=0.0, duration=1.44, channel=0, text='Of all . Parliament , Council and Commission are working together to deliver .', language='en', speaker='None', gender='male', custom={'orig_text': 'of all. Parliament, Council and Commission are working together to deliver.'}, alignment=None)], features=Features(type='kaldi-fbank', num_frames=144, num_features=80, frame_shift=0.01, sampling_rate=16000, start=557.34, duration=1.44, storage_type='lilcom_chunky', storage_path='data/fbank/voxpopuli-asr-en-train_feats/feats-59.lca', storage_key='395124474,12987', recording_id='None', channels=0), recording=Recording(id='20180612-0900-PLENARY-3', sources=[AudioSource(type='file', channels=[0], source='/mnt/matylda6/szoke/EU-ASR/DATA/voxpopuli/raw_audios/en/2018/20180612-0900-PLENARY-3_en.ogg')], sampling_rate=16000, num_samples=139896326, duration=8743.520375, channel_ids=[0], transforms=None), custom={'dataloading_info': {'rank': 3, 'world_size': 4, 'worker_id': None}})`

It is a 1.44 sec long cut in a very long recording (2.42 hrs).
And the 1.44 sec is very little to pronounce all the words in the reference text : 
"Of all . Parliament , Council and Commission are working together to deliver ."

csukuangfj:
yes, I think it should be good to filter out such kind of data.

KarelVesely84:
Hello, is there something needed for this to merge from my side ?
K.

KarelVesely84:
Well, without that `(T - 2)` change i was getting inf. value from the CTC loss.
There sholud be no inf. even if the data are prepared badly.

I also did not find any trace of the extra CTC symbols <bos> or <eos> or similar in the scripts.
The `torch.nn.functional.ctc_loss(.)` is getting the same set of symbols as transducer loss.

Could you try to reproduce the issue by adding a training example with a very lenghty transcript ?
(or I can create a branch to demonstrate it, say repeating the librispeech transcript 100x, just to make the error appear)

Best regards,
Karel

Vindicator645:
just letting you know that i had the same issue and this solved the problem

PR #1714, opened 2024-08-12
fix open-commands path in kws

yulianjie:
https://github.com/k2-fsa/icefall/blob/3b257dd5ae79bff99470ec1cbbeaa8fae84f956a/egs/wenetspeech/KWS/prepare.sh#L66
As shown in the figure below, the path of `open-commands->scripts` is incorrect, this MR just fix open-commands path in kws.
image


pkufool (approved):
LGTM

pkufool:
@yulianjie  Thanks! Merging.

PR #1757, opened 2024-09-26
Fix the batch_idx_train mismatch for model averaging

zzasdf:
The batch_idx_train save in epoch-*.pt may be different from the batch_idx_train when the model_avg is update. This may cause instability in the decode results, especially when params.avg is small and the train step for an epoch is not very large


yfyeung:
This is exactly a bug when performing model averaging. Could you help to review this PR? @csukuangfj

yaozengwei (approved):

yaozengwei:
Thanks a lot!

Issue #1709, opened 2024-08-06
early context injection


EmreOzkose:
Hi,

Are you planning to add early context injection approach ([Improving Neural Biasing for Contextual Speech Recognition by Early Context Injection and Text Perturbation](https://arxiv.org/pdf/2407.10303)) to Icefall?

JinZr:
Hi,

Sorry we dont have such plans in sight so far. 

best
jin

danpovey:
Ruizhe, how hard would it be to make a PR at least as a demo?

huangruizhe:
I think it won't be that hard. I can start working on the librispeech recipe on the side.

PR #1763, opened 2024-09-30
[WIP] neural biasing with early context injection and text perturbation

huangruizhe:
This PR implements the paper: [Improving Neural Biasing for Contextual Speech Recognition by Early Context Injection and Text Perturbation
](https://arxiv.org/abs/2407.10303). This also addresses issue #1709 .

This codes can replicate the paper, but it is still a WIP -- needing codes refactoring and cleaning up.

TODO:
- Make soft links to the original standard ASR recipe -- avoid copying scripts
- Expose neural biasing and so on as an external function call to contextualize the models
- Simplify the scripts for loading contexts -- only need an example for librispeech
- Get evaluation scripts for common/rare words as in the contextual librispeech setup


teinhonglo:
Hi RuiZhe,

Thanks for sharing this implementation. I noticed that this PR is still open and has not been merged yet. May I ask whether there are any known issues or blockers with the current code?

I’m interested in trying it out to reproduce the paper, so I would really appreciate any guidance on whether the current version is usable.

Thanks!
Tien-Hong

PR #1772, opened 2024-10-16
Fix the fixed num_splits bug

KIM7AZEN:
it use a fixed num_splits 1000 , should be pieces=$(find data/fbank/M_split_${num_splits} -name "cuts_M.*.jsonl.gz")


csukuangfj (approved):

PR #1775, opened 2024-10-21
minor fixes for the repo

JinZr:


icefall/utils.py
```diff
@@ -2282,13 +2282,12 @@ def time_warp(
...
-    """
+    """Apply time warping on a batch of features"""
     if time_warp_factor is None or time_warp_factor < 1:
         return features
-    assert len(features.shape) == 3, (
-        "SpecAugment only supports batches of single-channel feature matrices."
-    )
+    assert (
+        len(features.shape) == 3
+    ), "SpecAugment only supports batches of single-channel feature matrices."
```
csukuangfj, line 2290:
```suggestion
    ), f"SpecAugment only supports batches of single-channel feature matrices. {features.shape}"
```

It is helpful to print the shape when there is an assertion error.

PR #1778, opened 2024-10-21
Remove incomplete recipe: people's speech

yfyeung:
Since this recipe incomplete, and previous experiments have shown that training a model directly on People's Speech without data filtering yields very poor results, I recommend deleting this incomplete recipe.


csukuangfj (approved):
Thanks!

PR #1792, opened 2024-10-30
Fix Boolean Conversion Issue from Shell to Python

yfyeung:
Fixes a bug where the shell variable with a False value is passed to Python as a string, causing bool("False") to evaluate as True.


yfyeung:
For example:

test.py:
```python
import sys
print(type(sys.argv[1]))
```

```shell
python test.py False
```
Output:
```shell
<class 'str'>
```

csukuangfj (approved):

PR #1797, opened 2024-11-04
Performed end to end testing on the matcha recipe

JinZr:


.github/scripts/ljspeech/TTS/run-matcha.sh
yfyeung, line 1: I suggest to keep the name `inference` or `infer`. This is used in many other frameworks.

JinZr: to cope with the naming scheme of the valle and vits recipe, i added a infer.py script for performing inference on the entire test set, and the synth.py script accepts single sentence for synthesis.

yfyeung: Maybe we can combine them into one file since their logic is quate similar. Just like the implementation in https://github.com/lifeiteng/vall-e/blob/main/valle/bin/infer.py#L180-L221 and https://github.com/lifeiteng/vall-e/blob/main/valle/bin/infer.py#L223-L269.

JinZr: sounds good, i'll look into it

JinZr: the latest commit has the pattern applied to the recipe 🚀

PR #1821, opened 2024-11-30
Egs ljspeach autocast & GradScaler function deprecated

mablue:
FutureWarning: `torch.cuda.amp.autocast(args...)` is deprecated. Please use `torch.amp.autocast('cuda', args...)` instead.


csukuangfj:
Please use
```python3
import warnings
warnings.simplefilter(action='ignore', category=FutureWarning)
```
to suppress the warnings.

PR #1837, opened 2024-12-13
Refactor optimizer

zhu-han:
From Dan: "This has simplifying refactoring, but I also just added a change to diagnostics for when grads are clipped, which should help diagnose problems with models. Anyone who has a model that is not converging well or which is getting percent-clipped > 0 should try that version of optim.py, which has more informative
error messages."


csukuangfj (approved):
Thanks!

PR #1852, opened 2024-12-30
Warn instead of raising exceptions in inf-check

zhu-han:


icefall/hooks.py
```diff
@@ -40,8 +40,8 @@ def register_inf_check_hooks(model: nn.Module) -> None:
         def forward_hook(_module, _input, _output, _name=name):
             if isinstance(_output, Tensor):
                 if not torch.isfinite(_output.to(torch.float32).sum()):
-                    raise ValueError(
-                        f"The sum of {_name}.output is not finite: {_output}"
+                    logging.warning(
```
csukuangfj, line 43: Will the training continue with inf loss and inf grad?

zhu-han: Yes, like the backward code, it will just warn instead of printing the tensor and raising an exception.

csukuangfj: In that case, will it propagate the inf grad to model parameters and result in inf parameters?

csukuangfj: I see. Thanks!

csukuangfj:
Could you fix the style issues?

csukuangfj (approved):
Thanks!

PR #1862, opened 2025-01-09
Improve infinity-check

zhu-han:
1. Attach the inf-check hooks if the grad scale is getting too small.
2. Add try-catch to avoid OOM in the inf-check hooks.
3. Set warmup_start=0.1 to reduce chances of divergence.


csukuangfj (approved):
Thanks!

yfyeung (approved):

PR #1865, opened 2025-01-16
removed the erroneous ``continual'' implementation

JinZr:
the ``continual'' implementation directly uses the **gt** first-layer codec with nar generated part as its output, which is inconsistent with the original paper, hence should be removed


yfyeung:
See issue: https://github.com/lifeiteng/vall-e/issues/194

yuekaizhang:
Sorry, I didn't test the continual function before.

yuekaizhang (approved):

yfyeung (approved):

PR #1880, opened 2025-02-20
Add F5-TTS with semantic token training results

yuekaizhang:
The PR replaced the text tokens input in F5-TTS with pretrained cosyvoice2 semantic tokens. 
Or you can think of it as replacing the unet based flow matching model in the [cosyvoice2](https://arxiv.org/abs/2412.10117) with DiT.

Summary of the training results using wenetspeech4tts data：

|        Model                               | Seed-TTS test_zh CER  | Training Steps |
|---------------------------------------|--------------------|-----------------|
| f5-tts           | 3.02% (16 steps)  | 300k  |
| **f5-tts-semantic-token** |  **1.79% (16 steps)**   | 50k |

### DIFF
--- a/egs/wenetspeech4tts/TTS/prepare.sh
+++ b/egs/wenetspeech4tts/TTS/prepare.sh
@@ -111,7 +111,7 @@ if [ $stage -le 5 ] && [ $stop_stage -ge 5 ]; then
 fi
 
 if [ $stage -le 6 ] && [ $stop_stage -ge 6 ]; then
-  log "Stage 7: Split the ${prefix} cuts into train, valid and test sets (used by ./f5-tts)"
+  log "Stage 6: Split the ${prefix} cuts into train, valid and test sets (used by ./f5-tts)"
   if [ ! -f data/fbank/${prefix}_cuts_${subset}.jsonl.gz ]; then
     echo "Combining ${prefix} cuts"
     pieces=$(find data/fbank/ -name "${prefix}_cuts_${subset}.*.jsonl.gz")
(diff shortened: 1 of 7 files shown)


JinZr:
huge reduction on CER, i'll look into this pr.

egs/wenetspeech4tts/TTS/prepare.sh
```diff
@@ -139,3 +139,28 @@ if [ $stage -le 6 ] && [ $stop_stage -ge 6 ]; then
       touch data/fbank/.${prefix}_split.done
   fi
 fi
+
+if [ $stage -le 7 ] && [ $stop_stage -ge 7 ]; then
+  log "Stage 7: Extract cosyvoice2 FSQ token (used by ./f5-tts semantic token experiment)"
+  pip install s3tokenizer
```
JinZr, line 145: i would prefer adding an additional `requirements.txt` down the recipe directory as the manner adopted by the `valle` and `matcha` recipe.

yfyeung:
> The PR replaced the text tokens input in F5-TTS with pretrained cosyvoice2 semantic tokens. Or you can think of it as replacing the unet based flow matching model in the [cosyvoice2](https://arxiv.org/abs/2412.10117) with DiT.
> 
> Summary of the training results using wenetspeech4tts data：
> 
> Model	Seed-TTS test_zh CER	Training Steps
> f5-tts	3.02% (16 steps)	300k
> **f5-tts-semantic-token**	**1.79% (16 steps)**	50k

An interesting and impressive result! I am currently working on replacing MEL with CosyVoice2 tokens, i.e., discrete-E2TTS.

Benioh:
Is the model before F5 the same as CosyVoice2? If so, I believe the main capability comes from the language model , so it would be better to compare the results with CosyVoice2 to assess the improvement brought by Dit FlowMatching. Comparing it with F5 doesn’t make much sense.

egs/wenetspeech4tts/TTS/prepare.sh
yuekaizhang: Done. Thanks.

yuekaizhang:
> Is the model before F5 the same as CosyVoice2? If so, I believe the main capability comes from the language model , so it would be better to compare the results with CosyVoice2 to assess the improvement brought by Dit FlowMatching. Comparing it with F5 doesn’t make much sense.

Yes, the PR's results used pretrained Cosyvoice2 LLM to predict semantic tokens during inference time. Updated the README.md also to avoid the confusion.

JinZr (approved):
LGTM, thank you for the PR!

PR #1892, opened 2025-03-12
Add configurable max_jobs parameter to compute_fbank_librispeech.py

redsheep913:
### Problem
- In `compute_fbank_librispeech.py`, the number of parallel jobs was hardcoded to 80 (`num_jobs=num_jobs if ex is None else 80`)
- This can lead to system resource exhaustion on machines with limited capacity
- No way to configure the maximum number of concurrent processes

### Solution
- Added a new command-line parameter `--max_jobs` to control the maximum number of parallel processes
- Replaced the hardcoded value with this configurable parameter
- Set a reasonable default to prevent resource exhaustion

### Implementation
- Added parameter parsing for `--max_jobs` in the argument parser
- Used this parameter instead of the hardcoded value 80
- Ensured backward compatibility with existing scripts

### Benefits
- Improved resource utilization efficiency
- Prevents system crashes due to excessive parallelization
- Enhances script flexibility across different hardware environments

### Testing
- Tested with various `max_jobs` values on local environment
- Confirmed reasonable system resource usage
- Functionality works as expected with no performance degradation


egs/librispeech/ASR/chech_version.py
```diff
@@ -0,0 +1,11 @@
+import torch
```
csukuangfj, line 1: Can you add
```
#!/usr/bin/env python3
```
to the very beginning of this script?

(Also, can you make it executable?)
```python3
chmod +x ./check_version.py
```

egs/librispeech/ASR/local/compute_fbank_librispeech.py
```diff
@@ -134,20 +143,44 @@ def compute_fbank_librispeech(
...
-                extractor=extractor,
-                storage_path=f"{output_dir}/{prefix}_feats_{partition}",
-                # when an executor is specified, make more partitions
-                num_jobs=num_jobs if ex is None else 80,
-                executor=ex,
-                storage_type=LilcomChunkyWriter,
-            )
+            # 修改這部分代碼
+            if ex is None:
+                # 為 None 的情況（本地執行）創建自定義進程池上下文
```
csukuangfj, line 154: Can you use English instead of Chinese in the comment?
Or can you remove the comment since the code is easy to understand?

egs/librispeech/ASR/prepare.sh
```diff
@@ -99,9 +99,9 @@ if [ $stage -le 0 ] && [ $stop_stage -ge 0 ]; then
   #
   #   ln -sfv /path/to/LibriSpeech $dl_dir/LibriSpeech
   #
-  if [ ! -d $dl_dir/LibriSpeech/train-other-500 ]; then
-    lhotse download librispeech --full $dl_dir
-  fi
+  #if [ ! -d $dl_dir/LibriSpeech/train-other-500 ]; then
```
csukuangfj, line 102: Can you uncomment them?

egs/librispeech/ASR/prepare.sh
```diff
@@ -139,7 +139,7 @@ if [ $stage -le 3 ] && [ $stop_stage -ge 3 ]; then
   log "Stage 3: Compute fbank for librispeech"
   mkdir -p data/fbank
   if [ ! -e data/fbank/.librispeech.done ]; then
-    ./local/compute_fbank_librispeech.py
+    ./local/compute_fbank_librispeech.py --num-workers 4
```
csukuangfj, line 142:
```suggestion
    ./local/compute_fbank_librispeech.py --num-workers $nj
```

we have `nj=15` at the beginning of this script.

csukuangfj (changes requested):

egs/librispeech/ASR/local/compute_fbank_librispeech.py
redsheep913: > Can you use English instead of Chinese in the comment? Or can you remove the comment since the code is easy to understand?
I've updated the comments to English as requested.

egs/librispeech/ASR/prepare.sh
redsheep913: > Can you uncomment them?

Uncommented the code as requested. The conditional block is now active.

egs/librispeech/ASR/prepare.sh
redsheep913: > we have `nj=15` at the beginning of this script.

Good catch. Updated to use the $nj variable instead of the hardcoded value 4 for better flexibility.

egs/librispeech/ASR/local/compute_fbank_librispeech.py
```diff
@@ -134,20 +143,44 @@ def compute_fbank_librispeech(
...
-            cut_set = cut_set.compute_and_store_features(
-                extractor=extractor,
-                storage_path=f"{output_dir}/{prefix}_feats_{partition}",
-                # when an executor is specified, make more partitions
-                num_jobs=num_jobs if ex is None else 80,
-                executor=ex,
-                storage_type=LilcomChunkyWriter,
-            )
+
+            if ex is None:
```
csukuangfj, line 153: Can you leave this block unchanged? The original one is shorter and easier to follow.

csukuangfj (changes requested):

PR #1914, opened 2025-04-09
Fix bug: When resuming training from a checkpoint, model_avg was not assigned, resulting in a None error.

math345:
Considering moving the entire if rank == 0 code block outside could cause issues because:

1. If moved above, when continue_finetune is false, the model in copy.deepcopy(model) has not yet called model.load_state_dict().

2. If moved below, when continue_finetune is true, model_avg would come from copy.deepcopy(model) rather than the model_avg saved in the checkpoint.


pkufool (approved):
LGTM

PR #1941, opened 2025-05-12
fix bug: undefined name 'partial'

yfyeung:
```python
./egs/wenetspeech/KWS/zipformer/finetune.py:664:20: F821 undefined name 'partial'
    _encode_text = partial(encode_text, token_table=token_table, params=params)
                   ^
1     F821 undefined name 'partial'
```


csukuangfj (approved):
Thanks!

PR #1944, opened 2025-05-15
[WIP] Add Streaming Zipformer LLM recipe for ASR

yfyeung:


Johe-cqu:
Wonderful work on this branch—it looks very promising. I’m exploring a Zipformer + LLM setup and found your implementation. Could you share why it hasn’t been merged into main? Also, is it currently usable (training/inference) or still experimental? Thank you for your time and contributions. @yfyeung

yfyeung:
Maybe you can modify the projector and remove code about blank skipping. After that, it supports training and inference, but the performance is not better than whisper encoder.

whaozl:
Does this support streaming? Looking at the code, it doesn't seem to support streaming?

yfyeung:
> upport streaming? Looking at the code, it doesn't seem to support streaming?

No.

PR #1965, opened 2025-06-18
Add requirements.txt and pinyin.txt needed by zipvoice

pkufool:


egs/zipvoice/local/pinyin.txt
```diff
@@ -0,0 +1,1550 @@
+a
```
csukuangfj, line 1: Is it autogenerated or can it be autogerated?

pkufool: large? I don't think it is large.

pkufool: This is just like alphabets in English,  around 1500 items, the number of lines is even less than most of the python source files. And, note that, we actually have no chance to change it.  As far as I am concerned, this is the most suitable way.

PR #1814, opened 2024-11-27
Replace deprecated pytorch methods

lipeng31:
- torch.cuda.amp.GradScaler(...) => torch.amp.GradScaler("cuda", ...)
- torch.cuda.amp.autocast(...) => torch.amp.autocast("cuda", ...)

PR #1974, opened 2025-06-30
Fix CI tests.

csukuangfj:


csukuangfj:
See also #1814 

CC @lipeng31

PR #1990, opened 2025-07-14
Support left pad for `make_pad_mask` function

yfyeung:


danpovey (approved):
Looks good to me.  Test errors are in baker_zh recipe, likely an interface change in some package we depend on, we should ask maintainers but it's not related to this commit.

PR #2075, opened 2026-03-23
add linear warmup for cr-ctc loss

KarelVesely84:
- this prevents cr-ctc loss from diverging at the beginning of the training

crctc_warmup

Sometimes, the training was diverging with message:
`"If you see this error, it means that the gradient scale is too small."`

Investigation revealed that the cr-ctc loss can spike to high values at the beginning of the training (<1000 steps).
The influence of the cr-ctc loss should be diminished at the beginning of the training, when the ctc loss has
not learned the task yet.

Three variants were explored:
- original implementation
- linear warmup (0.0 .. 1.0) over initial 2k steps
- disabling cr-ctc for initial 2k steps

According to the `train/tot_cr_loss` curve the the linear warmup looks the best.

Best regards,
Karel from BUT

### DIFF
--- a/egs/aishell/ASR/zipformer/train.py
+++ b/egs/aishell/ASR/zipformer/train.py
@@ -882,8 +882,10 @@ def compute_loss(
         if params.use_ctc:
             loss += params.ctc_loss_scale * ctc_loss
             if use_cr_ctc:
-                loss += params.cr_loss_scale * cr_loss
-    
+                # linear warmup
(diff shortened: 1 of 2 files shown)


csukuangfj:
@yaozengwei Can you have a look?

danpovey:
I did bypass-and -merge since this is low risk, it only affects aishell recipe, and it looks OK to me.
BTW I am getting rid of CRCTC in the latest "zapformer" recipe.  For some reason it no longer seems to help in that context, after various modeling changes/improvements and getting rid of all randomization such as dropout.

PR #2082, opened 2026-05-13
Zapformer preview

danpovey:
We are working on the writeup but this is in case anyone wants to try the latest version.  Also note the --use-giga=True option in train.py and the --giga=True option in decode scripts.

## Summary by CodeRabbit

* **New Features**
  * Full Zapformer ASR: training, streaming & non‑streaming inference, many decoding modes (greedy/beam/LM/rescoring/oracle), ONNX/TorchScript export and runtimes, pretrained-model decode tools, profiling, and evaluation outputs.
  * End-to-end decoding utilities for batch and streaming audio, plus export/import tooling for ONNX/TorchScript models.

* **Documentation**
  * Added Zapformer training and decoding results with usage notes.

* **Chores**
  * Updated .gitignore to exclude generated PDF artifacts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/k2-fsa/icefall/pull/2082?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


pehonnet:
Thanks for sharing this early version, the training seems to be working fine. I noticed you mentioned sharing the onnx exporting script but I think it's not part of the PR... I was trying to make one but facing some issues so I thought I would ask, in case you already have the script available?

Thanks again for sharing your work!

danpovey:
I'm asking Kangwei about the exporting script.
I have tried to fix the issues identified by AI and have pushed the updated branch.  PLease let me know if there are important issues left un-fixed.

pehonnet:
I got an exporting script working, actually I think it's more that the model is not performing so well at the moment. 
Not sure if it is useful for you but this is how I trained on one RTX pro 6000:
```
./zapformer/train.py \
  --world-size 1 \
  --num-real-epochs 30 \
  --start-epoch 1  \
  --use-bf16 1 \
  --exp-dir zapformer/exp \
  --causal 1 \
  --full-libri 1 \
  --max-duration 1000
```
decoding command:
```
python ./zapformer/streaming_decode.py \
  --epoch 30 \
  --avg 15 \
  --causal 1 \
  --chunk-size 32 \
  --left-context-frames 256 \
  --exp-dir ./zapformer/exp \
  --decoding-method greedy_search \
  --num-decode-streams 1000 
```
results:
```
greedy_search	17.34	best for dev-clean
greedy_search	30.5	best for dev-other
greedy_search	17.21	best for test-clean
greedy_search	32.07	best for test-other
```
the commit was https://github.com/k2-fsa/icefall/pull/2082/commits/a56ebdec26609f5f4e1abfd5c01bd5f97dd2bfe9 so maybe things committed after improved it.

danpovey:
It seems the --base-lr default is set at an inappropriate value of 0.00065, it should be more like 0.02, although since you are using just one card I'd probably try more like 0.015.
Also you are averaging over way too many epochs, try more like 2 or 3.  (the --max-copies means that later epochs take longer)
Also try regular non-streaming decode first.

pehonnet:
Thanks!   should be resolved now, Kangwei made a fix.

danpovey:
Thanks for reporting your progress!  Here are some more updates, including some optimizer changes.
Kangwei recently made a bunch of fixes to the onnx-exporting, I believe, I have included those commits in what I just pushed.  (note, this is my local branch 3240.).  He said that he has not yet tested the streaming onnx-exporting though, although he may have made some fixes to it anyway.

danpovey:
BTW, it's remarkable that it worked at all because that default LR was 30 times too low.  If you have only one job you may need to reduce the LR a bit: for example, from the default of 0.02 to 0.015 or 0.013 or something like that.

videodanchik:
Hi @danpovey, first, thanks a lot to you and your team for sharing this new architecture. I have a question here, have you tried CR-CTC(+pruned transducer) with this? Does this new architecture constitute similar improvements as Zipformer with CR-CTC over no CR-CTC version?

danpovey:
> Hi @danpovey, first, thanks a lot to you and your team for sharing this new architecture. I have a question here, have you tried CR-CTC(+pruned transducer) with this? Does this new architecture constitute similar improvements as Zipformer with CR-CTC over no CR-CTC version?

No; for reasons that are unclear to me, the CR-CTC loss does not seem to help in this setup and with this model.

danpovey:
> For non streaming, I get indeed much better results:
> 
> ```shell
> python ./zapformer/decode.py     --epoch 30     --avg 3     --exp-dir ./zapformer/exp     --max-duration 1000     --decoding-method greedy_search --causal true --chunk-size 32 --left-context-frames 256
> greedy_search	3.56	best for dev-clean
> greedy_search	9.85	best for dev-other
> greedy_search	3.79	best for test-clean
> greedy_search	9.82	best for test-other
> ```
> 
> If I try with avg 15 to compare, I get similar results (~3.5% - 9.9%)
> 
> For the streaming script, with only 3 epochs it's even worse than 15:
> 
> ```shell
> python ./zapformer/streaming_decode.py   --epoch 30   --avg 3  --causal 1   --chunk-size 32   --left-context-frames 256   --exp-dir ./zapformer/exp   --decoding-method greedy_search   --num-decode-streams 1000
> greedy_search	31.47	best for dev-clean
> greedy_search	30.95	best for test-clean
> greedy_search	45.98	best for dev-other
> greedy_search	46.06	best for test-other
> ```
> 
> so there may be something wrong in the streaming decoding... It could also be due to some small changes I had to do to make the onnx export run in the first place. If I have time I will test your updated branch and report again.
> 
> Thanks!

Thanks for reporting this, I'll check with Kangwei.  Let us know what you had to do to make the onnx work!

pkufool:
> For non streaming, I get indeed much better results:
> 
> ```shell
> python ./zapformer/decode.py     --epoch 30     --avg 3     --exp-dir ./zapformer/exp     --max-duration 1000     --decoding-method greedy_search --causal true --chunk-size 32 --left-context-frames 256
> greedy_search	3.56	best for dev-clean
> greedy_search	9.85	best for dev-other
> greedy_search	3.79	best for test-clean
> greedy_search	9.82	best for test-other
> ```
> 
> If I try with avg 15 to compare, I get similar results (~3.5% - 9.9%)
> 
> For the streaming script, with only 3 epochs it's even worse than 15:
> 
> ```shell
> python ./zapformer/streaming_decode.py   --epoch 30   --avg 3  --causal 1   --chunk-size 32   --left-context-frames 256   --exp-dir ./zapformer/exp   --decoding-method greedy_search   --num-decode-streams 1000
> greedy_search	31.47	best for dev-clean
> greedy_search	30.95	best for test-clean
> greedy_search	45.98	best for dev-other
> greedy_search	46.06	best for test-other
> ```
> 
> so there may be something wrong in the streaming decoding... It could also be due to some small changes I had to do to make the onnx export run in the first place. If I have time I will test your updated branch and report again.
> 
> Thanks!

Can you share your training command, my model trained with num-real-epochs=30, and decoded with chunk-size=32 left-context-frames=128, the WER on test-clean is 2.38, much lower than yours (is 3.79).

pehonnet:
> > For non streaming, I get indeed much better results:
> > ```shell
> > python ./zapformer/decode.py     --epoch 30     --avg 3     --exp-dir ./zapformer/exp     --max-duration 1000     --decoding-method greedy_search --causal true --chunk-size 32 --left-context-frames 256
> > greedy_search	3.56	best for dev-clean
> > greedy_search	9.85	best for dev-other
> > greedy_search	3.79	best for test-clean
> > greedy_search	9.82	best for test-other
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If I try with avg 15 to compare, I get similar results (~3.5% - 9.9%)
> > For the streaming script, with only 3 epochs it's even worse than 15:
> > ```shell
> > python ./zapformer/streaming_decode.py   --epoch 30   --avg 3  --causal 1   --chunk-size 32   --left-context-frames 256   --exp-dir ./zapformer/exp   --decoding-method greedy_search   --num-decode-streams 1000
> > greedy_search	31.47	best for dev-clean
> > greedy_search	30.95	best for test-clean
> > greedy_search	45.98	best for dev-other
> > greedy_search	46.06	best for test-other
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > so there may be something wrong in the streaming decoding... It could also be due to some small changes I had to do to make the onnx export run in the first place. If I have time I will test your updated branch and report again.
> > Thanks!
> 
> Can you share your training command, my model trained with num-real-epochs=30, and decoded with chunk-size=32 left-context-frames=128, the WER on test-clean is 2.38, much lower than yours (is 3.79).

See for the training command:
https://github.com/k2-fsa/icefall/pull/2082#issuecomment-4541458276
but as mentioned, it was on an older commit. Probably more recent ones give better results.

```
./zapformer/train.py \
  --world-size 1 \
  --num-real-epochs 30 \
  --start-epoch 1  \
  --use-bf16 1 \
  --exp-dir zapformer/exp \
  --causal 1 \
  --full-libri 1 \
  --max-duration 1000

```

kobenaxie:
Hi @danpovey , thank you for sharing `Zapformer`, If training Zapformer with other datasets, how should the `--batches-per-epoch` parameter be adjusted?

danpovey:
> Hi @danpovey , thank you for sharing `Zapformer`, If training Zapformer with other datasets, how should the `--batches-per-epoch` parameter be adjusted?

batcher-per-epoch is not an important configuration value.  It is supposed to be the number of batches in the 1st epoch, but it will make almost no difference if you don't set it, for typical learning rate schedules, because it automatically figures out the real number of batches per epoch by the second epoch.

egs/librispeech/ASR/zapformer/train.py
```diff
@@ -0,0 +1,1796 @@
...
+        (model.module if isinstance(model, DDP) else model).encoder.compute_projection_overlap(verbose=True)  # for diagnostics
+
+        seed = params.seed + 50 * epoch  + 512 * rank
+
+        specaug = AlternatingSpecAugment(
+            seed=seed,
+        )  # otherwise use all default settings.
+
+        if torch.cuda.is_available():
+            with torch.cuda.device(rank):
```
kobenaxie, line 1646: The `local rank` should be passed here, but the `global rank` was used instead, which will cause errors during multi-machine DDP training.

danpovey: OK thanks.  I think I'll remove that code about setting the rank, it was set while I was debugging something, while I had a misconception about where the random number generator "lives".

danpovey: I will remove this code, I later realized it was not necessary.

kobenaxie:
> > Hi @danpovey , thank you for sharing `Zapformer`, If training Zapformer with other datasets, how should the `--batches-per-epoch` parameter be adjusted?
> 
> batcher-per-epoch is not an important configuration value. It is supposed to be the number of batches in the 1st epoch, but it will make almost no difference if you don't set it, for typical learning rate schedules, because it automatically figures out the real number of batches per epoch by the second epoch.

When training on a large dataset, such as 100Khours,  typically we don’t need to train for as many as 40 epochs. Should I need to adjust the `--max-copies` or some other parameter?

danpovey:
Yes, you would probably want a much smaller --max-copies, e.g. 2 or 1... it's there to prevent overfitting from a large number of epochs, if the num epochs is small it shouldn't be necessary.

PR #2086, opened 2026-06-05
Allow Streaming Zipformer Models to Be Exported as Non-Streaming Models

csukuangfj:
## Summary by CodeRabbit

* **New Features**
  * Standalone non-streaming ONNX export CLI.
  * Opt-in int32-input support for ONNX exports.
  * Optional FP16 and int8 quantized export outputs.
  * Increased default positional length limit (1000 → 2000).

* **Bug Fixes**
  * Prevented unintended int8 file generation when quantization is disabled.

* **Refactor**
  * Unified masking and tensor-type handling across export and runtime paths.
  * Added ONNX-friendly non-streaming convolution wrapper and adjusted padding/attention behavior for export.

PR #2095, opened 2026-06-29
Fix exporting streaming zipformer models as non-streaming models

csukuangfj:
See errors reported at 
https://github.com/k2-fsa/icefall/pull/2086#issuecomment-4708624020

cc @nshmyrev

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved convolution handling for short input sequences so results are now computed correctly when the sequence length is smaller than the kernel size.
  * Preserved the existing behavior for longer sequences, with no change to standard processing.


nshmyrev:
@csukuangfj wonderful, many thanks!

PR #2097, opened 2026-07-13
Fix voxpopuli prepare.sh Stage 6 failing for any lang other than en

HenryVarro666:
Fixes #2064.

Stage 6 of `egs/voxpopuli/ASR/prepare.sh` hardcodes the `voxpopuli-asr-en` manifest prefix, while stages 3-5 build it as `voxpopuli-${task}-${lang}`. With any `lang` other than `en` (e.g. `lang=it`, the reporter's case), validation fails with `file not found: data/fbank/voxpopuli-asr-en_cuts_voxpopuli.jsonl.gz`.

This parameterizes the cuts path and the log path the same way as the other stages — the exact workaround described in the issue:

```
data/fbank/voxpopuli-${task}-${lang}_cuts_${dataset}.jsonl.gz
data/fbank/log/validate_voxpopuli-${task}-${lang}_cuts_${dataset}.log
```

Verified with `bash -n` and by resolving the path with `task=asr lang=it dataset=dev` → `data/fbank/voxpopuli-asr-it_cuts_dev.jsonl.gz`, matching what Stage 5 produces.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated VoxPopuli cutset validation to use task- and language-specific manifest files.
  * Validation output logs now match the same task and language context for clearer traceability.

PR #2096, opened 2026-07-09
Add Europarl-ST SRT recipe with LCMA-SRT

linanjie0820:
## Summary

- Add joint multilingual speech recognition and translation (SRT) recipe for Europarl-ST dataset
- Model: LCMA-SRT ([Language-Conditional Mixture-of-Experts Adapters for Joint Multilingual Speech Recognition and Translation](https://aclanthology.org/2026.acl-long.1634.pdf))
- 9 languages, 72 translation directions
- Includes: data preparation pipeline, 2-stage training scripts, decoding scripts, and pre-trained model links

## Structure

egs/europarl_st/SRT/
├── prepare.sh                     # Data preparation
├── local/                             # Preprocessing scripts
├── lcma_srt/                       # Model code
│   ├── train/stage1/            # ASR pretraining scripts
│   ├── train/stage2/            # Joint SRT training scripts
│   ├── decode/stage1/       # ASR decoding scripts
│   └── decode/stage2/       # Joint decoding scripts
├── README.md
└── RESULTS.md

## Reference

- Paper: https://aclanthology.org/2026.acl-long.1634/
- Pre-trained checkpoints: https://osf.io/rnuhv/
- Standalone repo: https://github.com/linanjie0820/LCMA-SRT

## Summary by CodeRabbit

* **New Features**
  * Added an end-to-end Europarl-ST SRT recipe, including distributed training launchers, a configurable decoding pipeline, and multilingual stage-based decode scripts.
  * Introduced LCMA-SRT model components for ASR/translation, with streaming-capable encoding and optional MoE routing.

* **Documentation**
  * Added full project and recipe documentation, plus static multilingual results tables and checkpoint references.
  * Added detailed data-prep instructions and per-stage script guides.

* **Chores**
  * Added Apache 2.0 licensing text and expanded local preprocessing utilities (manifest checks/filtering, text normalization, audio conversion, FBANK cut generation, and BPE training).

PR #2098, opened 2026-07-13
Fix black formatting in europarl_st beam_search.py

HenryVarro666:
`egs/europarl_st/SRT/lcma_srt/beam_search.py` (added in #2096) fails the `style_check` workflow — black 22.3.0 wants to reformat it, so style_check is currently red for every PR (e.g. it's what fails on #2097, whose change is an unrelated shell script).

Reformatted with the exact versions the workflow installs (`black==22.3.0`, `click==8.1.0`); `black --check` and `isort --check` both pass on the file now. No functional change (+3/-1).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the error message shown when an invalid language token ID is provided during beam search.
  * Clarified that the language token ID must be an integer.

### DIFF
--- a/egs/europarl_st/SRT/lcma_srt/beam_search.py
+++ b/egs/europarl_st/SRT/lcma_srt/beam_search.py
@@ -1448,7 +1448,9 @@ def modified_beam_search(
         if lang_token_id is None:
             raise ValueError("force_first_lang=True but lang_token_id not provided")
         if not isinstance(lang_token_id, int):
-            raise TypeError("lang_token_id must be int (same target lang for entire batch).")
+            raise TypeError(
+                "lang_token_id must be int (same target lang for entire batch)."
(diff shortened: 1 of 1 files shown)


HenryVarro666:
Heads up for reviewers: the red CI here is unrelated to this PR, which is a formatting-only change to `europarl_st/beam_search.py`. The failures come from pre-existing environment issues on `master` — the shared test tree aborts during pytest **collection** with `ModuleNotFoundError: No module named 'kaldi_native_fbank'` (from `test_rknn_on_cpu_simulator_ctc_streaming.py`), and the `rknn` workflow fails for the same missing dependency. None of the failing jobs touch the file changed here. Happy to help fix the CI dependency separately if useful.

csukuangfj (approved):

